PR #1115, opened 2024-11-27
[CI] Install DependaBot

giordano:


DilumAluthge:
Can you increase the max number of open PRs? It defaults to 5, which is a little low.

Also, we could run it monthly if we want (instead of weekly), which might reduce the noise a little bit.

.github/dependabot.yml
```diff
@@ -0,0 +1,7 @@
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/" # Location of package manifests
+    schedule:
+      interval: "weekly"
```
inkydragon:
```suggestion
      interval: "monthly"
```

inkydragon:
> Can you increase the max number of open PRs?

How many do we want, 50? 100?

[Configuration options for the dependabot.yml file - GitHub Docs](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#open-pull-requests-limit)

fredrikekre:
Why would you need more than 5? There are only 2 or 3 actions which are dependabot-updateable anyway?

PR #1118, opened 2024-11-27
Run Linux tests with -O0

andreasnoack:
Since most of the time is spent on compilation, it might reduce CI time.

### DIFF
--- a/.buildkite/runtests.yml
+++ b/.buildkite/runtests.yml
@@ -23,27 +23,27 @@ steps:
       arch: "x86_64"
       cpuset_limited: "true"
     timeout_in_minutes: 120
-  # macOS aarch64
-  - label: ":macos: macos-aarch64"
-    plugins:
-      - JuliaCI/julia#v1:
-          version: "nightly"
-    command: |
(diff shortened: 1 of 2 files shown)


andreasnoack:
- [-O0 (this PR)](https://buildkite.com/julialang/linearalgebra-dot-jl/builds/29#01936e19-7c23-4ee6-818e-9214a819a2c4) ~25 minutes
- [-O2 (the default)](https://buildkite.com/julialang/linearalgebra-dot-jl/builds/28#01936da3-9400-4b5b-9765-b1054df6bcb0) ~57 minutes

So a few allocation tests fail with `-O0`. Maybe they could be run separately. The jobs are still highly unbalanced but with `-O0` we wouldn't pay with extra compilation time by splitting triagular.jl.

KristofferC:
I think we can do this but run with default settings on any `backports-*` and `release-*` branches.

ViralBShah:
@vtjnash suggests running with `-O1` may be better, and may pass the allocation tests.

andreasnoack:
It looks like fully interpreted is the winning strategy

ViralBShah:
Yeah, `-O0` is much faster.

PR #1119, opened 2024-11-27
use an explicit file extension when creating sysimage

KristofferC:
Otherwise the windows compiler adds a `.exe` to it.


KristofferC:
Woho green!

PR #1120, opened 2024-11-27
try use updated windows compilers in PackageCompiler

KristofferC:
See if this gets rid of the warnings. Ref https://github.com/JuliaLang/PackageCompiler.jl/pull/1012


KristofferC:
Seems like the warnings are gone 🎉

PR #1121, opened 2024-11-27
run with -O0 unless on release or backport branch

KristofferC:
Follow up to #1118


ViralBShah:
Worth getting this in?

KristofferC:
I think the CI time is pretty ok now.

ViralBShah:
I see it is still 40 mins.

dkarrasch:
The actual tests run in 10+/-2 minutes, only Windows is 15.

PR #1124, opened 2024-11-28
Remove old debug functionality in triangular.jl

andreasnoack:
I noticed that this old debug code was still in the test file when looking into speeding up the tests. It predates the testset functionality so we should just use that instead which was already the case for the more recent parts of the test file.

### DIFF
--- a/test/triangular.jl
+++ b/test/triangular.jl
@@ -2,7 +2,6 @@
 
 module TestTriangular
 
-debug = false
 using Test, LinearAlgebra, Random
 using LinearAlgebra: BlasFloat, errorbounds, full!, transpose!,
     UnitUpperTriangular, UnitLowerTriangular,
@@ -16,14 +15,13 @@ using .Main.SizedArrays
 isdefined(Main, :FillArrays) || @eval Main include(joinpath($(BASE_TEST_PATH), "testhelpers", "FillArrays.jl"))
 using .Main.FillArrays
 
-debug && println("Triangular matrices")
-
 n = 9
 Random.seed!(123)
 
-debug && println("Test basic type functionality")
-@test_throws DimensionMismatch LowerTriangular(randn(5, 4))
-@test LowerTriangular(randn(3, 3)) |> t -> [size(t, i) for i = 1:3] == [size(Matrix(t), i) for i = 1:3]
+@testset "Test basic type functionality" begin
+    @test_throws DimensionMismatch LowerTriangular(randn(5, 4))
+    @test LowerTriangular(randn(3, 3)) |> t -> [size(t, i) for i = 1:3] == [size(Matrix(t), i) for i = 1:3]
+end
 
 struct MyTriangular{T, A<:LinearAlgebra.AbstractTriangular{T}} <: LinearAlgebra.AbstractTriangular{T}
     data :: A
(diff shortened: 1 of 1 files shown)


test/triangular.jl
```diff
@@ -628,73 +620,69 @@ Aimg    = randn(n, n)/2
...
-        debug && println("\ntype of A: ", eltya, " type of b: ", eltyb, "\n")
+        @testset "Solve upper triangular system" begin
+            Atri = UpperTriangular(lu(A).U) |> t -> eltya <: Complex && eltyb <: Real ? real(t) : t # Here the triangular matrix can't be too badly conditioned
+            b = convert(Matrix{eltyb}, Matrix(Atri)*fill(1., n, 2))
+            x = Atri \ b
 
-        debug && println("Solve upper triangular system")
-        Atri = UpperTriangular(lu(A).U) |> t -> eltya <: Complex && eltyb <: Real ? real(t) : t # Here the triangular matrix can't be too badly conditioned
-        b = convert(Matrix{eltyb}, Matrix(Atri)*fill(1., n, 2))
-        x = Matrix(Atri) \ b
```
stevengj, line 645: Not sure why the old test converted to `Matrix` here?

andreasnoack: I also couldn't see the point in doing that so I went back to the PR (of mine) that introduced it, but it didn't provide any explanation. I guess it was a mistake.

test/triangular.jl
```diff
@@ -628,73 +620,69 @@ Aimg    = randn(n, n)/2
...
+                @test norm(abs.(b[:,i] - Atri*x[:,i]), Inf) <= γ * norm(Atri, Inf) * norm(x[:,i], Inf)
             end
         end
 
-        debug && println("Test backward error [JIN 5705]")
-        for i = 1:size(b, 2)
-            @test norm(abs.(b[:,i] - Atri*x[:,i]), Inf) <= γ * norm(Atri, Inf) * norm(x[:,i], Inf)
-        end
+        @testset "Solve lower triangular system" begin
+            Atri = UpperTriangular(lu(A).U) |> t -> eltya <: Complex && eltyb <: Real ? real(t) : t # Here the triangular matrix can't be too badly conditioned
```
dkarrasch: Should this be

```suggestion
            Atri = LowerTriangular(lu(A).U) |> t -> eltya <: Complex && eltyb <: Real ? real(t) : t # Here the triangular matrix can't be too badly conditioned
```
? At least, that's what the heading says, and also `UpperTriangular` seems to be covered above, lines 632ff.

test/triangular.jl
```diff
@@ -628,73 +620,69 @@ Aimg    = randn(n, n)/2
...
+            x = Atri \ b
 
-        debug && println("Solve lower triangular system")
-        Atri = UpperTriangular(lu(A).U) |> t -> eltya <: Complex && eltyb <: Real ? real(t) : t # Here the triangular matrix can't be too badly conditioned
-        b = convert(Matrix{eltyb}, Matrix(Atri)*fill(1., n, 2))
-        x = Matrix(Atri)\b
+            # Test error estimates
+            if eltya != BigFloat && eltyb != BigFloat
+                for i = 1:2
+                    @test  norm(x[:,1] .- 1) <= errorbounds(UpperTriangular(A), x, b)[1][i]
```
dkarrasch: Same here:

```suggestion
                    @test  norm(x[:,1] .- 1) <= errorbounds(LowerTriangular(A), x, b)[1][i]
```

dkarrasch (commented):
Maybe I misread the diff, maybe not.

test/triangular.jl
dkarrasch: I think it doesn't matter. Generic solves check for special structure, but I agree it's clearer to call triangular solves "directly" by types.

PR #1125, opened 2024-11-28
allow passing ipiv to lu! factorizations

longemen3000:
when performing LU factorizations, a `Vector{BlasInt}` is allocated each time. in conjuction with a matrix `A`.  `lu!` allows reusing a matrix, but it still allocates the vector. This PR allows passing `ipiv` as an input to `lu!`, so no allocations are performed, if required.

### DIFF
--- a/src/lu.jl
+++ b/src/lu.jl
@@ -87,7 +87,7 @@ end
 
 # the following method is meant to catch calls to lu!(A::LAPACKArray) without a pivoting strategy
 lu!(A::StridedMatrix{<:BlasFloat}; check::Bool = true, allowsingular::Bool = false) = lu!(A, RowMaximum(); check, allowsingular)
-function lu!(A::StridedMatrix{T}, ::RowMaximum,ipiv::AbstractVector{BlasInt} = Vector{BlasInt}(undef,min(size(A)...)); check::Bool = true, allowsingular::Bool = false) where {T<:BlasFloat}
+function lu!(A::StridedMatrix{T}, ::RowMaximum, ipiv::AbstractVector{BlasInt} = Vector{BlasInt}(undef,min(size(A)...)); check::Bool = true, allowsingular::Bool = false) where {T<:BlasFloat}
     lpt = LAPACK.getrf!(A, ipiv; check)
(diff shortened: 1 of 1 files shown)


src/lu.jl
```diff
@@ -87,20 +87,20 @@ end
 
 # the following method is meant to catch calls to lu!(A::LAPACKArray) without a pivoting strategy
 lu!(A::StridedMatrix{<:BlasFloat}; check::Bool = true, allowsingular::Bool = false) = lu!(A, RowMaximum(); check, allowsingular)
-function lu!(A::StridedMatrix{T}, ::RowMaximum; check::Bool = true, allowsingular::Bool = false) where {T<:BlasFloat}
-    lpt = LAPACK.getrf!(A; check)
+function lu!(A::StridedMatrix{T}, ::RowMaximum,ipiv::AbstractVector{BlasInt} = Vector{BlasInt}(undef,min(size(A)...)); check::Bool = true, allowsingular::Bool = false) where {T<:BlasFloat}
```
stevengj:
```suggestion
function lu!(A::StridedMatrix{T}, ::RowMaximum, ipiv::AbstractVector{BlasInt} = Vector{BlasInt}(undef,min(size(A)...)); check::Bool = true, allowsingular::Bool = false) where {T<:BlasFloat}
```

stevengj:
Should this go in the new LinearAlgeba.jl repo instead?

longemen3000:
Isn't this the new LinearAlgebra Repo?

stevengj:
Whoops, right.  Sorry, I thought was looking at `JuliaLang/julia`.

stevengj:
An alternative API would be:
```jl
lu!(F::LU, A::StridedMatrix, ...)
```
that lets you overwrite an existing `LU` object.  This way we wouldn't have to expose internals like `ipiv`.

Indeed, it looks like we already have such a method for LU factorization of `Tridiagonal` matrices, and we also have such a method for sparse arrays.  Not only would adding this to `StridedMatrix` be more consistent, but it is more flexible because the internals of the factorization object may change for different matrix types.

longemen3000:
closing in favor of #1131

PR #1131, opened 2024-12-01
define lu!(F::LU,A) for general matrices

longemen3000:
alternative to #1125 . Defines `lu!(F::LU{<:Any,<:AbstractMatrix},A)` and `lu!(F::LU{<:Any,<:StridedMatrix{<:BlasFloat}},A)`.

closes #1125


src/lu.jl
```diff
@@ -102,6 +102,39 @@ function lu!(A::HermOrSym{T}, pivot::Union{RowMaximum,NoPivot,RowNonZero} = lupi
     end
     lu!(A.data, pivot; check, allowsingular)
 end
+
+#reusing LU object
+#lu!(F::LU,A) should be dispatched on the type of matrix stored in the LU factorization.
+#but special care needs to be done in the HermOrSym case
+function _lu_copy!(A,x)
```
stevengj: you don't use an underscore in the other versions?

Also would like nicer if you called the second argument `X` or `B` to make it look more matrix-like

src/lu.jl
```diff
@@ -102,6 +102,39 @@ function lu!(A::HermOrSym{T}, pivot::Union{RowMaximum,NoPivot,RowNonZero} = lupi
...
+    @inbounds if isa(x, Hermitian) # realify diagonal
+        for i in axes(x, 1)
+            A[i,i] = x[i,i]
+        end
+    end
+    return A
+end
+
+function lu!(F::LU{<:Any,<:StridedMatrix{<:BlasFloat}}, A; check::Bool = true, allowsingular::Bool = false)
+    lu_copy!(F.factors,A)
```
stevengj: I'm unclear on why `copy!(F.factors, A)` or `copyto!(F.factors, A)` doesn't work here.

longemen3000: some code above does this to speed up the `copyto!(A,B::Union{Hermitian,Symmetric})` case, but you are right, this should be optimized at the `copyto!` level, not here.

src/lu.jl
```diff
@@ -102,6 +102,25 @@ function lu!(A::HermOrSym{T}, pivot::Union{RowMaximum,NoPivot,RowNonZero} = lupi
     end
     lu!(A.data, pivot; check, allowsingular)
 end
+
+#reusing LU object
+#lu!(F::LU,A) should be dispatched on the type of matrix stored in the LU factorization.
+
+function lu!(F::LU{<:Any,<:StridedMatrix{<:BlasFloat}}, A; check::Bool = true, allowsingular::Bool = false)
+    copyto!(F.factors,A)
```
stevengj:
```suggestion
    copyto!(F.factors, A)
```
The usual style is to separate arguments with commas (similarly below).

src/lu.jl
```diff
@@ -102,6 +102,25 @@ function lu!(A::HermOrSym{T}, pivot::Union{RowMaximum,NoPivot,RowNonZero} = lupi
...
+    lpt = LAPACK.getrf!(F.factors, F.ipiv; check)
+    check && _check_lu_success(lpt[3], allowsingular)
+    return F
```
stevengj: Shouldn't this be something like
```suggestion
    return LU{eltype(F),typeof(lpt[1]),typeof(lpt[2])}(lpt[1], lpt[2], lpt[3])
```
as otherwise it doesn't update the `info` field?

Returning a new LU object should be cheap since it is immutable and just contains pointers to the same arrays.

Of course, in the common case where you don't allow singular factorizations, the `info` field should always be zero and so this should be equivalent to the original `F` object.

src/lu.jl
longemen3000: seems like there is already a specialized `copyto!` for this case, introduced here: https://github.com/JuliaLang/LinearAlgebra.jl/commit/c9c36266bab698f43e345645662a6f4a564b8494

src/lu.jl
```diff
@@ -102,6 +102,24 @@ function lu!(A::HermOrSym{T}, pivot::Union{RowMaximum,NoPivot,RowNonZero} = lupi
...
+#lu!(F::LU,A) should be dispatched on the type of matrix stored in the LU factorization.
+
+function lu!(F::LU{<:Any,<:StridedMatrix{<:BlasFloat}}, A; check::Bool = true, allowsingular::Bool = false)
+    copyto!(F.factors, A)
+    lpt = LAPACK.getrf!(F.factors, F.ipiv; check)
+    check && _check_lu_success(lpt[3], allowsingular)
+    return LU{T,typeof(lpt[1]),typeof(lpt[2])}(lpt[1], lpt[2], lpt[3])
+end
+
+function lu!(F::LU{<:Any,<:AbstractMatrix}, A; check::Bool = true, allowsingular::Bool = false)
```
stevengj, line 151: Should have a docstring.  I would move the generic function up to come before the `StridedMatrix` one, attach its docstring, and delete the `#reusing LU object` comment (which will be redundant with the docstring).

src/lu.jl
```diff
@@ -102,6 +102,65 @@ function lu!(A::HermOrSym{T}, pivot::Union{RowMaximum,NoPivot,RowNonZero} = lupi
...
+
+#lu!(F::LU,A) should be dispatched on the type of matrix stored in the LU factorization.
+"""
+    lu!(F::LU, pivot = RowMaximum(); check = true, allowsingular = false) -> LU
+
+`lu!` is the same as [`lu`](@ref), but saves space by overwriting the
+input `F`, instead of creating a copy.
+
+!!! compat "Julia 1.12"
+    reusing `LU` factorizations in `lu!` require Julia 1.12 or later.
```
stevengj:
```suggestion
    Reusing dense `LU` factorizations in `lu!` require Julia 1.12 or later.
```

ViralBShah:
Is this good to merge?

stevengj (approved):

stevengj:
Does this go into a NEWS entry somewhere?

ViralBShah:
We should add to the LinearAlgebra section of the Julia NEWS.

ViralBShah:
The docs CI is failing.

ViralBShah:
@KristofferC  is this a problem?

┌ Warning: The active manifest file has dependencies that were resolved with a different julia version (1.12.0-DEV). Unexpected behavior may occur.

PR #1141, opened 2024-12-05
don't refer to internal variable names in gemv exceptions

stevengj:
Currently, the `DimensionMismatch` exceptions refer to internal parameter names of `gemv!`, which is confusing because those don't necessarily match the caller's variables.  For example:
```jl
julia> B = rand(1000,100); c = rand(1000);

julia> B * c
ERROR: DimensionMismatch: second dimension of A, 100, does not match length of x, 1000
Stacktrace:
 [1] gemv!
   @ ~/.julia/juliaup/julia-1.11.2+0.x64.apple.darwin14/share/julia/stdlib/v1.11/LinearAlgebra/src/matmul.jl:438 [inlined]
```
This PR changes the error messages to refer to "matrix" instead of "A" and "input vector" instead of "x", etc.

### DIFF
--- a/src/matmul.jl
+++ b/src/matmul.jl
@@ -572,9 +572,9 @@ Base.@constprop :aggressive function gemv!(y::StridedVector{T}, tA::AbstractChar
                α::Number=true, β::Number=false) where {T<:BlasFloat}
     mA, nA = lapack_size(tA, A)
     nA != length(x) &&
-        throw(DimensionMismatch(lazy"second dimension of A, $nA, does not match length of x, $(length(x))"))
+        throw(DimensionMismatch(lazy"second dimension of matrix, $nA, does not match length of input vector, $(length(x))"))
     mA != length(y) &&
-        throw(DimensionMismatch(lazy"first dimension of A, $mA, does not match length of y, $(length(y))"))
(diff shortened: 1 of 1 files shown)


jishnub (approved):

mcabbott:
Seems like a good idea. It's not just `gemv!` though:
```julia
julia> [1 2 3; 4 5 6] * (1:10)
ERROR: DimensionMismatch: matrix A has dimensions (2,3), vector B has length 10
Stacktrace:
 [1] _generic_matvecmul!
   @ ~/.julia/dev/julia/usr/share/julia/stdlib/v1.12/LinearAlgebra/src/matmul.jl:946 [inlined]
  
julia> rand(3,4) * rand(3,4)
ERROR: DimensionMismatch: A has dimensions (3,4) but B has dimensions (3,4)
Stacktrace:
 [1] gemm_wrapper!
   @ ~/.julia/dev/julia/usr/share/julia/stdlib/v1.12/LinearAlgebra/src/matmul.jl:752 [inlined]
```

stevengj:
Yeah, I knew there were more cases, but didn't have a chance to track them all down in this PR.  File an issue?

PR #1142, opened 2024-12-06
ci: add linux-i686

### LINKED ISSUE #1109: Enable Linux 32-bit in CI
I noticed in the PRs that we only test 64-bit platforms in CI. We should at least add Linux 32-bit to make sure we are not regressing.

inkydragon:
Closed #1109

### DIFF
--- a/.buildkite/runtests.yml
+++ b/.buildkite/runtests.yml
@@ -14,6 +14,7 @@ steps:
       # Julia installation inside the sandbox
       - JuliaCI/julia#v1:
           version: "nightly"
+          arch: "i686"
     command: |
       julia --color=yes --project=.ci -e 'using Pkg; Pkg.instantiate()'
       julia --color=yes --project=.ci .ci/create_sysimage_and_run_tests.jl
--- a/.buildkite/runtests.yml
+++ b/.buildkite/runtests.yml
@@ -23,7 +23,6 @@ steps:
       os: "linux"
       arch: "x86_64"
       cpuset_limited: "true"
-    timeout_in_minutes: 120
   # Linux x86_64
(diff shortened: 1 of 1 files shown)


.buildkite/runtests.yml
```diff
@@ -1,5 +1,28 @@
...
 steps:
+  # Linux x86
+  - label: ":linux: linux-i686"
+    plugins:
+      # Julia installation outside the sandbox
+      - JuliaCI/julia#v1:
+          version: '1.10'
+      - staticfloat/sandbox#v2:
+          rootfs_url: https://github.com/JuliaCI/rootfs-images/releases/download/v6.00/package_linux.i686.tar.gz
+          rootfs_treehash: "a0c4756ee7827b3a75d4a62ff543315c1547e3d6"
```
inkydragon, line 11: https://github.com/JuliaCI/julia-buildkite/blob/161cc53f04ca85a28a315b314500857009dd8c9b/pipelines/main/platforms/build_linux.arches#L2

.buildkite/runtests.yml
```diff
@@ -1,5 +1,28 @@
...
+          version: "nightly"
+    command: |
+      julia --color=yes --project=.ci -e 'using Pkg; Pkg.instantiate()'
+      julia --color=yes --project=.ci .ci/create_sysimage_and_run_tests.jl
+    agents:
+      queue: "julia"
+      os: "linux"
+      arch: "x86_64"
+      cpuset_limited: "true"
+    timeout_in_minutes: 120
```
inkydragon: Let's see if we don't need this.
I think `#default TIMEOUT 80` is enough.

PR #1144, opened 2024-12-10
setup coverage

KristofferC:
Maybe this will work?

### DIFF
--- a/.buildkite/runtests.yml
+++ b/.buildkite/runtests.yml
@@ -17,8 +17,8 @@ steps:
       - JuliaCI/julia#v1:
           version: "nightly"
     command: |
-      julia --color=yes --project=.ci -e 'using Pkg; Pkg.instantiate()'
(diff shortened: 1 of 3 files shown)


andreasnoack:
Looks like there are no coverage files and there is also an error https://buildkite.com/julialang/linearalgebra-dot-jl/builds/96#0193b06d-7eab-427d-bf58-4416f7a7db8a/412-527. Do you need to set `--code-coverage`? Is a CODECOV_TOKEN needed?

KristofferC:
We have a CODECOV_TOKEN set but that is for GitHub Actions which will not be of any use, indeed. I'm looking at https://github.com/JuliaGPU/GPUCompiler.jl/blob/a94e6e2b7102f18e4b9aa08c623357a79ac62486/.buildkite/pipeline.yml#L146 which has the token encoded somehow.

KristofferC:
Seems to work now.

gnome-shell-screenshot-dbvnlb

PR #1153, opened 2024-12-21
don't use reducedim_initarray

adienes:
to be applied after / on top of https://github.com/JuliaLang/julia/pull/55318


dkarrasch:
IIUC then this is outdated, right? Can we close this?

adienes:
the successor, https://github.com/JuliaLang/julia/pull/58418, would still deprecate `reducedim_initarray` so there would still be action required in LA.jl, but that PR is still slightly a moving target and it's possible it will end up including a more "canonical" migration than the one I suggested here, so I'm happy to close this until that PR is closer to finalizing

PR #1163, opened 2025-01-03
evalpoly for matrix polynomials

### LINKED ISSUE #1161: `evalpoly` is broken for matrices
```julia-repl
help?> evalpoly
evalpoly(x, p)

Evaluate the polynomial \sum_k x^{k-1} p[k] ...

julia> _evalpoly(x, p) = sum(x^(k-1)*p[k] for k in eachindex(p))
_evalpoly (generic function with 1 method)

julia> x = 10*one(zeros(2,2))
2×2 Matrix{Float64}:
 10.0   0.0
  0.0  10.0

julia> p = [1,2,3]
3-element Vector{Int64}:
 1
 2
 3

julia> _evalpoly(x, p)
2×2 Matrix{Float64}:
 321.0    0.0
   0.0  321.0

julia> evalpoly(x, p)
ERROR: MethodError: no method matching +(::Matrix{Float64}, ::Int64)
For element-wise addition, use broadcasting with dot syntax: array .+ scalar
The function `+` exists, but no method is defined for this combination of argument types.

Closest candidates are:
  +(::Any, ::Any, ::Any, ::Any...)
   @ Base operators.jl:596
  +(::BigFloat, ::Union{Int16, Int32, Int64, Int8})
   @ Base mpfr.jl:506
  +(::Base.CoreLogging.LogLevel, ::Integer)
   @ Base logging/logging.jl:132
  ...

Stacktrace:
 [1] muladd(x::Matrix{Float64}, y::Int64, z::Int64)
   @ Base.Math ./math.jl:1470
 [2] _evalpoly
   @ ./math.jl:114 [inlined]
 [3] evalpoly(x::Matrix{Float64}, p::Vector{Int64})
   @ Base.Math ./math.jl:107
 [4] top-level scope
   @ REPL[27]:1
```

stevengj:
Closes #1161.

Includes a generic out-of-place fallback implementation as well as an in-place `evalpoly!` method; the latter is called by default only for `X::StridedMatrix{<:Number}` for now, to be conservative.   (e.g. it's much trickier to allocate the in-place buffers correctly for matrices of matrices, which we don't support very well, and I'm not sure this would work well for sparse matrices.)  More optimized cases can always be added later.


mcabbott:
In addition to `X::StridedMatrix{<:Number}`, doesn't the in-place version need to widen based on the coefficients' types? Some failure cases:
```julia
julia> evalpoly([1 2; 3 4], (5, 6))  # error without this PR
2×2 Matrix{Int64}:
 11  12
 18  29

julia> evalpoly([1 2; 3 4], (5, 6.7))
ERROR: InexactError: Int64(6.7)
Stacktrace:
...
 [11] evalpoly!(Y::Matrix{Int64}, X::Matrix{Int64}, p::Tuple{Int64, Float64})
    @ Main ./REPL[4]:19

julia> evalpoly([1.0 2.0; 3.0 4.0], (2, 3+im, 4))
ERROR: InexactError: Float64(7.0 + 1.0im)
```

stevengj:
Argh, this is a PITA to get right for heterogeneous tuples of coefficients, especially for a tuple of dimensionful coefficients.  (e.g. suppose that `X` is in meters, and the coefficients `p` are in `(s, s/m, s/m^2)` — this is dimensionfully correct with an `evalpoly` result in seconds, but you can't simply promote the elements of `p`).

What is a good way to compute the correct element type of the result matrix?

For `p` a tuple of numbers, I'm tempted to do `typeof(evalpoly(zero(eltype(X)), p))`, but that will fail if `eltype(X)` is abstract.

One option would be to just punt on `evalpoly!` for now.  Another option would be to only support homogeneous tuples.

mcabbott:
Ah now I see some logic for `p[begin]` which my examples happened to avoid.

Only accepting `p::NTuple` (and concretely typed vectors) seems fine really. And perhaps even `StridedMatrix{<:AbstractFloat}` with real coefficients (to avoid my integer example), sim. complex with real-or-complex? 

Easy to add methods to send your own fancy types to `evalpoly!` if you want something not covered by this.

stevengj:
I tightened up the type signatures for dispatch to `evalpoly!` and it fixes your examples, along with abstractly typed arrays.  (Also added tests.)

mcabbott:
This is a little harder to fool, but:
```julia
julia> evalpoly([1 2; 3 4], Number[5.5, 6])
ERROR: InexactError: Int64(11.5)

julia> evalpoly([1.0 2.0; 3.0 4.0], Number[2, 3, 4+im])
ERROR: InexactError: Float64(4.0 + 1.0im)
```

LilithHafner:
I think it makes sense to dispatch to `_evalpoly` in the case of vectors with non-concrete eltype. We could do a reducion with promotion but I don't think it's worth trying particularly hard to optimize that case.

src/evalpoly.jl
```diff
@@ -0,0 +1,85 @@
+# matrix methods for evalpoly(X, p) = ∑ₖ Xᵏ⁻¹ p[k]
+
+# non-inplace fallback for evalpoly(X, p)
+function _evalpoly(X::AbstractMatrix, p)
+    Base.require_one_based_indexing(p)
+    p0 = isempty(p) ? Base.reduce_empty_iter(+, p) : p[end]
+    Xone = one(X)
+    S = Base.promote_op(*, typeof(Xone), typeof(Xone))(Xone) * p0
```
martinholters: Should this be
```suggestion
    S = convert(Base.promote_op(*, typeof(Xone), typeof(Xone)), Xone) * p0
```
to avoid unnecessary allocation of a new matrix in the likely event that `Xone isa Base.promote_op(*, typeof(Xone), typeof(Xone))`?

That aside, I'm not a fan of `promote_op` in general. What's the benefit here anyway?

stevengj: What would you use instead of `promote_op`?

martinholters: I'm missing background here to answer that. To start: what's wrong with just `S = Xone * p0`?

dkarrasch: I guess you want to make sure that a `Bool` eltype is promoted to `Int`:

```julia
julia> X = rand(Bool, 2, 2)
2×2 Matrix{Bool}:
 0  0
 0  1

julia> one(X)
2×2 Matrix{Bool}:
 1  0
 0  1

julia> one(X)^2
2×2 Matrix{Int64}:
 1  0
 0  1
```

And I agree about using `convert`.

ViralBShah:
Is this good to merge (assuming tests pass)?

LilithHafner:
@mcabbott's [comment](https://github.com/JuliaLang/LinearAlgebra.jl/pull/1163#issuecomment-2569992351) is still not addressed. That said, those cases already error.

ViralBShah:
Would it be better to merge this if it is at least addressing some of the issues?

LilithHafner:
CI failure is in matmul so probably unrelated. If so, I think it's likely reasonable to merge this even with @mcabbott's unresolved comment. Do you agree, @mcabbott?

mcabbott:
I have no strong feelings... my examples above invented to break this do seem unlikely to show up in the wild.

ViralBShah:
@stevengj We should have merged this earlier, but it would be great if you can update the PR and get it in.

PR #1169, opened 2025-01-14
Let `cond` of an empty square matrix return zero

### LINKED ISSUE #778: `cond` fails for null dimensions
The following command fails (the result should be probably 0, as in MATLAB):
```

cond(ones(0,0)) 

ERROR: ArgumentError: reducing over an empty collection is not allowed
Stacktrace:
 [1] _empty_reduce_error() at .\reduce.jl:299
 [2] reduce_empty(::Function, ::Type{Float64}) at .\reduce.jl:309
 [3] mapreduce_empty(::typeof(identity), ::Function, ::Type{T} where T) at .\reduce.jl:343
 [4] reduce_empty(::Base.MappingRF{typeof(identity),typeof(max)}, ::Type{Float64}) at .\reduce.jl:329
 [5] reduce_empty_iter at .\reduce.jl:355 [inlined]
 [6] mapreduce_empty_iter(::Function, ::Function, ::Array{Float64,1}, ::Base.HasEltype) at .\reduce.jl:351
 [7] _mapreduce(::typeof(identity), ::typeof(max), ::IndexLinear, ::Array{Float64,1}) at .\reduce.jl:400
 [8] _mapreduce_dim at .\reducedim.jl:318 [inlined]
 [9] #mapreduce#620 at .\reducedim.jl:310 [inlined]
 [10] mapreduce at .\reducedim.jl:310 [inlined]
 [11] _maximum at .\reducedim.jl:727 [inlined]
 [12] _maximum at .\reducedim.jl:726 [inlined]
 [13] #maximum#631 at .\reducedim.jl:722 [inlined]
 [14] maximum at .\reducedim.jl:722 [inlined]
 [15] cond(::Array{Float64,2}, ::Int64) at D:\buildbot\worker\package_win64\build\usr\share\julia\stdlib\v1.5\LinearAlgebra\src\dense.jl:1431
 [16] cond(::Array{Float64,2}) at D:\buildbot\worker\package_win64\build\usr\share\julia\stdlib\v1.5\LinearAlgebra\src\dense.jl:1429
 [17] top-level scope at REPL[11]:1
```

martinholters:
Transfer of https://github.com/JuliaLang/julia/pull/38372. Summary of the discussion there:
We have this inconsistency:
```julia
julia> cond(ones(0,0), 1)
0.0

julia> cond(ones(0,0), 2)
ERROR: ArgumentError: reducing over an empty collection is not allowed; consider supplying `init` to the reducer

julia> cond(ones(0,0), Inf)
0.0
```
If it wasn't breaking, making all cases throw would be acceptable behavior. This PR goes the other way, making the `p==2` case also return zero. Although met with some skepticism, if there is to be chosen a value, zero is probably it.

I think the main question is whether the inconsistency or the somewhat dubious return-value is the lesser evil. Mainly porting this over as a reminder to reach a decision. Either way we should probably close #778; either by resolving it with this PR or as won't-fix.

Fixes #778.

### DIFF
--- a/src/dense.jl
+++ b/src/dense.jl
@@ -1779,6 +1779,10 @@ Condition number of the matrix `M`, computed using the operator `p`-norm. Valid 
 """
 function cond(A::AbstractMatrix, p::Real=2)
     if p == 2
+        if isempty(A)
+            checksquare(A)
+            return zero(real(eigtype(eltype(A))))
+        end
         v = svdvals(A)
         maxv = maximum(v)
         return iszero(maxv) ? oftype(real(maxv), Inf) : maxv / minimum(v)
--- a/test/dense.jl
+++ b/test/dense.jl
@@ -57,23 +57,16 @@ Random.seed!(1234323)
         @test cond(Mars, Inf) ≈ 7.1
     end
     @testset "Empty matrices" begin
-        # zero for square (i.e. 0×0) matrices
-        @test cond(zeros(Int, 0, 0), 1) === 0.0
-        @test cond(zeros(Int, 0, 0), 2) === 0.0
-        @test cond(zeros(Int, 0, 0), Inf) === 0.0
-        @test cond(zeros(0, 0), 1) === 0.0
-        @test cond(zeros(0, 0), 2) === 0.0
-        @test cond(zeros(0, 0), Inf) === 0.0
-        @test cond(zeros(ComplexF64, 0, 0), 1) === 0.0
-        @test cond(zeros(ComplexF64, 0, 0), 2) === 0.0
-        @test cond(zeros(ComplexF64, 0, 0), Inf) === 0.0
-        # error for non-square matrices
-        @test_throws DimensionMismatch cond(zeros(10, 0), 1)
-        @test_throws DimensionMismatch cond(zeros(0, 10), 1)
-        @test_throws DimensionMismatch cond(zeros(10, 0), 2)
-        @test_throws DimensionMismatch cond(zeros(0, 10), 2)
-        @test_throws DimensionMismatch cond(zeros(10, 0), Inf)
-        @test_throws DimensionMismatch cond(zeros(0, 10), Inf)
+        for norm in (1,2,Inf)
+            # zero for square (i.e. 0×0) matrices
+            @test cond(zeros(Int, 0, 0), norm) === 0.0
+            @test cond(zeros(0, 0), norm) === 0.0
+            @test cond(zeros(ComplexF64, 0, 0), norm) === 0.0
+            # error for non-square matrices
+            for size in ((10,0), (0,10))
+                @test_throws DimensionMismatch cond(zeros(size...), norm)
+            end
+        end
     end
 end
 
--- a/test/dense.jl
+++ b/test/dense.jl
@@ -56,6 +56,25 @@ Random.seed!(1234323)
         @test cond(Mars, 2)   ≈ 6.181867355918493
         @test cond(Mars, Inf) ≈ 7.1
     end
+    @testset "Empty matrices" begin
+        # zero for square (i.e. 0×0) matrices
+        @test cond(zeros(Int, 0, 0), 1) === 0.0
+        @test cond(zeros(Int, 0, 0), 2) === 0.0
+        @test cond(zeros(Int, 0, 0), Inf) === 0.0
+        @test cond(zeros(0, 0), 1) === 0.0
+        @test cond(zeros(0, 0), 2) === 0.0
+        @test cond(zeros(0, 0), Inf) === 0.0
+        @test cond(zeros(ComplexF64, 0, 0), 1) === 0.0
+        @test cond(zeros(ComplexF64, 0, 0), 2) === 0.0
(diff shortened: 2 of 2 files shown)


stevengj:
Thinking about this a bit more, this seems sensible, at least for **square** empty matrices.  For square matrices, we define the condition number as $\Vert A \Vert \cdot \Vert A^{-1} \Vert$.  For an empty matrix, the norm should be zero (because it's the additive identity), and the inverse of a *square* empty matrix is itself (since $Ax = b$ has a unique solution: an empty $b$ is unique, and empty $x$ solves it).

Returning 0 is also [what Matlab does](https://www.mathworks.com/matlabcentral/answers/1744315-condition-number-of-empty-matrix), it seems.

src/dense.jl
```diff
@@ -1779,6 +1779,7 @@ Condition number of the matrix `M`, computed using the operator `p`-norm. Valid
 """
 function cond(A::AbstractMatrix, p::Real=2)
     if p == 2
+        isempty(A) && return zero(real(eigtype(eltype(A))))
```
```suggestion
        if isempty(A)
            checksquare(A)
            return zero(real(eigtype(eltype(A))))
        end
```

stevengj: (Note that the `p == 1 || p == Inf` branch also calls `checksquare`).

martinholters:
Agree, restricting to square `A` is reasonable.

stevengj:
Would be good to have a `@test_throws` to ensure it throws for non-square empty matrices.

test/dense.jl
```diff
@@ -56,6 +56,25 @@ Random.seed!(1234323)
...
+        @test cond(zeros(ComplexF64, 0, 0), 1) === 0.0
+        @test cond(zeros(ComplexF64, 0, 0), 2) === 0.0
+        @test cond(zeros(ComplexF64, 0, 0), Inf) === 0.0
+        # error for non-square matrices
+        @test_throws DimensionMismatch cond(zeros(10, 0), 1)
+        @test_throws DimensionMismatch cond(zeros(0, 10), 1)
+        @test_throws DimensionMismatch cond(zeros(10, 0), 2)
+        @test_throws DimensionMismatch cond(zeros(0, 10), 2)
+        @test_throws DimensionMismatch cond(zeros(10, 0), Inf)
+        @test_throws DimensionMismatch cond(zeros(0, 10), Inf)
```
stevengj, line 76:
```suggestion
        for size in ((10,0), (0,10)), norm in (1,2,Inf)
            @test_throws DimensionMismatch cond(zeros(size...), norm)
        end
```

stevengj:
```suggestion
        for norm in (1,2,Inf)
            # zero for square (i.e. 0×0) matrices
            @test cond(zeros(Int, 0, 0), norm) === 0.0
            @test cond(zeros(0, 0), norm) === 0.0
            @test cond(zeros(ComplexF64, 0, 0), norm) === 0.0
            # error for non-square matrices
            for size in ((10,0), (0,10))
                @test_throws DimensionMismatch cond(zeros(size...), norm)
            end
        end
```

martinholters: Sure, why not.

test/dense.jl
```diff
@@ -56,6 +56,18 @@ Random.seed!(1234323)
...
     end
+    @testset "Empty matrices" begin
+        for norm in (1,2,Inf)
+            # zero for square (i.e. 0×0) matrices
+            @test cond(zeros(Int, 0, 0), norm) === 0.0
+            @test cond(zeros(0, 0), norm) === 0.0
+            @test cond(zeros(ComplexF64, 0, 0), norm) === 0.0
+            # error for non-square matrices
+            for size in ((10,0), (0,10))
+                @test_throws DimensionMismatch cond(zeros(size...), norm)
```
martinholters:
```suggestion
        for p in (1,2,Inf)
            # zero for square (i.e. 0×0) matrices
            @test cond(zeros(Int, 0, 0), p) === 0.0
            @test cond(zeros(0, 0), p) === 0.0
            @test cond(zeros(ComplexF64, 0, 0), p) === 0.0
            # error for non-square matrices
            for size in ((10,0), (0,10))
                @test_throws DimensionMismatch cond(zeros(size...), p)
```
Making the variable match the parameter name might be slightly more readable.

stevengj (approved):

Jutho:
I am trying to understand the argumentation here. I see zero was the least breaking change. But in principle, for any consistent norm, we have
$$\kappa(A) = \lVert A\rVert \lVert A^{-1}\rVert \geq \lVert A A^{-1}\rVert = 1$$,
i.e. $$\kappa(A)$$ lies in the range $$[+1, +\infty]$$. Instead of saying that the empty matrix is the additive identity, one could also say that it is the multiplicative identity, which I think makes much more sense in the setting where one uses condition numbers. Indeed, as pointed out by @stevengj , $$A * x = b$$ where $$A$$ and $$b$$ are empty has a unique solution, namely also $$x$$ is empty or thus $$x = b$$ or thus $$A$$ is the (multiplicative) identity, and thus I would expect condition number $$\kappa(A)=1$$.

Jutho:
Never mind, reading the Matlab discussion linked to above made things a bit clearer. My apologies for the noise.

PR #1171, opened 2025-01-16
fix error messages

araujoms:
The code is doing the right thing, I'm changing the error messages to match.


dkarrasch (approved):

dkarrasch:
Thanks, @araujoms!

PR #1186, opened 2025-01-29
remove copy-allocation on accessing `cholesky` factors (`.L`, `.U`)

st--:
(Port of https://github.com/JuliaLang/julia/pull/42920)

In master, accessing fields L or U of Cholesky or CholeskyPivoted calls copy() on the transpose. Here I add tests thayt check that the accessors don't allocate more memory than is needed for e.g. the Adjoint() construction, and remove the unnecessary copy(). Generally julia seems to be trying hard to avoid unnecessary allocations, so I would consider this to be a bugfix, not a breaking change.

There is a related discourse thread: https://discourse.julialang.org/t/implementation-for-accessing-cholesky-factors/56219

Side note 1: there is also a copy() in the definition of _chol!(A::AbstractMatrix, ::Type{UpperTriangular}) - it's a bit less clear to me in what cases this will get run and whether it's okay to remove the copy(), so let me know if you think it'd be worth spending the time to go deeper into that also.

Side note 2: some of the code tests uplo using Cuplo === char_uplo(d), some of it using sym_uplo(Cuplo) == d, is there any particular reason for this, or would it be worth unifying it?

### DIFF
--- a/src/cholesky.jl
+++ b/src/cholesky.jl
@@ -65,7 +65,7 @@ julia> C.U
   ⋅    ⋅    3.0
 
 julia> C.L
-3×3 LowerTriangular{Float64, Matrix{Float64}}:
+3×3 LowerTriangular{Float64, Adjoint{Matrix{Float64}}}:
   2.0   ⋅    ⋅
   6.0  1.0   ⋅
  -8.0  5.0  3.0
@@ -530,7 +530,7 @@ julia> C.U
   ⋅    ⋅    3.0
 
 julia> C.L
-3×3 LowerTriangular{Float64, Matrix{Float64}}:
+3×3 LowerTriangular{Float64, Adjoint{Matrix{Float64}}}:
   2.0   ⋅    ⋅
   6.0  1.0   ⋅
  -8.0  5.0  3.0
@@ -668,14 +668,14 @@ function _choleskyUfactor(Cfactors, Cuplo)
     if Cuplo === 'U'
         return UpperTriangular(Cfactors)
     else
-        return copy(LowerTriangular(Cfactors)')
+        return LowerTriangular(Cfactors)'
     end
 end
 function _choleskyLfactor(Cfactors, Cuplo)
     if Cuplo === 'L'
         return LowerTriangular(Cfactors)
     else
-        return copy(UpperTriangular(Cfactors)')
+        return UpperTriangular(Cfactors)'
     end
 end
 
--- a/test/cholesky.jl
+++ b/test/cholesky.jl
@@ -595,9 +595,9 @@ end
     B = cholesky(A)
     B32 = cholesky(Float32.(A))
     @test B isa Cholesky{Float16, Matrix{Float16}}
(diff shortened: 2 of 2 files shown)


stevengj:
Right now this leads to slightly slower triangular solves:
```jl
julia> using BenchmarkTools, LinearAlgebra

julia> A = rand(1000,1000); b = rand(1000); x = copy(b);

julia> U1 = LowerTriangular(A)';

julia> U2 = copy(U1);

julia> @btime ldiv!($x, $U1, $b);
  81.000 μs (0 allocations: 0 bytes)

julia> @btime ldiv!($x, $U2, $b);
  69.875 μs (0 allocations: 0 bytes)
```
although it still might be worth it to save allocations.  (Presumably the `ldiv!` performance can be fixed by an optimized method if needed.)

Not sure what other code might be affected by this … in general, it seems okay since (a) the documentation does not specify whether a copy is made and (b) it is still returning a subtype of `UpperTriangular` or `LowerTriangular`.

ViralBShah:
Test failures need to be fixed.

dkarrasch:
> Right now this leads to slightly slower triangular solves:

This is just a matter of memory access, I think. The solves all run LAPACK's `trtrs!`.

dkarrasch:
... or BLAS: #1194.

st--:
Great, thanks for accepting my PR!

dkarrasch:
Thanks for making it!

src/cholesky.jl
```diff
@@ -668,14 +668,14 @@ function _choleskyUfactor(Cfactors, Cuplo)
     if Cuplo === 'U'
         return UpperTriangular(Cfactors)
     else
-        return copy(LowerTriangular(Cfactors)')
+        return LowerTriangular(Cfactors)'
```
devmotion, line 671: I was slightly confused when reading this the first time, IMO from
```suggestion
        return UpperTriangular(Cfactors')
```
it would have been more obvious that in both branches an `UpperTriangular` is returned.

dkarrasch: Yes, we can do that as well.

ViralBShah: We should get this one in too.

ViralBShah: Won't this create a copy?

ViralBShah:
Initially I thought this was a nice memory savings gain (even though at the cost of some performance).

As shown in the doctest, I suspect that this may be more breaking than imagined. Perhaps won't actually break code, but the return type changes (which now has an Adjoint in it) which may lead to unpredictable performance changes or in some cases even breakage.

There's also now an asymmetry in the types for the L and the U, which I don't like.

src/cholesky.jl
dkarrasch: No, the `UpperTriangular` just wraps the lazy adjoint.

ViralBShah:
I am wondering if we should temporarily revert this PR so that the other Base.import stuff can get into Julia master, and then figure out exactly what we want to do and also fix the doctest etc.

@KristofferC

PR #1188, opened 2025-01-31
update manifest from JuliaSyntaxHighlighting UUID change

KristofferC:
Ref https://github.com/JuliaLang/julia/pull/56852


KristofferC:
Windows nightly doesn't seem to be updated.

PR #1190, opened 2025-02-02
More resizing for truncating return values from LAPACK

BioTurboNick:
Found additional locations that could benefit from the changes in #1176 , but I haven't tested these as yet.

### DIFF
--- a/src/lapack.jl
+++ b/src/lapack.jl
@@ -4090,7 +4090,7 @@ for (stev, stebz, stegr, stein, elty) in
                     resize!(iwork, liwork)
                 end
             end
-            Zm = min(size(Z, 2), m[])
+            Zm = min(Zn, m[])
             resize!(w, m[]), reshape(resize!(Z, ldz * Zm), ldz, Zm)
         end
 
--- a/src/lapack.jl
+++ b/src/lapack.jl
@@ -3875,6 +3875,9 @@ for (trcon, trevc, trrfs, elty, relty) in
                 work, rwork, info, 1, 1)
             chklapackerror(info[])
 
+            VLn = size(VL, 1)
+            VRn = size(VR, 1)
+            #Decide what exactly to return
             if howmny == 'S' #compute selected eigenvectors
                 if side == 'L' #left eigenvectors only
                     return select, reshape(resize!(vec(VL), VLn * m[]), VLn, m[])
--- a/src/lapack.jl
+++ b/src/lapack.jl
@@ -4090,7 +4090,8 @@ for (stev, stebz, stegr, stein, elty) in
                     resize!(iwork, liwork)
                 end
             end
-            resize!(w, m[]), reshape(resize!(Z, ldz * m[]), ldz, m[])
+            Zm = min(size(Z, 2), m[])
+            resize!(w, m[]), reshape(resize!(Z, ldz * Zm), ldz, Zm)
         end
 
         function stein!(dv::AbstractVector{$elty}, ev_in::AbstractVector{$elty}, w_in::AbstractVector{$elty}, iblock_in::AbstractVector{BlasInt}, isplit_in::AbstractVector{BlasInt})
--- a/src/lapack.jl
+++ b/src/lapack.jl
@@ -2835,7 +2835,7 @@ for (orglq, orgqr, orgql, orgrq, ormlq, ormqr, ormql, ormrq, gemqrt, elty) in
                 end
             end
             if n < size(A,2)
-                A[:,1:n]
+                reshape(resize!(vec(A), m * n), m, n)
             else
                 A
             end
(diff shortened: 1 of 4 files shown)


src/lapack.jl
```diff
@@ -3873,22 +3875,21 @@ for (trcon, trevc, trrfs, elty, relty) in
                 work, rwork, info, 1, 1)
             chklapackerror(info[])
 
-            #Decide what exactly to return
```
dkarrasch, line 3876: This needs

```julia
            VLn = size(VL, 1)
            VRn = size(VR, 1)
```

from what I understand.

BioTurboNick:
More to come. ~These two don't look all *that* much like an improvement... I wonder if it's because the wrapping to vec and then back to a matrix isn't that helpful. Can't create it as a vector because it's what's provided as input to the function...~

EDIT 2/12/25: Determined I wasn't using the right inputs to test the new path. Using a 100x110 array to test, which triggers the truncation path:
```
using Random
Random.seed!(100)
A,tau = LAPACK.geqrf!(rand(elty,100,110))
function f(A, tau)
    for i = 1:100
        LAPACK.orgqr!(A,tau)
    end
end
@benchmark f(A, tau) seconds = 60

Random.seed!(100)
A,tau = LAPACK.geqlf!(rand(elty,100,110))
function f(A, tau)
    for i = 1:100
        LAPACK.orgql!(A,tau)
    end
end
@benchmark f(A, tau) seconds = 60
```

`orgqr!` Before:
image
After:
image


`orgql!` Before:
image
After:
image

(more to come)

dkarrasch:
Test failures may be real?

ViralBShah:
@BioTurboNick Looks like a nice PR but seems to break some tests.

BioTurboNick:
Thanks! I'm mystified why it would work on `linux-i686` and `macos-aarch64` but not `linux-x86_64` or `windows-x86_64`, though it's suggestive.

Linux segfaults and isn't helpful in figuring out the cause, but Windows provides a `ReadOnlyMemoryError` that is converted to a Julia exception, with a nice stack trace.

The proximate cause is in `reshape(resize!(Z, ldz * m[]), ldz, m[])` at the end of `stegr!`. Nearest I can figure is that `m[]` can sometimes be larger than the second dimension of `Z` or of the length of `w`, which would explain why the conditional was needed in the original code. Let's see if that fixes it...

But still, that should only produce garbage values in the extra part, not a segfault? And why only on x86_64?

BioTurboNick:
Oy. Either the debug code caused it to go away, or the change in nightly from DEV.33 to DEV.38 did it. EDIT: Nope, not fixed by the newer nightlies. It's a Heisenbug.

BioTurboNick:
Finally, a minimal reproducer:
```
a = SymTridiagonal(rand(12), rand(11)) # SymTridiagonal([1.0129540363709792, 0.5125544740879742, 1.4404024510136622, 0.9645500736079791, 0.8349894041911953, 1.0404785847552251, 0.9418430974856163, 0.1401583600348122, 0.38426770335791244, 0.4551768643291618, 0.4200328536585094, 1.8593998829530265e-7], [0.40739613651353346, 0.10643163002740841, 0.6222058356713083, 0.33352809216796453, 0.3799591152214755, 0.21073828255014115, 0.8158852900425236, 0.1683998539581063, 0.5002702262460241, 0.6123852845929411, 0.7701699043551441])"
A = copy(a); LAPACK.stegr!('V', 'A', A.dv, A.ev, 0.0, 0.0, 0, 0)
```

It seems that this is a bug in Julia (1.11 and nightly)? Seems to be producing undefined behavior. With various configurations of the end code I get a Julia crash with `EXCEPTION_ACCESS_VIOLATION`, or a Julia-handled `ReadOnlyMemoryError()`. And this is happening when the size actually isn't changing.

It seems some necessary ingredients are 1) initiating as a vector*; 2) passing the vector into the ccall; 3) reshaping the vector into a matrix (resize call not necessary, though can independently trigger a similar error).

If the array is initiated as a matrix, and then later reshaped with `reshape(resize!(vec(Z), ldz * Zm), ldz, Zm)`, there's no problem at all.

And the issue goes away entirely if the offending line is wrapped in try/catch.

*As far as C is concerned, an array is just a pointer to the start of a chunk of memory, so should make no difference on the C side of things?

BioTurboNick:
Well, that was dumb. Changed it to a vector, didn't change the line that relied on looking up its second dimension.

dkarrasch:
IIUC, you solved the issue, right? So, there is no need to catch anything in the tests? I wasn't able to run tests locally, so I removed the try-catch-block and added a simple test to cover what was already missed before your PR. If tests pass, then I think this should be ready to go.

BioTurboNick:
Ah thanks, yeah good to go.

PR #1191, opened 2025-02-03
Materialize adjoint in mul with `HermOrSym`

### LINKED ISSUE #850: Multiplying `Adjoint` with `Symmetric` uses slow generic matmul
Multiplying an `Adjoint` with a `Symmetric` matrix falls back to a very slow generic matmul.
```julia
n = 500
A = randn(n, n)
Q = Symmetric(randn(n,n))

@btime $A'*$Q

```

```
277.406 ms (2 allocations: 1.91 MiB)
```

Converting `A'` to `Matrix` makes it a lot faster
```julia
@btime Matrix($A')*$Q
```
```
  4.281 ms (4 allocations: 3.81 MiB)
```

The break-even poirnt for when converting `A'` to `Matrix` is faster is `n = 5`.

This seems a bit slow? Should this be fixed in `Base` or should one convert to `Matrix` explicitly?

Perhaps  fallback conversion to `Matrix` should be used more widely when `eltype(A) <: BlasFloat` ?

dkarrasch:
Fixes #850, at least the original issue. Promoting `Bool` to some `BlasFloat` seems to be very tricky in terms method ambiguities.

### DIFF
--- a/src/symmetric.jl
+++ b/src/symmetric.jl
@@ -703,6 +703,19 @@ for f in (:+, :-)
 end
 
 *(A::HermOrSym, B::HermOrSym) = A * copyto!(similar(parent(B)), B)
+# catch a few potential BLAS-cases
+function *(A::HermOrSym{<:BlasFloat,<:StridedMatrix}, B::AdjOrTrans{<:BlasFloat,<:StridedMatrix})
+    T = promote_type(eltype(A), eltype(B))
+    mul!(similar(B, T, (size(A, 1), size(B, 2))),
+            convert(AbstractMatrix{T}, A),
+            copy_oftype(B, T)) # make sure the AdjOrTrans wrapper is resolved
+end
+function *(A::AdjOrTrans{<:BlasFloat,<:StridedMatrix}, B::HermOrSym{<:BlasFloat,<:StridedMatrix})
+    T = promote_type(eltype(A), eltype(B))
+    mul!(similar(B, T, (size(A, 1), size(B, 2))),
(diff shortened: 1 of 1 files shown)


dkarrasch:
x-ref https://github.com/JuliaLang/julia/pull/52787 for the Bool case.

araujoms:
Why not implement @mcabbott 's suggestion, though? It is faster for small matrices and avoids a copy.

dkarrasch:
We seem to avoid wrapping results in `adjoint` or `transpose`, or, better said, we materialize the final wrapper:

```julia
A' * Q = copy((Q' * A)')
```

We return wrapped results only when the first factor is an adjoint vector.

araujoms:
Why, though? I don't see the problem with wrapping results in `adjoint` or `transpose`.

dkarrasch:
@andreasnoack Do you remember why we typically materialize products that we compute by the adjoint formula, except for `AdjOrTransAbsVec`?

andreasnoack:
I don't remember the specifics here. Only that the wrapped versions can sometimes lead to performance issues because they wrappers sometimes causes dispatch to slow fallback definitions for `AbstractMatrix`.

dkarrasch:
Another reason for going with this as is is that this currently returns a `Matrix`, and not an `Adjoint`. I suggest we go with it, and if we consider returning `Adjoint`s, we should do this in several other places consistently, and we would need to evaluate the effect on the ecosystem. Perhaps now is a good idea, early in the release cycle, or would this be generally regarded as "breaking", because it changes return types?

PR #1192, opened 2025-02-03
Fix `LAPACK.posv!` tests

dkarrasch:
The test failures are, unsurprisingly, related to high condition numbers, as can be seen from running the tests many times in a loop and displaying the condition number. We could set an appropriate relative tolerance in the test below, or find a random seed that works. I tried the second option here.

### DIFF
--- a/test/lapack.jl
+++ b/test/lapack.jl
@@ -676,6 +676,7 @@ end
 @testset "posv and some errors for friends" begin
     @testset for elty in (Float32, Float64, ComplexF32, ComplexF64)
         local n = 10
+        Random.seed!(123)
         A = randn(elty,n,n)
         A = A'*A
         B = rand(elty,n,n)
--- a/test/lapack.jl
+++ b/test/lapack.jl
@@ -676,14 +676,13 @@ end
 @testset "posv and some errors for friends" begin
     @testset for elty in (Float32, Float64, ComplexF32, ComplexF64)
         local n = 10
-        Random.seed!(123)
-        A = randn(elty,n,n)
-        A = A'*A
-        B = rand(elty,n,n)
+        a = randn(elty, n, n)
+        A = a'*a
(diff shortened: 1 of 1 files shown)


ViralBShah:
The condition number of the `randn` generated matrices jumps around a fair bit - no doubt because we are squaring. The scaling should greatly help.

alanedelman:
yehh randn when squared will produce condition numbers of order n^2, von Neumann fell into this trap decades ago, leading to the belief that finite precision could be dreadful on these new machines that were being called computers.
Of course you can generate Q'DQ with any positive diagonal D you like and a random Q.  If speed is a concern we can be more clever, and LAPACK probably has such routines, but quick and dirty is to use `Q=Matrix(qr(randn(n,n)).Q)`
which is O(n^3) (there are O(n^2) ways)

ViralBShah:
```
julia> for i=1:10 
            Q=Matrix(qr(randn(n,n)).Q);
            @show cond(Q'*Diagonal(rand(n))*Q)
       end
cond(Q' * Diagonal(rand(n)) * Q) = 54.84525143924938
cond(Q' * Diagonal(rand(n)) * Q) = 68.73534833441128
cond(Q' * Diagonal(rand(n)) * Q) = 96.3033669626725
cond(Q' * Diagonal(rand(n)) * Q) = 87.56679319167324
cond(Q' * Diagonal(rand(n)) * Q) = 66255.80209047809
cond(Q' * Diagonal(rand(n)) * Q) = 41.302179179853
cond(Q' * Diagonal(rand(n)) * Q) = 45.178741706939526
cond(Q' * Diagonal(rand(n)) * Q) = 164.87644137765756
cond(Q' * Diagonal(rand(n)) * Q) = 93.56871249448481
cond(Q' * Diagonal(rand(n)) * Q) = 257.1891863712371

```

PR #1194, opened 2025-02-04
Use `BLAS.trsm!` instead of `LAPACK.trtrs!` in left-triangular solves

dkarrasch:
We have used that in right-triangular solves already.

### DIFF
--- a/src/blas.jl
+++ b/src/blas.jl
@@ -1369,6 +1369,11 @@ for (fname, elty) in ((:dtrsv_,:Float64),
                 throw(DimensionMismatch(lazy"size of A is $n != length(x) = $(length(x))"))
             end
             chkstride1(A)
+            if diag == 'N'
+                for i in 1:n
+                    iszero(A[i,i]) && throw(SingularException(i))
+                end
+            end
             px, stx = vec_pointer_stride(x, ArgumentError("input vector with 0 stride is not allowed"))
             GC.@preserve x ccall((@blasfunc($fname), libblastrampoline), Cvoid,
                 (Ref{UInt8}, Ref{UInt8}, Ref{UInt8}, Ref{BlasInt},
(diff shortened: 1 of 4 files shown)


dkarrasch:
@nanosoldier `runbenchmarks("linalg"; vs=":master")`

dkarrasch:
In a manual 1000x1000 test, I see slight to no improvement, but certainly no regression.

dkarrasch:
It turns out that the LAPACK function is calling the BLAS function internally (https://netlib.org/lapack/explore-html/d4/dc1/group__trtrs_gab0b6a7438a7eb98fe2ab28e6c4d84b21.html), but checks for exact singularity upfront, nothing else. It seems we could rather switch to the "augmented" LAPACK functions throughout instead?

@jishnub @andreasnoack @amontoison

dkarrasch:
Nevermind, there is no triangular right-solve in LAPACK.

amontoison:
@dkarrasch I was thinking of LBT if we switch the backend, like BLIS. Sometimes, we only have the BLAS routine available.

dkarrasch:
Ok, we can of course switch, but then we should perhaps add the exact singularity check?

amontoison:
Yes, and we can do that for both `trsv!` and `trsm!` such that it is consistent.
Using `trsv!` for one right-hand and `trtrs!` for multiple ones have a different behaviour if the triangular matrix is singular.

amontoison (approved):

vtjnash:
This appears to have introduced a bug into the benchmarking tests (I cannot vouch for their quality, only that this commit seems to have introduced a regression in them): https://github.com/JuliaLang/julia/commit/c8ed17f7b4c148a887d8eb34babe529dd6b541a7#commitcomment-153581870

PR #1197, opened 2025-02-08
Simplify `getproperty` for `Cholesky*`

dkarrasch:


stevengj (approved):

stevengj:
LGTM.  (Seems like both previous and new versions are type-unstable, but not sure how to fix that without putting `uplo` as a type parameter.)

dkarrasch:
Yes, this is as good/bad as it always was. It just makes the outer wrapper explicit. Before, it was implicitly changed by the outer adjoint.

PR #1199, opened 2025-02-10
Fix doctest

ViralBShah:
Fix doctest seen in:

https://buildkite.com/julialang/julia-master/builds/44616#0194f191-ee57-4a81-ab55-0192700d1df9

### DIFF
--- a/src/cholesky.jl
+++ b/src/cholesky.jl
@@ -65,7 +65,7 @@ julia> C.U
   ⋅    ⋅    3.0
 
 julia> C.L
-3×3 LowerTriangular{Float64, Adjoint{Matrix{Float64}}}:
+3×3 LowerTriangular{Float64, {Matrix{Float64}}:
   2.0   ⋅    ⋅
   6.0  1.0   ⋅
  -8.0  5.0  3.0


src/cholesky.jl
```diff
@@ -65,7 +65,7 @@ julia> C.U
   ⋅    ⋅    3.0
 
 julia> C.L
-3×3 LowerTriangular{Float64, Adjoint{Matrix{Float64}}}:
+3×3 LowerTriangular{Float64, {Matrix{Float64}}:
```
KristofferC, line 68: This doesn't look right, or?  Maybe

```suggestion
3×3 LowerTriangular{Float64, Matrix{Float64}}:
```

dkarrasch: There must be a mismatch between used code and tests in the failing doctest. This has been changed deliberately lately IIRC.

ViralBShah:
Can someone look at https://buildkite.com/julialang/julia-master/builds/44616#0194f191-ee57-4a81-ab55-0192700d1df9/823-1641?

I don't see why Adjoint is in the output  for LowerTriangular in the CI, both before and after:
```
2025-02-10 16:13:20 EST | │   diff =
-- | --
  | 2025-02-10 16:13:20 EST | │    3×3 LowerTriangular{Float64, Adjoint{Matrix{Float64}}}:
  | 2025-02-10 16:13:20 EST | │      Adjoint{Float64, Matrix{Float64}}}:
  | 2025-02-10 16:13:20 EST | │      2.0   ⋅    ⋅
  | 2025-02-10 16:13:20 EST | │      6.0  1.0   ⋅
  | 2025-02-10 16:13:20 EST | │     -8.0  5.0  3.0
```

None of this got caught in the CI in this repo - and at least in my local build on `master` it seems to do what I expect, which is what the PR does (modulo manual editing and messing up - and wondering why CI passed):

dkarrasch:
#1186

src/cholesky.jl
KristofferC: My guess is that the docs don't run with the custom sysimage and are in fact testing the bundled LinearAlgebra. Could that be correct?

ViralBShah:
I still don't understand how these tests pass here but not on julia master.

ViralBShah:
Ok - the issue in this repo seems to be that the `CI/docs` workflow only generates the docs but does not run the doctests.

dkarrasch:
I quickly ran the doctest in my REPL manually, and it does show the `Adjoint`, so this PR is certainly going in the wrong direction. Should we close it and open another one on fixing the whole sysimg-stdlib-doctest-whatnot issue?

ViralBShah:
This PR is certainly incorrect. We should do the whole fix the sysimg doctest thing. But there is a separate point about whether the Adjoint is a good idea. 

I suppose fix the sysimg thing and this doctest so that we can bump LinearAlgebra on julia master once again. Afterwards, if we can decide if we want to revert #1186.

dkarrasch:
I'm closing this. @KristofferC spotted the typo in the doctest (which was truly failing) and fixed it.

PR #1235, opened 2025-03-11
Initial implementation of lazy JLLs for LinearAlgebra

staticfloat:
This alters CompilerSupportLibraries_jll, OpenBLAS_jll and libblastrampoline_jll to use `LazyLibrary` objects and thereby be loaded only upon first `dlopen()` or `ccall()` to the individual library objects.  Note that this is one of the more complicated cases, as `libblastrampoline` must have OpenBLAS_jll added as a dynamic dependency (as it does not actually have it listed in its shared object headers) and also has some on-load callbacks that must be invoked.

This must be paired with the appropriate base Julia changes [0].

[0] https://github.com/JuliaLang/julia/pull/57719

### DIFF
--- a/src/LinearAlgebra.jl
+++ b/src/LinearAlgebra.jl
@@ -820,25 +820,27 @@ function versioninfo(io::IO=stdout)
     return nothing
 end
 
-function __init__()
-    try
-        verbose = parse(Bool, get(ENV, "LBT_VERBOSE", "false"))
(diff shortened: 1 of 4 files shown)


ViralBShah:
Should this backport to 1.12

staticfloat:
Yes, assuming the base PR also gets back ported

ViralBShah:
Marked for backport to 1.12 since the Julia PR also is.

cc @KristofferC

PR #1247, opened 2025-03-26
fix dispatch to herk

araujoms:
The herk wrapper was testing whether `C` was symmetric instead of Hermitian, thus preventing the dispatch from occurring for essentially all nonzero `β`.

### DIFF
--- a/src/matmul.jl
+++ b/src/matmul.jl
@@ -692,7 +692,7 @@ end
 # the aggressive constprop pushes tA and tB into gemm_wrapper!, which is needed for wrap calls within it
 # to be concretely inferred
 Base.@constprop :aggressive function syrk_wrapper!(C::StridedMatrix{T}, tA::AbstractChar, A::StridedVecOrMat{T},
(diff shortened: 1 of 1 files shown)


jishnub (approved):

stevengj:
LGTM.  Do we have test coverage for this branch?

ViralBShah:
Why are all the tests failing? I couldn't tell from a quick glance.

KristofferC:
My guess is that the lazy jll stuff broke them.

araujoms:
I've found another typo, this time in `syrk_wrapper!`, I thought it would be better to add it here.

dkarrasch (approved):

ViralBShah:
The dot-jl job seems to fail with:

```
🚨 Error: mkdir C:\cache: Cannot create a file when that file already exists.
```

KristofferC:
This happens to many windows buildkite workers in the ecosystem (and for Julia itself).

dkarrasch:
I'll merge this and move forward with cleaning up the `herk_wrapper!`.

PR #1255, opened 2025-04-01
use smaller matrix size in `peakflops` on 32-bit

JeffBezanson:
This might help prevent failures like https://buildkite.com/julialang/julia-master/builds/46307#0195de97-a496-4daa-b71e-3124e3592505. Allocating several 128MB arrays on 32-bit certainly sounds like asking for an OOM.

### DIFF
--- a/src/LinearAlgebra.jl
+++ b/src/LinearAlgebra.jl
@@ -726,6 +726,8 @@ end
 (\)(F::TransposeFactorization{T,<:LU}, B::VecOrMat{Complex{T}}) where {T<:BlasReal} =
     ldiv(F, B)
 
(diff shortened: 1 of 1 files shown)


ViralBShah (approved):

ViralBShah:
The actual tests seem to use `1024` so this shouldn't be causing a CI OOM. In any case, this is a good default for 32-bit systems.

https://github.com/JuliaLang/LinearAlgebra.jl/blob/master/test/generic.jl#L642

ViralBShah:
I see the actual issue comes from `test/threads_exec.jl` in the Julia tests that calls peakflops with no arguments. That test should probably also use `1024`.

PR #1277, opened 2025-04-09
`AbstractQ` bugfix: define `Base.IteratorSize`

nsajko:
This used to incorrectly fall back to `HasLength`. The correct type is `HasShape{2}`, as `AbstractQ` is two-dimensional.

Makes the `collect` method unnecessary.


jishnub (approved):

dkarrasch:
Nice, thank you!

PR #1331, opened 2025-05-02
Document SingularException throw for inv(::AbstractMatrix)

Antoinemarteau:
I suggest documenting that `inv(::AbstractMatrix)` throws on (numerically) singular matrices, this is different from `inv(::Number)` that returns `Inf`.

```julia 
inv( 0. )    # Inf, no exception
inv( [0;;] ) # SingularException 
```

I guess documenting this would mean that `inv` guarantees to return a matrix with only finite (`isfinite`) entries if the argument has only finite entries, but I don't know if the implementation guarantees it / if this should be tested ?

### DIFF
--- a/src/generic.jl
+++ b/src/generic.jl
@@ -1152,6 +1152,8 @@ Matrix inverse. Computes matrix `N` such that
 Computed by solving the left-division
 `N = M \\ I`.
 
+A [`SingularException`](@ref) is thrown if `M` fails to be numerically inverted.
+
 # Examples
 ```jldoctest
 julia> M = [2 5; 1 3]
--- a/src/generic.jl
+++ b/src/generic.jl
@@ -1152,7 +1152,7 @@ Matrix inverse. Computes matrix `N` such that
 Computed by solving the left-division
 `N = M \\ I`.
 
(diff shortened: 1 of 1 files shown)


Antoinemarteau:
Also, `inv(M)` where `M = [Nan;;]` currently returns `[Nan;;]`, but then `M*inv(M) == Matrix(I,1,1) ` is `false`, so what is returned isn't an inverse and the documentation doesn't specify it.

Should this be also documented, or fixed to throw a SingularExcteption ? On one hand `[Nan;;]` do not have a zero eigenvalue like `SingularException` says, but it isn't invertible either.

Other inconsistencies with NaN:
```juliadoctest
julia> M = [NaN 1; 1 0;]

julia> inv(M) 
ERROR: ArgumentError: matrix contains Infs or NaNs
... 

julia> M = [NaN 0; 0 1;]

julia> inv(M)
2×2 Matrix{Float64}:
 NaN    NaN
   0.0    1.0

julia> Matrix(I,2,2) == inv(M)*M
false
```

ViralBShah:
Welcome @Antoinemarteau !

PR #1358, opened 2025-05-13
Cleanup and generalize functions of Hermitian matrices (#1340 encore)

araujoms:


araujoms:
The test failure on Windows is unrelated, the method ambiguities themselves are fixed, as the tests on Linux and Mac show.

jishnub (approved):

src/symmetric.jl
```diff
@@ -956,49 +912,28 @@ function sincos(A::RealHermSymSymTri)
...
-    T = float(eltype(F.values))
-    S, C = Diagonal(similar(A, T, (n,))), Diagonal(similar(A, T, (n,)))
-    for i in eachindex(S.diag, C.diag, F.values)
-        S.diag[i], C.diag[i] = sincos(F.values[i])
-    end
-    retmatS, retmatC = (F.vectors * S) * F.vectors', (F.vectors * C) * F.vectors'
-    for i in diagind(retmatS, IndexStyle(retmatS))
-        retmatS[i] = real(retmatS[i])
-        retmatC[i] = real(retmatC[i])
+    if all(λ -> λ ≥ 0, F.values)
```
dkarrasch, line 918: I think this should be `>0`, shouldn't it? https://en.wikipedia.org/wiki/Logarithm_of_a_matrix is talking about invertible matrices, and zero eigenvalues preclude invertibility. Even in the scalar case, ``\log(0)`` is not defined.

araujoms: I had actually done that, but reverted myself because it makes no difference. This branch is only switching between `log(x)` and `log(complex(x))`, and in both cases the answer is `-Inf`, and the resulting matrix is garbage. I decided to stay with `≥` because it's better not to change anything when there's no need. Also this makes it consistent with `log(::Diagonal)`.

araujoms:
I've been thinking about how to test matrix functions on `Hermitian{<:Quaternion}`, to make sure the capability doesn't get accidentally lost in future refactorings. It would quite overkill to include a spectral eigensolver for quaternions in the test files, so perhaps one could instead mock one with something like
```julia
function eigen(A::Hermitian{<:Quaternion})
    U = copy(Matrix(A))
    d = checksquare(A)
    λ = randn(real(eltype(A)), d)
    return Eigen(λ, U)
end
```
Then one could simply test whether the functions don't throw when called on `Hermitian{<:Quaternion}`, trusting that correctness would follow from the real and complex cases.

What do you think? Perhaps the entire thing is a waste of time and I should just leave it be.

dkarrasch:
Hm, perhaps you could design a spectral decomposition right away? Do we have some generic `qr`, that would give an orthogonal `U`, and then you specify the real eigenvalues as you like (positive, non-negative, any).

araujoms:
It's easy to produce an orthogonal `U`, but I don't see how would that help.

dkarrasch:
I thought you wanted to test for quaternion matrices, and for that purpose you needed a spectral decomposition. If you start with one, you could multiply out `A = U'D*U` and use that one to run the functions on? And even compare the results with a manual `U' * f.(D) *U`? Maybe I misunderstood.

araujoms:
What I want to test is e.g. that `exp(A)` works when `A` is a Hermitian quaternionic matrix. The current implementation of `exp` will work whenever `eigen` supports such `A` (this happens with `GenericLinearAlgebra`). So in order to test that I don't simply need a spectral decomposition of `A`, I need `eigen` to give it to me.

PR #1360, opened 2025-05-17
type stability of matrix functions

### LINKED ISSUE #460: `^(::Matrix{Float64},::Float64)` not type stable
`@code_warntype [1. 1.;1. 0.]^2.` shows return type `Any`, same holds for Complex types, i.e. the test

```julia
@testset "type stability of ^(::Matrix{Float64},::Float64)" for elty in (Float64, Float32)
    @test Base.return_types(^, (Matrix{elty}, elty)) == [Matrix{elty}]
    @test Base.return_types(^, (Matrix{Complex{elty}}, elty)) == [Matrix{elty}]
end
```
fails (both on Julia 0.6 and 0.7). This issue is related to but different from JuliaLang/julia#23366 (Float64 vs Int64).

araujoms:
I've fixed the type stability of matrix functions as far as possible without changing the API as discussed in #1130. Also added tests. Closes #460.

### DIFF
--- a/src/dense.jl
+++ b/src/dense.jl
@@ -628,7 +628,7 @@ function _safe_parent(fA)
     elseif isa(fA, Symmetric)
         return copytri_maybe_inplace(parentfA, 'U')
     else
-        return parentfA
+        return fA
     end
 end
 """
--- a/src/dense.jl
+++ b/src/dense.jl
@@ -616,7 +616,13 @@ function (^)(A::AbstractMatrix{T}, p::Real) where T
     if ishermitian(A)
         powerHermA = Hermitian(A)^p
         PP = parent(powerHermA)
-        return ishermitian(powerHermA) ? copytri_maybe_inplace(PP, 'U', true) : PP
+        if isa(powerHermA, Hermitian) || isa(powerHermA, Symmetric{<:Real})
+            return copytri_maybe_inplace(PP, 'U', true)
+        elseif isa(powerHermA, Symmetric)
+            return copytri_maybe_inplace(PP, 'U')
+        else
+            return PP
+        end
     end
 
     # Otherwise, use Schur decomposition
--- a/src/dense.jl
+++ b/src/dense.jl
@@ -614,7 +614,9 @@ function (^)(A::AbstractMatrix{T}, p::Real) where T
 
     # If possible, use diagonalization
     if ishermitian(A)
-        return parent(Hermitian(A)^p)
+        powerHermA = Hermitian(A)^p
+        PP = parent(powerHermA)
+        return ishermitian(powerHermA) ? copytri_maybe_inplace(PP, 'U', true) : PP
     end
 
     # Otherwise, use Schur decomposition
@@ -744,7 +746,7 @@ function exp!(A::StridedMatrix{T}) where T<:BlasFloat
         end
         return A
     elseif ishermitian(A)
-        return parent(exp(Hermitian(A)))
+        return copytri!(parent(exp(Hermitian(A))), 'U', true)
     end
     ilo, ihi, scale = LAPACK.gebal!('B', A)    # modifies A
     nA   = opnorm(A, 1)
@@ -1010,7 +1012,9 @@ function sqrt(A::AbstractMatrix{T}) where {T<:Union{Real,Complex}}
             return applydiagonal(sqrt, A)
         end
     elseif ishermitian(A)
-        return parent(sqrt(Hermitian(A)))
+        sqrtHermA = sqrt(Hermitian(A))
+        PS = parent(sqrtHermA)
+        return ishermitian(sqrtHermA) ? copytri_maybe_inplace(PS, 'U', true) : PS
     elseif istriu(A)
         return triu!(parent(sqrt(UpperTriangular(A))))
     elseif isreal(A)
@@ -1077,7 +1081,7 @@ function cbrt(A::AbstractMatrix{<:Real})
     elseif isdiag(A)
         return applydiagonal(cbrt, A)
     elseif issymmetric(A)
-        return copytri!(parent(cbrt(Symmetric(A))), 'U')
+        return copytri_maybe_inplace(parent(cbrt(Symmetric(A))), 'U')
     else
         S = schur(A)
         return S.Z * _cbrt_quasi_triu!(S.T) * S.Z'
--- a/src/dense.jl
+++ b/src/dense.jl
@@ -592,7 +592,7 @@ function schurpow(A::AbstractMatrix, p)
     end
 
     # if A has nonpositive real eigenvalues, retmat is a nonprincipal matrix power.
-    if isreal(retmat)
+    if eltype(A) <: Real && isreal(retmat)
         return real(retmat)
     else
         return retmat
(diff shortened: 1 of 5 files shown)


ViralBShah:
Would be good for @jishnub to take a look.

src/dense.jl
```diff
@@ -602,20 +602,19 @@ function (^)(A::AbstractMatrix{T}, p::Real) where T
...
-        return retmat
     end
 
     # For integer powers, use power_by_squaring
     isinteger(p) && return integerpow(A, p)
 
     # If possible, use diagonalization
     if ishermitian(A)
-        return (Hermitian(A)^p)
+        return parent(Hermitian(A)^p)
```
jishnub: Since this is a general method that accepts arbitrary `AbstractMatrix` types, we can't always be sure that the parent of `Hermitian(A)^p` contains meaningful values in the non-stored half (or is initialized at all). If we want to return the parent, we should call `copytri!` to ensure that the non-stored triangular half is populated as well. We don't have any tests for such cases at present, I guess.

Edit: I later saw that a similar change has been used in line 1080 (in `cbrt`).

A secondary, but possibly less serious, issue is that `Hermitian(A)^p` may not return a `Hermitian` at all, so we may not know what we are calling `parent` on. This isn't a big concern in practice because `parent(A)` returns `A` by default.

src/dense.jl
```diff
@@ -1004,13 +1003,14 @@ sqrt(::AbstractMatrix)
...
+        if T <: Real && any(<(0), diagview(A))
+            return applydiagonal(sqrt ∘ complex, A)
+        else
+            return applydiagonal(sqrt, A)
+        end
     elseif ishermitian(A)
-        sqrtHermA = sqrt(Hermitian(A))
-        PS = parent(sqrtHermA)
-        return ishermitian(sqrtHermA) ? copytri_maybe_inplace(PS, 'U', true) : PS
+        return parent(sqrt(Hermitian(A)))
```
jishnub: Similar concern about exposing the `parent` as above, as it may not be fully initialized or contain meaningful values.

test/symmetric.jl
```diff
@@ -1199,13 +1199,64 @@ end
...
+@testset "type inference of matrix functions" begin
+    for T ∈ (Float32, Float64, ComplexF32, ComplexF64)
+        a = randn(T, 2, 2)
+        syma = Symmetric(a)
+        symtria = SymTridiagonal(syma)
+        herma = Hermitian(a)
+        #nasty functions
+        for f in (x->x^real(T)(0.3), sqrt, log, asin, acos, acosh, atanh)
+            if T <: Real
+                @test @inferred Matrix{Complex{T}} f(a) isa Matrix
```
jishnub:
```suggestion
                @test @inferred(Matrix{Complex{T}}, f(a)) isa Matrix
```
Otherwise, the `@inferred` acts on the entire expression. Similar changes to the other tests below.

src/dense.jl
araujoms: This works because of #1359. It doesn't apply to `cbrt`, so I added a `copytri!` for it. `cbrt` does test for it already, I don't know about `^`, though.

jishnub: The symmetric methods are generic fallbacks, but packages may always define narrower methods. As a contrived example, we define a custom matrix type and specialize `^` for a `Hermitian` that wraps the type:
```julia
using LinearAlgebra, GenericLinearAlgebra

struct MyMatrix{T} <: AbstractMatrix{T}
    A::Matrix{T}
end
Base.size(A::MyMatrix) = size(A.A)
Base.getindex(A::MyMatrix, i::Int, j::Int) = A.A[i, j]

Base.parent(A::MyMatrix) = A.A

function Base.:(^)(A::Hermitian{T, MyMatrix{T}}, p::Real) where {T}
    copy(Hermitian(parent(parent(A))) ^ p)
end
```
There is nothing wrong with the definition as such, but with this, we obtain:
```julia
julia> M = MyMatrix(BigFloat[1 2; 2 4])
2×2 MyMatrix{BigFloat}:
 1.0  2.0
 2.0  4.0

julia> M ^ 2.5
2×2 Matrix{BigFloat}:
  11.1803  22.3607
 #undef    44.7214
```
This is because we are exposing the parent without ensuring that the parent is populated.

araujoms: Fair enough, I undid the corresponding changes. I do have the same complaint as before, though, it's hard to program for unknown features, that only get used by other packages.

src/dense.jl
```diff
@@ -602,20 +602,21 @@ function (^)(A::AbstractMatrix{T}, p::Real) where T
...
 
     # For integer powers, use power_by_squaring
     isinteger(p) && return integerpow(A, p)
 
     # If possible, use diagonalization
     if ishermitian(A)
-        return (Hermitian(A)^p)
+        powerHermA = Hermitian(A)^p
+        PP = parent(powerHermA)
+        return ishermitian(powerHermA) ? copytri_maybe_inplace(PP, 'U', true) : PP
```
jishnub: Is `ishermitian` necessary here, or can we do something similar for a `Symmetric` matrix as well? Also, perhaps we may check for types instead of values, as the check would be `O(N^2)` otherwise (and the `copytri` is also `O(N^2)` here).

araujoms: Checking only for types is a good idea. We can also do the analogous check and copy for `Symmetric{<:Complex}`. I'm similarly unenthusiastic about it. It causes a slowdown and is only needed by some unknown package.

Note that we do the analogous returns in `sqrt`, `log`, `asin`, `acos`, `acosh`, and `atanh`, so anything we change here should also be changed there. Perhaps it's better to write a helper function to avoid code duplication.

araujoms: Done.

src/dense.jl
```diff
@@ -602,26 +602,35 @@ function (^)(A::AbstractMatrix{T}, p::Real) where T
...
 end
 
+function _safe_parent(fA)
+    parentfA = parent(fA)
+    if isa(fA, Hermitian) || isa(fA, Symmetric{<:Real})
+        return copytri_maybe_inplace(parentfA, 'U', true)
+    elseif isa(fA, Symmetric)
+        return copytri_maybe_inplace(parentfA, 'U')
+    else
+        return parentfA
```
jishnub: Perhaps it might be better to return `fA` here, as we don't really have a notion of what the parent of an arbitrary matrix represents. In common cases, this branch will be hit when `Hermitian(A::Matrix)^p` returns a `Matrix`, so returning `fA` directly seems ok for type-stability.

araujoms: Good point. Here I only wanted to take care of the `Symmetric` and `Hermitian` cases, and that's done. If some evil package returns a transpose as a result of `f(Hermitian(A))` the result here would be wrong. Done.

src/dense.jl
```diff
@@ -602,26 +602,35 @@ function (^)(A::AbstractMatrix{T}, p::Real) where T
     checksquare(A)
     # Quicker return if A is diagonal
     if isdiag(A)
-        TT = promote_op(^, T, typeof(p))
-        retmat = copymutable_oftype(A, TT)
-        for i in diagind(retmat, IndexStyle(retmat))
-            retmat[i] = retmat[i] ^ p
+        if T <: Real && any(<(0), diagview(A))
```
jishnub, line 605: From what I understand, these changes aren't related to the type-stability issue that is addressed in this PR. Would you mind making a separate PR with such changes that should be easy to merge?

araujoms: I do mind, this is pointless bureaucracy.

ViralBShah: In the past, in Julia, we have been bitten when issues come up and diagnosis is difficult when PRs that do a lot of different things. I would generally agree to keep PRs clean and to the point, which makes future diagnosis easier.

Maybe we can merge this one for now.

jishnub: We need to check the performance impact of this. I haven't had a lot of time to look into this yet.

jishnub: The impact seems minimal in the Hermitian case that I tried, which is probably expected, as the `eigen`, being O(n^3) would dominate.

PR #1386, opened 2025-06-19
Add low-level wrappers for BLAS and LAPACK

amontoison:
@ViralBShah I was at the Householder Symposium last week and had a discussion with Julien Langou (@langou), the main maintainer of reference-LAPACK, about what we’re doing in Julia for BLAS / LAPACK and our nice LBT feature.

Some feedback I got is that Julien doesn’t understand why we don’t expose all the arguments of the BLAS / LAPACK routines, and why we took the liberty of fusing `om*` / `un*` routines but not `sy*` / `he*`.

What I concluded with Julien is that we should maybe expose the low-level wrappers directly and call them from our higher-level interfaces.
This would make it easier to update with new releases and would also give users the ability to reuse buffers, which is currently a significant limitation of our API.
I also think it would help Julien interact more easily with us.
He’s starting to get interested in using Julia for unit testing reference-LAPACK and teaching!

A few months ago, I wrote a script to generate wrappers for Fortran routines implemented in F77 (for HSL), and I was able to adapt it for BLAS / LAPACK.
It is similar to what we obtain with `Clang.jl` from the C header files.
I open the PR as a draft for now to get feedback.


ViralBShah:
This is something I have always wanted to do myself as well. 

Is there a noticeable impact on system image size?

ViralBShah:
It would be great if this PR could be a branch on this repo (easier to work on than a fork). Possibly needs these fixes:

```
diff --git a/src/blas.jl b/src/blas.jl
index 15c6467..3ebe212 100644
--- a/src/blas.jl
+++ b/src/blas.jl
@@ -1417,7 +1417,7 @@ Return the updated `C`.
 """
 function gemmt! end
 
-for (gemmt, elty) in
+for (fname, elty) in
         ((:dgemmt,:Float64),
          (:sgemmt,:Float32),
          (:zgemmt,:ComplexF64),
@@ -1488,7 +1488,7 @@ Update `C` as `alpha*A*B + beta*C` or the other three variants according to
 """
 function gemm! end
 
-for (gemm, elty) in
+for (fname, elty) in
         ((:dgemm,:Float64),
          (:sgemm,:Float32),
          (:zgemm,:ComplexF64),
@@ -1595,9 +1595,9 @@ for (mfname, elty) in ((:dsymm,:Float64),
             chkstride1(A)
             chkstride1(B)
             chkstride1(C)
-            $fname(side, uplo, m, n,
-                   alpha, A, max(1,stride(A,2)), B,
-                   max(1,stride(B,2)), beta, C, max(1,stride(C,2)))
+            $mfname(side, uplo, m, n,
+                    alpha, A, max(1,stride(A,2)), B,
+                    max(1,stride(B,2)), beta, C, max(1,stride(C,2)))
             C
         end
         function symm(side::AbstractChar, uplo::AbstractChar, alpha::($elty), A::AbstractMatrix{$elty}, B::AbstractMatrix{$elty})
```

amontoison:
I wanted to create a branch but it seems that I don't have the rights on this repo.
It is why I did a fork.

Thanks for the patch, I also need to find how to use `@blasfunc` with a `@ccall`.

ViralBShah:
I didn't realize you don't have access. Just invited.

ViralBShah:
Should we put energy into this and see if we can get it done? Just asking mainly because things will start diverging at some point and it will be more work to bring it in.

amontoison:
@ViralBShah I am still traveling this week for a conference but I can work on it next week and do a meeting if needed.

amontoison:
Viral, I opened #1403, we need a smart way to add the suffix `_64` within a `@ccall`.

PR #1392, opened 2025-06-27
specialize kron!(::Diagonal, ::Diagonal, ::Diagonal)

tam724:
see https://github.com/JuliaLang/LinearAlgebra.jl/issues/1391.
Implements a specialized version of `kron!` that dispatches on ::Diagonal as the first argument.

### DIFF
--- a/test/diagonal.jl
+++ b/test/diagonal.jl
@@ -1396,15 +1396,15 @@ end
 end
 
 @testset "kron! for Diagonal" begin
-    a = Diagonal([1,2])
-    b = Diagonal([3,4])
+    a = Diagonal([1, 2])
+    b = Diagonal([3, 4])
     # Diagonal out
-    c = Diagonal([0,0,0,0])
-    kron!(c,b,a)
+    c = Diagonal([0, 0, 0, 0])
+    kron!(c, b, a)
     @test c == Diagonal([3, 6, 4, 8])
     @test c == kron!(fill(0, 4, 4), Matrix(b), Matrix(a)) # against dense kron!
-    c=Diagonal(Vector{Float64}(undef, 4))
-    kron!(c,a,b)
+    c = Diagonal(Vector{Float64}(undef, 4))
+    kron!(c, a, b)
     @test c == Diagonal([3.0, 4.0, 6.0, 8.0])
 
     # AbstractArray out
--- a/test/diagonal.jl
+++ b/test/diagonal.jl
@@ -1415,6 +1415,7 @@ end
     c = Matrix{Float64}(undef, 4, 4)
     kron!(c, a, b)
     @test c == diagm([3.0, 4.0, 6.0, 8.0])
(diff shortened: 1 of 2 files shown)


dkarrasch:
Since the 2-arg method doesn't redirect to the in-place method, we would need a test for this.

test/diagonal.jl
```diff
@@ -1396,14 +1396,25 @@ end
...
-    b = Diagonal([1,1])
+    a = Diagonal([1,2])
+    b = Diagonal([3,4])
+    # Diagonal out
     c = Diagonal([0,0,0,0])
     kron!(c,b,a)
-    @test c == Diagonal([2,2,2,2])
+    @test c == Diagonal([3, 6, 4, 8])
+    @test c == kron!(fill(0, 4, 4), Matrix(b), Matrix(a)) # against dense kron!
     c=Diagonal(Vector{Float64}(undef, 4))
```
dkarrasch:
```suggestion
    c = Diagonal(Vector{Float64}(undef, 4))
```

test/diagonal.jl
```diff
@@ -1396,14 +1396,25 @@ end
 end
 
 @testset "kron! for Diagonal" begin
-    a = Diagonal([2,2])
-    b = Diagonal([1,1])
+    a = Diagonal([1,2])
+    b = Diagonal([3,4])
+    # Diagonal out
     c = Diagonal([0,0,0,0])
     kron!(c,b,a)
```
dkarrasch:
```suggestion
    kron!(c, b, a)
```

dkarrasch (approved):
Since we're at it, we might as well "fix" the style bits. Otherwise LGTM.

tam724:
Thanks for the review. I added the whitespaces.

dkarrasch:
Not sure why it says those lines are not covered... This LGTM, so I'll merge.

test/diagonal.jl
```diff
@@ -1396,14 +1396,25 @@ end
...
+    @test c == Diagonal([3.0, 4.0, 6.0, 8.0])
+
+    # AbstractArray out
+    c = fill(0, 4, 4)
+    kron!(c, b, a) 
+    @test c == diagm([3, 6, 4, 8])
+    @test c == kron!(fill(0, 4, 4), Matrix(b), Matrix(a)) # against dense kron!
+    c = Matrix{Float64}(undef, 4, 4)
+    kron!(c, a, b)
+    @test c == diagm([3.0, 4.0, 6.0, 8.0])
```
dkarrasch, line 1417: I just hesitated, but realized that size-checking is contained in the `kron!` for vectors, so we don't need to add it in the method. We should, however, safeguard against regression if someone decides to not forward the call.

```suggestion
    @test c == diagm([3.0, 4.0, 6.0, 8.0])
    @test_throws DimensionMismatch kron!(Diagonal(zeros(5)), Diagonal(zeros(2)), Diagonal(zeros(2)))
```

dkarrasch:
Thank you very much, @tam724!

PR #1398, opened 2025-07-02
return SVD from pinv(::SVD)

stevengj:
As suggested by @andreasnoack in https://github.com/JuliaLang/LinearAlgebra.jl/pull/1387#issuecomment-3024355447, this changes the `pinv(::SVD)` function to instead return the SVD of the pseudo-inverse, rather than an explicit matrix, so that it can be applied stably.

(No backwards-compatibility issue since the `pinv(::SVD)` method was introduced in #1387.)

### DIFF
--- a/src/svd.jl
+++ b/src/svd.jl
@@ -306,7 +306,7 @@ function ldiv!(F::SVD{T}, B::AbstractVecOrMat; atol::Real=0, rtol::Real = (eps(r
     return B
 end
 
-function pinv(F::SVD{T,<:Any,M}; atol::Real=0, rtol::Real = (eps(real(float(oneunit(T))))*min(size(F)...))*iszero(atol)) where {T,M}
+function pinv(F::SVD{T}; atol::Real=0, rtol::Real = (eps(real(float(oneunit(T))))*min(size(F)...))*iszero(atol)) where {T}
     k = _count_svdvals(F.S, atol, rtol)
     @views SVD(copy(F.Vt[1:k, :]'), inv.(F.S[1:k]), copy(F.U[:,1:k]'))
 end
--- a/src/svd.jl
+++ b/src/svd.jl
@@ -306,9 +306,9 @@ function ldiv!(F::SVD{T}, B::AbstractVecOrMat; atol::Real=0, rtol::Real = (eps(r
     return B
 end
 
-function pinv(F::SVD{T}; atol::Real=0, rtol::Real = (eps(real(float(oneunit(T))))*min(size(F)...))*iszero(atol)) where T
+function pinv(F::SVD{T,<:Any,M}; atol::Real=0, rtol::Real = (eps(real(float(oneunit(T))))*min(size(F)...))*iszero(atol)) where {T,M}
     k = _count_svdvals(F.S, atol, rtol)
-    @views (F.S[1:k] .\ F.Vt[1:k, :])' * F.U[:,1:k]'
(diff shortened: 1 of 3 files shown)


src/svd.jl
```diff
@@ -306,17 +306,18 @@ function ldiv!(F::SVD{T}, B::AbstractVecOrMat; atol::Real=0, rtol::Real = (eps(r
...
 end
 
 function inv(F::SVD)
     # TODO: checksquare(F)
     @inbounds for i in eachindex(F.S)
         iszero(F.S[i]) && throw(SingularException(i))
     end
-    pinv(F; rtol=eps(real(eltype(F))))
+    k = _count_svdvals(F.S, 0, eps(real(eltype(F))))
+    return @views (F.S[1:k] .\ F.Vt[1:k, :])' * F.U[:,1:k]'
```
stevengj, line 320: Note that I could have also just done `Matrix(pinv(F; rtol=eps(real(eltype(F)))))` here, at the cost of a couple more allocations.  Not sure it is worth it to save the allocations here.

src/svd.jl
```diff
@@ -306,17 +306,18 @@ function ldiv!(F::SVD{T}, B::AbstractVecOrMat; atol::Real=0, rtol::Real = (eps(r
     return B
 end
 
-function pinv(F::SVD{T}; atol::Real=0, rtol::Real = (eps(real(float(oneunit(T))))*min(size(F)...))*iszero(atol)) where T
+function pinv(F::SVD{T,<:Any,M}; atol::Real=0, rtol::Real = (eps(real(float(oneunit(T))))*min(size(F)...))*iszero(atol)) where {T,M}
     k = _count_svdvals(F.S, atol, rtol)
-    @views (F.S[1:k] .\ F.Vt[1:k, :])' * F.U[:,1:k]'
+    @views SVD(M(F.Vt[1:k, :]'), inv.(F.S[1:k]), M(F.U[:,1:k]'))
```
stevengj: Not sure `M` is the right type here — what if it is a `SubArray`?   

Is there a way to get the "unwrapped" dense matrix type?  `typeof(parent(F.Vt))` is not recursive.

stevengj:
Whoops, failing because `*(::SVD, ::AbstractVector)` is not defined.  This should be fixed — otherwise returning an `SVD` object from `pinv` is useless.

*Update:* fixed.

src/svd.jl
dkarrasch: What if you left this to the promotion mechanism inside the `SVD` constructor?

stevengj: It doesn't work IIRC; it tries to promote the views to a common subarray type and fails.

stevengj: I just realized that simply calling `copy` does the right thing here.

dkarrasch (approved):

stevengj:
CI failure on Windows is an unrelated server glitch.

stevengj:
Yay, green again.

MasonProtter:
Maybe a dumb question, but why not also have `inv(::SVD)` return an `SVD`?

dkarrasch:
We have an open issue (#635) regarding whether `inv` of factorizations should return a `Factorization`. The issue is that not every factorization has an inverse that can be represented by a type that we already have. `SVD` is tempting, of course, but I guess consistency is a potential issue here.

stevengj:
@MasonProtter, the other issue is that changing `inv(::SVD)` would be a breaking change, whereas `pinv(::SVD)` was a new method (added a few days ago).

PR #1400, opened 2025-07-03
make `dot` docstring match the implementation

### LINKED ISSUE #1453: `dot` overflows
Filing for future reference (would be too much of an interruption to fix now):

``` jl
julia> a = rand(UInt8, 5)
5-element Array{UInt8,1}:
 0x08
 0xcd
 0x4a
 0x5d
 0x1f

julia> b = rand(UInt8, 5)
5-element Array{UInt8,1}:
 0x7c
 0xf4
 0x2c
 0x86
 0xef

julia> dot(a, b)
0x9b
```

The fix is less trivial than it seems: the `r_promote` infrastructure in `reduce.jl` seems promising, but it doesn't handle the product of elements, only the sum of the accumulator. We might need a `widen_op` analogous to `promote_op` to prevent these kinds of overflows.

adienes:
as written it's incorrect since `sum` will promote the types but `+` (and `dot`) overflow

```
julia> x = collect(0xf0:0xff); y = copy(x);

julia> sum(dot(i, j) for (i, j) in zip(x, y))
0x00000000000004d8

julia> dot(x, y)
0xd8
```

closes https://github.com/JuliaLang/LinearAlgebra.jl/issues/1453

### DIFF
--- a/src/generic.jl
+++ b/src/generic.jl
@@ -938,7 +938,7 @@ vector is conjugated.
 `dot` also works on arbitrary iterable objects, including arrays of any dimension,
 as long as `dot` is defined on the elements.
 
-`dot` is semantically equivalent to `sum(dot(vx,vy) for (vx,vy) in zip(x, y))`,
+`dot` is semantically equivalent to `reduce(+, dot(vx,vy) for (vx,vy) in zip(x, y))`,
 with the added restriction that the arguments must have equal lengths.
 
 `x ⋅ y` (where `⋅` can be typed by tab-completing `\\cdot` in the REPL) is a synonym for


stevengj:
Should we instead fix the code to match the docstring?

adienes:
I suppose that depends on whether you think of dot more as a reducing operation or a "scalar" operation on scalars-that-are-actually-vectors, in that reducing operations typically widen types and arithmetic does not.

note that numpy also chooses to overflow on `np.dot` (so Julia and numpy are at the moment consistent)

araujoms (approved):
The docstring has been wrong for a decade now. One can fix it immediately. If it is ever decided to make `dot` widen types one changing it back is not an issue.

And I don't think `dot` will ever be changed.

adienes:
I'd also note that the people doing linear algebra on fixed-width unsigned ints (I suppose cryptographers) likely intentionally want modular arithmetic anyway. so I'm not sure that "fixing" the behavior would be any improvement. it also currently matches a naive dot with pointwise wise multiplication & `+` summation.

PR #1402, opened 2025-07-05
fix: add BLAS.geru! manual doc

### LINKED ISSUE #1401: `BLAS.geru!` doc is missing
https://github.com/JuliaLang/LinearAlgebra.jl/blob/67c1103e350bcce0ecc0c92a25ceac148e14b399/src/blas.jl#L54
The `geru!` function is exported but not documented. Since it's different from `ger!` for not conjugating complex numbers, I would wish it to be documented for advanced use? 
Not sure what's the best way to document it. If there's any way I could help I'd love to try out!

hz-xiaxz:
closes #1401

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -745,6 +745,7 @@ LinearAlgebra.BLAS.trsv
 
 ```@docs
 LinearAlgebra.BLAS.ger!
+LinearAlgebra.BLAS.geru!
 # xGERU
 # xGERC
 LinearAlgebra.BLAS.her!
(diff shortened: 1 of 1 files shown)


dkarrasch:
I think the comment is like a placeholder, like a to-do, that you have just done. ;-) You could remove that line.

hz-xiaxz:
Yes but I found similar patterns here, https://github.com/JuliaLang/LinearAlgebra.jl/blob/67c1103e350bcce0ecc0c92a25ceac148e14b399/src/blas.jl#L53C4-L57C10, not sure why geru! is missing. I would fix it.

stevengj (approved):

PR #1403, opened 2025-07-07
Add low-level wrappers for BLAS and LAPACK

amontoison:
Superseed #1386 
cc @ViralBShah @langou

### DIFF
--- a/src/blas.jl
+++ b/src/blas.jl
@@ -86,6 +86,17 @@ export
 
 using ..LinearAlgebra: libblastrampoline, BlasReal, BlasComplex, BlasFloat, BlasInt, DimensionMismatch, checksquare, chkstride1
 
+if USE_BLAS64
+    macro blasfunc(x)
+        return Expr(:quote, Symbol(x, "64_"))
+    end
+else
+    macro blasfunc(x)
+        return Expr(:quote, x)
+    end
+end
+
+include("libblas.jl")
 include("lbt.jl")
 
 # Legacy bindings that some packages (such as NNlib.jl) use.
@@ -106,16 +117,6 @@ Return an object representing the current `libblastrampoline` configuration.
     `get_config()` requires at least Julia 1.7.
 """
 get_config() = lbt_get_config()
-
-if USE_BLAS64
-    macro blasfunc(x)
-        return Expr(:quote, Symbol(x, "64_"))
-    end
-else
-    macro blasfunc(x)
-        return Expr(:quote, x)
-    end
-end
 
 _tryparse_env_int(key) = tryparse(Int, get(ENV, key, ""))
 
@@ -207,16 +208,14 @@ Copy `n` elements of array `X` with stride `incx` to array `Y` with stride `incy
 """
 function blascopy! end
 
-for (fname, elty) in ((:dcopy_,:Float64),
-                      (:scopy_,:Float32),
-                      (:zcopy_,:ComplexF64),
-                      (:ccopy_,:ComplexF32))
+for (fname, elty) in ((:dcopy,:Float64),
+                      (:scopy,:Float32),
+                      (:zcopy,:ComplexF64),
+                      (:ccopy,:ComplexF32))
     @eval begin
         # SUBROUTINE DCOPY(N,DX,INCX,DY,INCY)
         function blascopy!(n::Integer, DX::Union{Ptr{$elty},AbstractArray{$elty}}, incx::Integer, DY::Union{Ptr{$elty},AbstractArray{$elty}}, incy::Integer)
-            ccall((@blasfunc($fname), libblastrampoline), Cvoid,
-                (Ref{BlasInt}, Ptr{$elty}, Ref{BlasInt}, Ptr{$elty}, Ref{BlasInt}),
-                 n, DX, incx, DY, incy)
+            $fname(n, DX, incx, DY, incy)
             DY
         end
     end
@@ -236,18 +235,16 @@ first `n` elements of array `Y` with stride `incy`. Returns `X` and `Y`.
 """
 function rot! end
 
-for (fname, elty, cty, sty, lib) in ((:drot_, :Float64, :Float64, :Float64, libblastrampoline),
-                                     (:srot_, :Float32, :Float32, :Float32, libblastrampoline),
-                                     (:zdrot_, :ComplexF64, :Float64, :Float64, libblastrampoline),
-                                     (:csrot_, :ComplexF32, :Float32, :Float32, libblastrampoline),
-                                     (:zrot_, :ComplexF64, :Float64, :ComplexF64, libblastrampoline),
-                                     (:crot_, :ComplexF32, :Float32, :ComplexF32, libblastrampoline))
+for (fname, elty, cty, sty) in ((:drot, :Float64, :Float64, :Float64),
+                                (:srot, :Float32, :Float32, :Float32),
+                                (:zdrot, :ComplexF64, :Float64, :Float64),
+                                (:csrot, :ComplexF32, :Float32, :Float32),
+                                (:zrot, :ComplexF64, :Float64, :ComplexF64),
+                                (:crot, :ComplexF32, :Float32, :ComplexF32))
     @eval begin
         # SUBROUTINE DROT(N,DX,INCX,DY,INCY,C,S)
         function rot!(n::Integer, DX::Union{Ptr{$elty},AbstractArray{$elty}}, incx::Integer, DY::Union{Ptr{$elty},AbstractArray{$elty}}, incy::Integer, C::$cty, S::$sty)
-            ccall((@blasfunc($fname), $lib), Cvoid,
-                (Ref{BlasInt}, Ptr{$elty}, Ref{BlasInt}, Ptr{$elty}, Ref{BlasInt}, Ref{$cty}, Ref{$sty}),
-                 n, DX, incx, DY, incy, C, S)
+            $fname(n, DX, incx, DY, incy, C, S)
             DX, DY
         end
     end
@@ -275,16 +272,14 @@ If `n` and `incx` are not provided, `length(X)` and `stride(X,1)` are used.
 """
 function scal end
 
-for (fname, elty) in ((:dscal_,:Float64),
-                      (:sscal_,:Float32),
-                      (:zscal_,:ComplexF64),
-                      (:cscal_,:ComplexF32))
+for (fname, elty) in ((:dscal,:Float64),
+                      (:sscal,:Float32),
+                      (:zscal,:ComplexF64),
+                      (:cscal,:ComplexF32))
     @eval begin
         # SUBROUTINE DSCAL(N,DA,DX,INCX)
         function scal!(n::Integer, DA::$elty, DX::Union{Ptr{$elty},AbstractArray{$elty}}, incx::Integer)
-            ccall((@blasfunc($fname), libblastrampoline), Cvoid,
-                  (Ref{BlasInt}, Ref{$elty}, Ptr{$elty}, Ref{BlasInt}),
-                  n, DA, DX, incx)
+            $fname(n, DA, DX, incx)
             DX
         end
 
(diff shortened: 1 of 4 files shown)


ViralBShah:
The `@ccall` macro is bit annoying. I suppose this would have been simpler to do with just `ccall`. @amontoison Did you auto-generate these from some spec file?

amontoison:
Hello @ViralBShah, I autogenerated them from a custom Julia script that directly parse the Fortran files in Reference-LAPACK.
I will regenerate them with only `ccall`directly.

Note that that I will visit @langou next month during ten days and we plan to work on that.

ViralBShah:
There are a few lapack calls with one argument that need fixing too. Perhaps the generator script can be updated? It would be nice to have a `gen` directory where the generator scripts could be checked in.

In theory, we can use Clang.jl to automatically generate the entire API for CBLAS.

amontoison:
@ViralBShah I can upload my Fortran parser of F77 but I need to clean it a little bit.
We could the wrappers for CBLAS but we don't use any routine of it in Julia so it will just add more functions in `LinearAlgebra`.

What would be great to have the "buffer" version of the LAPACK routines with LAPACKE.
But it means a quite big revamp of `LinearAlgebra` and I am scared that a few people complain.

ViralBShah:
The reason to do cblas would be then to wrap that into the low level Julia interface and have Linear Algebra use that and delete the fortran bindings altogether.

The LAPACKE style overhaul can easily be done in a new external package.

ViralBShah:
We'll also need to see the impact this has on system image size.

amontoison:
All wrappers are removed in `blas.jl` and `liblapack.jl` so it is more modular now if we want to rely on CBLAS / LAPACKE.
Yes, the system image size could be an issue.

ViralBShah:
The tests are all passing now. I will check the system size impact. Hopefully it isn’t much.

Given that it is all working now, perhaps not much incentive to change to cblas now, but this puts us in a really good position to do it whenever we need to.

ViralBShah:
These methods are being generated twice. Should we just edit these by hand or would you like to regenerate the files?

```
WARNING: Method definition csrot(Any, Any, Any, Any, Any, Any, Any) in module BLAS at /home/viralbshah/julia/usr/share/julia/stdlib/v1.13/LinearAlgebra/src/libblas.jl:82 overwritten at /home/viralbshah/julia/usr/share/julia/stdlib/v1.13/LinearAlgebra/src/libblas.jl:343.
WARNING: Method definition drot(Any, Any, Any, Any, Any, Any, Any) in module BLAS at /home/viralbshah/julia/usr/share/julia/stdlib/v1.13/LinearAlgebra/src/libblas.jl:61 overwritten at /home/viralbshah/julia/usr/share/julia/stdlib/v1.13/LinearAlgebra/src/libblas.jl:506.
WARNING: Method definition dsymv(Any, Any, Any, Any, Any, Any, Any, Any, Any, Any) in module BLAS at /home/viralbshah/julia/usr/share/julia/stdlib/v1.13/LinearAlgebra/src/libblas.jl:9 overwritten at /home/viralbshah/julia/usr/share/julia/stdlib/v1.13/LinearAlgebra/src/libblas.jl:576.
WARNING: Method definition dsyr(Any, Any, Any, Any, Any, Any, Any) in module BLAS at /home/viralbshah/julia/usr/share/julia/stdlib/v1.13/LinearAlgebra/src/libblas.jl:40 overwritten at /home/viralbshah/julia/usr/share/julia/stdlib/v1.13/LinearAlgebra/src/libblas.jl:583.
WARNING: Method definition srot(Any, Any, Any, Any, Any, Any, Any) in module BLAS at /home/viralbshah/julia/usr/share/julia/stdlib/v1.13/LinearAlgebra/src/libblas.jl:68 overwritten at /home/viralbshah/julia/usr/share/julia/stdlib/v1.13/LinearAlgebra/src/libblas.jl:770.
WARNING: Method definition ssymv(Any, Any, Any, Any, Any, Any, Any, Any, Any, Any) in module BLAS at /home/viralbshah/julia/usr/share/julia/stdlib/v1.13/LinearAlgebra/src/libblas.jl:1 overwritten at /home/viralbshah/julia/usr/share/julia/stdlib/v1.13/LinearAlgebra/src/libblas.jl:834.
WARNING: Method definition ssyr(Any, Any, Any, Any, Any, Any, Any) in module BLAS at /home/viralbshah/julia/usr/share/julia/stdlib/v1.13/LinearAlgebra/src/libblas.jl:33 overwritten at /home/viralbshah/julia/usr/share/julia/stdlib/v1.13/LinearAlgebra/src/libblas.jl:841.
WARNING: Method definition zdrot(Any, Any, Any, Any, Any, Any, Any) in module BLAS at /home/viralbshah/julia/usr/share/julia/stdlib/v1.13/LinearAlgebra/src/libblas.jl:75 overwritten at /home/viralbshah/julia/usr/share/julia/stdlib/v1.13/LinearAlgebra/src/libblas.jl:951.
```

ViralBShah:
This PR adds 5MB to the system image on Linux. This seems acceptable to me.

My hope is that we can separate out the bits for `*`, `\` and `/` and relevant BLAS/LAPACK calls to be included in the system image in LinearAlgebraBase, and then not build LinearAlgebra itself into the system image.

ViralBShah:
Is it possible for the wrapper generator script to annotate the types for all the function calls? Currently these are all `Any`.

amontoison:
Hello @ViralBShah, in general we don't type the arguments of the wrappers because it is low-level routines for advance users. I can add types but I am scared to be too much restrictive.

With @langou, we discussed about the Julia wrappers and he thinks that we should interface CBLAS  / LAPACKE instead in LinearAlgebra.
It will be also easier to maintain because we can regenerate the wrappers with Clang.jl directly.

Viral, if you want to meet, we are available this week with Julien. We are working together this week on reference-LAPACK.

ViralBShah:
I am happy to meet, althoughI do not have much more to contribute than what I already have. I was wondering myself if we should use the C interfaces and auto generate.

I think it should be ok to add types to the methods and widen them for things like Integer. The rest have very little room and will be enforced by ccall anyways. Perhaps we get this for free in Clang.jl.

ViralBShah:
It seems like Apple Accelerate does not use LAPACKE.

ViralBShah:
The other question is LBT support. However it should be easy to add if C API support is missing.

Perhaps worth considering doing this in an external package?

stevengj:
If we have these, shouldn't they be called `unsafe_foo`, since they could crash if passed inconsistent arguments?

Who is this PR for, exactly?

ViralBShah:
People generally keep asking for a way to call the entire LAPACK and BLAS from Julia - not just our prepackaged factorizations. This adds the full API surface. While it is definitely useful, it is worthwhile to discuss where it should live.

stevengj:
In my mind, exposing a bunch of unsafe functions like this is not much better than telling them to use ccall. In both cases you need to look up a bunch of low-level arguments and understand low-level types etc. or it will crash. 

One alternative would just be to add a `@blas_ccall` macro that automates calling the correct trampoline, along with documenting a couple of blas types like the correct integer type.

amontoison:
The complaint of @langou when he tried Julia is that he doesn’t understand why some arguments are exposed and others are not, and I agree with him. The design of the routines / high-level interfaces in `LinearAlgebra.LAPACK` routines also doesn’t allow reusing buffers and, in the end, we need to do our own `ccall` in our Julia packages.

Example: https://github.com/JuliaSmoothOptimizers/Krylov.jl/blob/main/src/block_krylov_utils.jl#L210-L292 

What would be nice is to expose documented high-level wrappers without the prefixes `s`, `d`, `c`, `z`, with type arguments (like we have right now) but with all arguments. That way, it would be consistent with all calls to reference BLAS/LAPACK.

If we want something with very limited arguments, then we interface them in high-level Julia functions like `mul!`, `qr`, `lu`, etc. But currently we have a hybrid with `LinearAlgebra.BLAS` and `LinearAlgebra.LAPACK`, and the solution proposed is adding the low-level wrappers now so that we don’t break Julia users.

Another possible approach could be to keep the old methods (in `LinearAlgebra.LAPACK`), potentially with a limited number of arguments, to avoid breaking existing code, while providing a clean API with all arguments fully documented and typed.  

This would "just" involve duplicating some methods like `potrf` or `geqrf` that are already interfaced.

ViralBShah:
@amontoison How do you feel about a separate package? That package could even use Preferences to pick a BLAS and LAPACK and do other nice things.

stevengj:
> The complaint of @langou when he tried Julia is that he doesn’t understand why some arguments are exposed and others are not, and I agree with him. The design of the routines / high-level interfaces in LinearAlgebra.LAPACK routines also doesn’t allow reusing buffers and, in the end, we need to do our own ccall in our Julia packages.

That's a good argument for adding more high-level methods of these functions that expose more arguments, I agree.  I'm not sure what it has to do with the present PR, which is not much better than doing your own `ccall`?

imciner2:
> > All wrappers are removed in blas.jl and liblapack.jl so it is more modular now if we want to rely on CBLAS / LAPACKE.
> 
> > I was wondering myself if we should use the C interfaces and auto generate.

Picking up on this from the LBT perspective, LBT will forward the CBLAS and LAPACKE functions to the library only if it provides them. The exception to that is a library that contains both LP64 & ILP64 symbols in Fortran but only LP64 symbols in CBLAS - then LBT will use internal forwards for the ILP64 CBLAS functions to the library's ILP64 Fortran interface. We don't in general have a bridge layer for the CBLAS symbols, or for any LAPACKE symbols, in LBT. We could extend the CBLAS bridge creation to be a more general system that would use LBT bridge functions to go C->Fortran if CBLAS symbols are not detected, and a similar thing for LAPACKE as well.

One question is, how complete is the interface layer for CBLAS and LAPACKE in the "reference" version? I do see a few issues open where people say a LAPACK function is missing in LAPACKE (https://github.com/Reference-LAPACK/lapack/issues/962, https://github.com/OpenMathLib/OpenBLAS/discussions/5523)

Edit: Although looking at the LAPACKE functions some more, they aren't just simple wrappers around the LAPACK functions I guess. I don't know if we really would want to bring an entire fallback system into LBT that isn't just a simple wrapper.

ViralBShah:
We really should create an external package for this capability which will be way quicker to iterate on and improve.

langou:
> One question is, how complete is the interface layer for CBLAS and LAPACKE in the "reference" version? I do see a few issues open where people say a LAPACK function is missing in LAPACKE (Reference-LAPACK/lapack#962, OpenMathLib/OpenBLAS#5523)

CBLAS should be fully covering BLAS. I do not know how much LAPACKE is covering LAPACK. We have added wrappers to LAPACKE organically based on users' demands. The rationale has been that internal auxiliary LAPACK routines do not need LAPACKE wrappers. So we made an initial guess when we released LAPACKE. We only LAPACKE-ed the LAPACK routines labelled as "drivers". And then as people needed deeper routines we added these in LAPACKE based on request. Some takeaways: • LAPACKE does not cover the whole LAPACK, • But it covers what matters most, • If something is missing that seems relevant, we would consider adding it to LAPACKE.

> Although looking at the LAPACKE functions some more, they aren't just simple wrappers around the LAPACK functions I guess. I don't know if we really would want to bring an entire fallback system into LBT that isn't just a simple wrapper.

This is absolutely correct. In addition of being a C wrapper to LAPACK, LAPACKE has features such as (*) column major / row major, (*) automatic memory allocation, (*) NaN checking.

imciner2:
Thinking about the CBLAS and LAPACKE bits a bit more, I think what would be the best way of doing it would be building CBLAS-only and LAPACKE-only libraries in Yggdrasil (based on the Reference-LAPACK repo) that link against LBT, and then at the Julia level we can load those libraries if needed (e.g., we have a BLAS/LAPACK library that doesn't provide them). That can keep LBT light-weight still, and allow us to not need to duplicate the LAPACKE/CBLAS wrappers.

I think that would still need a small change in LBT, where we add a CBLAS and a LAPACKE symbol to the automatic detection logic to identify suffixes, although I don't think we can reliably do integer autodetect for things like CBLAS because it would call BLAS, so we would need the BLAS library loaded already.

ViralBShah:
It will be nice to have these be in a standalone package.

langou:
"It will be nice to have these be in a standalone package." => +1, I agree.

PR #1404, opened 2025-07-09
Fix incorrect dispatch for Cholesky equality

penelopeysm:
`==` on Cholesky is not correctly being dispatched when the two operands have exactly the same type:

```julia
julia> x = LowerTriangular([0.0 1.0; 0.0 0.0])
2×2 LowerTriangular{Float64, Matrix{Float64}}:
 0.0   ⋅
 0.0  0.0

julia> y = LowerTriangular([0.0 0.0; 0.0 0.0])
2×2 LowerTriangular{Float64, Matrix{Float64}}:
 0.0   ⋅
 0.0  0.0

julia> x == y
true

julia> Cholesky(x) == Cholesky(y)
false

julia> @which Cholesky(x) == Cholesky(y)
==(F::T, G::T) where T<:Factorization
     @ LinearAlgebra ~/.julia/juliaup/julia-1.11.5+0.aarch64.apple.darwin14/share/julia/stdlib/v1.11/LinearAlgebra/src/factorization.jl:116
```

That's because that method

https://github.com/JuliaLang/LinearAlgebra.jl/blob/2c3fe9b7e0ca4e2c7bf506bd16ae5900f04a8023/src/factorization.jl#L116

is more specific than the intended method

https://github.com/JuliaLang/LinearAlgebra.jl/blob/2c3fe9b7e0ca4e2c7bf506bd16ae5900f04a8023/src/cholesky.jl#L683-L686

This PR fixes it by adding an extra, more specific method.

### DIFF
--- a/src/cholesky.jl
+++ b/src/cholesky.jl
@@ -680,12 +680,10 @@ end
 Base.propertynames(F::Cholesky, private::Bool=false) =
     (:U, :L, :UL, (private ? fieldnames(typeof(F)) : ())...)
 
-function _equal(C1::Cholesky, C2::Cholesky)
+function Base.:(==)(C1::C, C2::D) where {C<:Cholesky, D<:Cholesky}
     C1.uplo == C2.uplo || return false
     C1.uplo == 'L' ? (C1.L == C2.L) : (C1.U == C2.U)
 end
-Base.:(==)(C1::Cholesky, C2::Cholesky) = _equal(C1, C2)
-Base.:(==)(C1::T, C2::T) where {T<:Cholesky} = _equal(C1, C2)
 
 function getproperty(C::CholeskyPivoted{T}, d::Symbol) where {T}
     Cfactors = getfield(C, :factors)
--- a/src/cholesky.jl
+++ b/src/cholesky.jl
@@ -680,10 +680,12 @@ end
 Base.propertynames(F::Cholesky, private::Bool=false) =
     (:U, :L, :UL, (private ? fieldnames(typeof(F)) : ())...)
 
-function Base.:(==)(C1::Cholesky, C2::Cholesky)
+function _equal(C1::Cholesky, C2::Cholesky)
     C1.uplo == C2.uplo || return false
     C1.uplo == 'L' ? (C1.L == C2.L) : (C1.U == C2.U)
 end
+Base.:(==)(C1::Cholesky, C2::Cholesky) = _equal(C1, C2)
+Base.:(==)(C1::T, C2::T) where {T<:Cholesky} = _equal(C1, C2)
 
 function getproperty(C::CholeskyPivoted{T}, d::Symbol) where {T}
     Cfactors = getfield(C, :factors)
--- a/test/cholesky.jl
+++ b/test/cholesky.jl
@@ -583,6 +583,22 @@ end
         @test C.L == L
         @test C.U == L'
     end
+    let
+        B = [1.0 5.0; 3.0 4.0]
+        CA = Cholesky(LowerTriangular(A))
+        CB = Cholesky(LowerTriangular(B))
+        @test CA == CB
+        @test CA.L == CB.L
+        @test CA.U == CB.U
+    end
+    let
+        C = [1.0 2.0; 5.0 4.0]
+        CA = Cholesky(UpperTriangular(A))
+        CC = Cholesky(UpperTriangular(C))
+        @test CA == CC
+        @test CA.L == CC.L
+        @test CA.U == CC.U
+    end
 end
 
 @testset "adjoint of Cholesky" begin


src/cholesky.jl
```diff
@@ -680,10 +680,12 @@ end
 Base.propertynames(F::Cholesky, private::Bool=false) =
     (:U, :L, :UL, (private ? fieldnames(typeof(F)) : ())...)
 
-function Base.:(==)(C1::Cholesky, C2::Cholesky)
+function _equal(C1::Cholesky, C2::Cholesky)
```
dkarrasch: This would already solve the issue:

```suggestion
function Base.:(==)(C1::C, C2::D) where {C<:Cholesky, D<:Cholesky}
```

and doesn't require that internal function.

src/cholesky.jl
```diff
@@ -680,10 +680,12 @@ end
...
     C1.uplo == C2.uplo || return false
     C1.uplo == 'L' ? (C1.L == C2.L) : (C1.U == C2.U)
 end
+Base.:(==)(C1::Cholesky, C2::Cholesky) = _equal(C1, C2)
+Base.:(==)(C1::T, C2::T) where {T<:Cholesky} = _equal(C1, C2)
```
dkarrasch: In that case these are obsolete.

```suggestion
```

dkarrasch (commented):
Thanks for identifying the issue and the cause, and for proposing a fix. I have an alternative suggestion and another thought:

There are a few more instances that probably don't work as intended (or at least worth checking):

```julia
julia> methods(==, (Factorization, Factorization))
# 5 methods for generic function "==" from Base:
 [1] ==(A::LinearAlgebra.QRCompactWY, B::LinearAlgebra.QRCompactWY)
     @ LinearAlgebra ~/.julia/juliaup/julia-1.11.5+0.x64.linux.gnu/share/julia/stdlib/v1.11/LinearAlgebra/src/qr.jl:160
 [2] ==(C1::C, C2::D) where {C<:Cholesky, D<:Cholesky}
     @ LinearAlgebra REPL[2]:1
 [3] ==(A::Eigen, B::Eigen)
     @ LinearAlgebra ~/.julia/juliaup/julia-1.11.5+0.x64.linux.gnu/share/julia/stdlib/v1.11/LinearAlgebra/src/eigen.jl:664
 [4] ==(F::T, G::T) where T<:Factorization
     @ LinearAlgebra ~/.julia/juliaup/julia-1.11.5+0.x64.linux.gnu/share/julia/stdlib/v1.11/LinearAlgebra/src/factorization.jl:116
 [5] ==(x, y)
     @ Base.jl:207
```

So, `Eigen` and `QRCompactWY` are further candidates.

src/cholesky.jl
penelopeysm: Ah, indeed, that's fair!

penelopeysm:
> So, `Eigen` and `QRCompactWY` are further candidates.

Thanks for pulling that up! I took a look. I think `Eigen` is fine because it doesn't have special equality behaviour (it just compares the fields, and its fields don't carry any redundant information, so it seems to me it doesn't matter which method it dispatches to).

As for `QRCompactWY`, I have to confess I am completely baffled, because it dispatches to the correct method:

```julia
julia> m = [1.0 2.0; 3.0 4.0]
2×2 Matrix{Float64}:
 1.0  2.0
 3.0  4.0

julia> x = y = qr(m)
LinearAlgebra.QRCompactWY{Float64, Matrix{Float64}, Matrix{Float64}}
Q factor: 2×2 LinearAlgebra.QRCompactWYQ{Float64, Matrix{Float64}, Matrix{Float64}}
R factor:
2×2 Matrix{Float64}:
 -3.16228  -4.42719
  0.0      -0.632456

julia> @which x == y
==(A::LinearAlgebra.QRCompactWY, B::LinearAlgebra.QRCompactWY)
     @ LinearAlgebra ~/.julia/juliaup/julia-1.11.5+0.aarch64.apple.darwin14/share/julia/stdlib/v1.11/LinearAlgebra/src/qr.jl:160
```

in fact `Eigen` dispatches to the right method too:

```julia
julia> p = q = eigen(m)
Eigen{Float64, Float64, Matrix{Float64}, Vector{Float64}}
values:
2-element Vector{Float64}:
 -0.3722813232690143
  5.372281323269014
vectors:
2×2 Matrix{Float64}:
 -0.824565  -0.415974
  0.565767  -0.909377

julia> @which p == q
==(A::Eigen, B::Eigen)
     @ LinearAlgebra ~/.julia/juliaup/julia-1.11.5+0.aarch64.apple.darwin14/share/julia/stdlib/v1.11/LinearAlgebra/src/eigen.jl:664
```

whereas this doesn't happen with Cholesky (pre-this PR, of course):

```julia
julia> l = LowerTriangular(zeros(2, 2))
2×2 LowerTriangular{Float64, Matrix{Float64}}:
 0.0   ⋅
 0.0  0.0

julia> a = b = Cholesky(l)
Cholesky{Float64, Matrix{Float64}}
L factor:
2×2 LowerTriangular{Float64, Matrix{Float64}}:
 0.0   ⋅
 0.0  0.0

julia> @which a == b
==(F::T, G::T) where T<:Factorization
     @ LinearAlgebra ~/.julia/juliaup/julia-1.11.5+0.aarch64.apple.darwin14/share/julia/stdlib/v1.11/LinearAlgebra/src/factorization.jl:116
```

Do you have any idea why?

dkarrasch:
Ah, that's because the double-Cholesky method hasn't been released yet: https://github.com/JuliaLang/LinearAlgebra.jl/commit/c68619a4db8aead3eb46008d35f3a40b83b33014

It will be included in v1.12, so this PR can be closed (sorry for that).

dkarrasch:
I found out after realizing that on v1.11 there is no such method in the list of methods.

penelopeysm:
I was just looking into that! I did see the commit was from a year ago and I thought that it would have made it in by now, but clearly not.

Thanks for reviewing and sorry to have taken up your time :)

penelopeysm:
(Just out of curiosity, since I've never gone as far upstream as the standard library: Do changes in LinearAlgebra.jl only get merged into new minor versions of Julia, i.e., new Julia patch releases don't get bugfixes from standard libraries?)

dkarrasch:
That's actually a good point. That commit might be considered a bugfix, and in that case should be backported to the LTS and the current version. @jishnub What do you think?

BTW, you can see at the top if some commit is contained in a release. For the relevant commit, it just says `main` or `master`, otherwise it would mention some version (perhaps try older linalg commits on the main julia repo to see the difference).

> Thanks for reviewing and sorry to have taken up your time :)

No worries at all!

PR #1408, opened 2025-08-07
[1.12] Run `__init__()` even in sysimage builds

topolarity:
1.12-specific version of https://github.com/JuliaLang/LinearAlgebra.jl/pull/1407 (1.12 does not have LazyLibrary)

This change makes the LinearAlgebra initialization code run during the sysimage build, similar to how it would run when using LinearAlgebra at top-level in a normal package.

In this configuration, we can't rely on the serializer to isolate side-effects for us, so we have to manually ensure that any mutations that we perform here are not accidentally persisted to run-time. Accordingly, the `OncePerProcess` is here as a reminder that any code in this init has to think about phasing.

Requires https://github.com/JuliaLang/julia/pull/59232

Resolves the segfault in https://github.com/JuliaLang/julia/issues/59215

### DIFF
--- a/src/LinearAlgebra.jl
+++ b/src/LinearAlgebra.jl
@@ -823,7 +823,9 @@ function versioninfo(io::IO=stdout)
     return nothing
 end
 
-function __init__()
+# This OncePerProcess does not yield a useful value, but it causes side-effects (configuring LBT / BLAS)
+# which should only last for the duration of the current process.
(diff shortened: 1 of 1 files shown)


topolarity:
Closing in preference of https://github.com/JuliaLang/LinearAlgebra.jl/pull/1407#issuecomment-3167388152

PR #1409, opened 2025-08-08
Add bidiagonal cholesky version

### LINKED ISSUE #1383: Cholesky factorization of a Tridiagonal matrix should not be dense
The Cholesky factorization $A = G G^T$ of a banded matrix $A$ preserves the lower bandwidth, i.e., $G$ has the same lower bandwidth as $A$. In particular, the Cholesky factorization of a `Tridiagonal` matrix should be `Bidiagonal`. However, (at least) Julia 1.10 uses a dense matrix:

```julia
julia> using LinearAlgebra

julia> A = Tridiagonal(ones(4), 2 * ones(5), ones(4))
5×5 Tridiagonal{Float64, Vector{Float64}}:
 2.0  1.0   ⋅    ⋅    ⋅ 
 1.0  2.0  1.0   ⋅    ⋅ 
  ⋅   1.0  2.0  1.0   ⋅ 
  ⋅    ⋅   1.0  2.0  1.0
  ⋅    ⋅    ⋅   1.0  2.0

julia> cholesky(A)
Cholesky{Float64, Matrix{Float64}}
U factor:
5×5 UpperTriangular{Float64, Matrix{Float64}}:
 1.41421  0.707107  0.0       0.0       0.0
  ⋅       1.22474   0.816497  0.0       0.0
  ⋅        ⋅        1.1547    0.866025  0.0
  ⋅        ⋅         ⋅        1.11803   0.894427
  ⋅        ⋅         ⋅         ⋅        1.09545

julia> LinearAlgebra.cholcopy(A)
5×5 Matrix{Float64}:
 2.0  1.0  0.0  0.0  0.0
 1.0  2.0  1.0  0.0  0.0
 0.0  1.0  2.0  1.0  0.0
 0.0  0.0  1.0  2.0  1.0
 0.0  0.0  0.0  1.0  2.0
```

The issue seems to be mostly related to `cholcopy`, since the in-place factorization looks better (but I did not benchmark it in detail):

```julia
julia> cholesky!(A)
Cholesky{Float64, Tridiagonal{Float64, Vector{Float64}}}
U factor:
5×5 UpperTriangular{Float64, Tridiagonal{Float64, Vector{Float64}}}:
 1.41421  0.707107  0.0       0.0       0.0
  ⋅       1.22474   0.816497  0.0       0.0
  ⋅        ⋅        1.1547    0.866025  0.0
  ⋅        ⋅         ⋅        1.11803   0.894427
  ⋅        ⋅         ⋅         ⋅        1.09545
```

I consider this to be a performance bug since this makes the `cholesky` decomposition of a `Tridiagonal` matrix as slow as that of a dense matrix.

dkarrasch:
This adds a generic version of cholesky for tridiagonal matrices. This fixes #1383, and massively boosts performance both for `(Sym)Tridiagonal`s. Previously, we haven't had a kernel for the symtridiagonal case.

### DIFF
--- a/src/bidiag.jl
+++ b/src/bidiag.jl
@@ -1472,7 +1472,14 @@ function _chol!(A::Bidiagonal{<:BlasFloat,<:StridedVector}, ::Type{UpperTriangul
     d = real(A.dv)
     e = A.ev
     dv, ev = LAPACK.pttrf!(d, e)
-    map!(sqrt, dv)
+    for k in eachindex(dv)
+        Akk = dv[k]
+        Akk, info = _chol!(Akk, UpperTriangular)
+        if info != 0
+            return UpperTriangular(A), convert(BlasInt, k)
+        end
+        dv[k] = Akk
+    end
     @views ev .*= dv[1:end-1]
     U = Bidiagonal(dv, ev, :U)
     return UpperTriangular(U), convert(BlasInt, 0)
@@ -1507,7 +1514,14 @@ function _chol!(A::Bidiagonal{<:BlasFloat,<:StridedVector}, ::Type{LowerTriangul
     d = real(A.dv)
     e = A.ev
     dv, ev = LAPACK.pttrf!(d, e)
-    map!(sqrt, dv)
+    for k in eachindex(dv)
+        Akk = dv[k]
+        Akk, info = _chol!(Akk, LowerTriangular)
+        if info != 0
+            return LowerTriangular(A), convert(BlasInt, k)
+        end
+        dv[k] = Akk
+    end
     @views ev .*= dv[1:end-1]
     L = Bidiagonal(dv, ev, :L)
     return LowerTriangular(L), convert(BlasInt, 0)
--- a/src/bidiag.jl
+++ b/src/bidiag.jl
@@ -1468,89 +1468,32 @@ function inv(B::Bidiagonal{T}) where T
 end
 
 # cholesky-version for (sym)tridiagonal matrices
-function _chol!(A::Bidiagonal{<:BlasFloat,<:StridedVector}, ::Type{UpperTriangular})
-    d = real(A.dv)
-    e = A.ev
-    dv, ev = LAPACK.pttrf!(d, e)
-    for k in eachindex(dv)
-        Akk = dv[k]
-        Akk, info = _chol!(Akk, UpperTriangular)
-        if info != 0
-            return UpperTriangular(A), convert(BlasInt, k)
-        end
-        dv[k] = Akk
-    end
-    @views ev .*= dv[1:end-1]
-    U = Bidiagonal(dv, ev, :U)
-    return UpperTriangular(U), convert(BlasInt, 0)
-end
-function _chol!(A::Bidiagonal, ::Type{UpperTriangular})
-    require_one_based_indexing(A)
-    n = checksquare(A)
-    realdiag = eltype(A) <: Complex
-    dv = A.dv
-    ev = A.ev
-    @inbounds begin
-        for k = 1:n
-            Akk = realdiag ? real(dv[k]) : dv[k]
-            if k > 1
-                Akk -= realdiag ? abs2(ev[k-1]) : ev[k-1]'ev[k-1]
-            end
-            dv[k] = Akk
+for (T, uplo) in ((:UpperTriangular, :(:U)), (:LowerTriangular, :(:L)))
+    @eval function _chol!(A::Bidiagonal, ::Type{$T})
+        dv = real(A.dv)
+        ev = A.ev
+        n = length(dv)
+        @inbounds for i in 1:n-1
+            iszero(dv[i]) && throw(ZeroPivotException(i))
+            ev[i] /= dv[i]
+            dv[i+1] -= abs2(ev[i])*dv[i]
+            Akk = dv[i]
             Akk, info = _chol!(Akk, UpperTriangular)
             if info != 0
-                return UpperTriangular(A), convert(BlasInt, k)
+                return $T(A), convert(BlasInt, i)
             end
-            dv[k] = Akk
-            AkkInv = inv(copy(Akk'))
-            if k < n
-                ev[k] = AkkInv*ev[k]
-            end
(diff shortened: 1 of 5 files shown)


araujoms:
Getting the generic version is even better than just calling `pttrf`, but isn't that worth it for `BlasFloat`?

src/tridiag.jl
```diff
@@ -1039,7 +1039,7 @@ function dot(x::AbstractVector, A::Tridiagonal, y::AbstractVector)
     return r
 end
 
-function cholesky(S::SymTridiagonal, ::NoPivot = NoPivot(); check::Bool = true)
+function cholesky(S::Union{SymTridiagonal,Tridiagonal}, ::NoPivot = NoPivot(); check::Bool = true)
```
araujoms, line 1042: Shouldn't there be also a method for `Hermitian{<:Number,<:Bidiagonal}`? It is a bit weird to let that go to the fallback.

test/cholesky.jl
```diff
@@ -492,9 +492,16 @@ end
...
-    C = cholesky(S)
-    @test C.L * C.U ≈ S
+    for M in (SymTridiagonal(fill(2.0, 4), ones(3)),
+        Tridiagonal(ones(3), fill(2.0, 4), ones(3)),
+        )
+        C = cholesky(M)
+        @test C.L * C.U ≈ M
+    end
+    # test LowerTriangular version
+    M = Hermitian(Bidiagonal(fill(2.0, 4), ones(3), :L), :L)
```
araujoms:
```suggestion
    M = Hermitian(Bidiagonal(fill(2.0, 4), im * ones(3), :L), :L)
```
Your code also works for the complex case, I think it's worth testing it there.

src/tridiag.jl
dkarrasch: I'll double check, but when the outermost wrapper is `Hermitian`, a different method gets called, basically unwrapping the data and appending `(Upper/Lower)Triangular`, depending on which triangle is viewed by the `Hermitian` wrapper.

dkarrasch:
Thanks for your review. That was helpful!

> Getting the generic version is even better than just calling `pttrf`, but isn't that worth it for `BlasFloat`?

I'll need to benchmark that.

src/bidiag.jl
```diff
@@ -1467,6 +1467,92 @@ function inv(B::Bidiagonal{T}) where T
...
+            end
+            dv[k] = Akk
+            AkkInv = inv(copy(Akk'))
+            if k < n
+                ev[k] *= AkkInv
+            end
+        end
+     end
+    return LowerTriangular(A), convert(BlasInt, 0)
+end
```
araujoms:
```suggestion
for tritype in (:UpperTriangular, :LowerTriangular)
    @eval begin
        function _chol!(A::Bidiagonal{<:BlasFloat,<:StridedVector}, ::Type{$tritype})
            d = real(A.dv)
            e = A.ev
            dv, ev = LAPACK.pttrf!(d, e)
            for k in eachindex(dv)
                Akk = dv[k]
                Akk, info = _chol!(Akk, $tritype)
                if info != 0
                    return $tritype(A), convert(BlasInt, k)
                end
                dv[k] = Akk
            end
            @views ev .*= dv[1:end-1]
            R = Bidiagonal(dv, ev, A.uplo)
            return $tritype(R), convert(BlasInt, 0)
        end
        function _chol!(A::Bidiagonal, ::Type{$tritype})
            require_one_based_indexing(A)
            n = checksquare(A)
            realdiag = eltype(A) <: Complex
            dv = A.dv
            ev = A.ev
            @inbounds begin
                for k = 1:n
                    Akk = realdiag ? real(dv[k]) : dv[k]
                    if k > 1
                        Akk -= realdiag ? abs2(ev[k-1]) : ev[k-1]'ev[k-1]
                    end
                    dv[k] = Akk
                    Akk, info = _chol!(Akk, $tritype)
                    if info != 0
                        return $tritype(A), convert(BlasInt, k)
                    end
                    dv[k] = Akk
                    AkkInv = inv(copy(Akk'))
                    if k < n
                        ev[k] *= AkkInv
                    end
                end
            end
            return $tritype(A), convert(BlasInt, 0)
        end
    end
end
```
I think it would be better to avoid code duplication.

src/bidiag.jl
```diff
@@ -1467,6 +1467,92 @@ function inv(B::Bidiagonal{T}) where T
...
+    e = A.ev
+    dv, ev = LAPACK.pttrf!(d, e)
+    for k in eachindex(dv)
+        Akk = dv[k]
+        Akk, info = _chol!(Akk, UpperTriangular)
+        if info != 0
+            return UpperTriangular(A), convert(BlasInt, k)
+        end
+        dv[k] = Akk
+    end
```
araujoms: Isn't this redundant? As far as I know `pttrf` already checks whether the input is positive definite.

src/bidiag.jl
dkarrasch: Actually, I ended up "inlining" the ldlt-code and merge it with the rest, so that everything is in one loop, and performance equals the LAPACK-version.

src/bidiag.jl
dkarrasch: That's the reason I ended up not calling LAPACK, because we want to be able to set the `check=false` keyword and not make it throw.

dkarrasch:
So, to give an update: I replaced the LAPACK-call to the ldlt-routine by the Julia code that we use in the ldlt-module. Apparently, we don't reach out to LAPACK for that either, even though `pttrf!` exists. I merged that loop with your corresponding two lines `map!(sqrt, d); e .*= d`. For `Float64`, performance is the same as in the last but one commit with the explicit LAPACK calls, which is faster than my initial simple "reduction of the dense algorithm". The reason, after all, for that is that we want to be able to avoid throws when `check=false`, which would be impossible with the LAPACK call.

araujoms:
It would be possible to edit the `pttrf` wrapper, as it was done with the `potrf` wrapper, but since the performance is similar it's better to avoid a ccall anyway.

araujoms (approved):

PR #1410, opened 2025-08-10
Generalize 3-arg `dot` to `HermOrSym`

dkarrasch:
This was unnecessarily restrictive, and didn't handle, e.g., complex symmetric and quaternionic hermitian matrices.

### DIFF
--- a/src/symmetric.jl
+++ b/src/symmetric.jl
@@ -229,6 +229,9 @@ const SelfAdjoint = Union{SymTridiagonal{<:Real}, Symmetric{<:Real}, Hermitian}
 wrappertype(::Union{Symmetric, SymTridiagonal}) = Symmetric
 wrappertype(::Hermitian) = Hermitian
 
+hswrapperop(::Symmetric) = symmetric
+hswrapperop(::Hermitian) = hermitian
+
 nonhermitianwrappertype(::SymSymTri{<:Real}) = Symmetric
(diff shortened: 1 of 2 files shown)


araujoms:
Your PR breaks the 3-argument dot product of Hermitian/Symmetric matrices of matrices, which currently works:
```julia
julia> m = Symmetric([randn(ComplexF64, 2, 2) for i in 1:2, j in 1:2]);

julia> x = [randn(ComplexF64, 2) for i in 1:2];

julia> y = [randn(ComplexF64, 2) for i in 1:2];

julia> dot(x,m,y)
```

stevengj:
I guess `dot` is not required to implement an inner product (even in the Hermitian case, we don't require positive-definiteness), so broadening it to complex-symmetric etcetera seems okay.

araujoms (approved):

stevengj:
That being said, $x^* A y$ is a pretty unnatural operation to perform if $A = A^T$ is complex-symmetric.  A more natural operation in that case would be $x^T A y$, i.e. the unconjugated `dotu` operation ([julia#8300](https://github.com/JuliaLang/julia/issues/8300)).

dkarrasch:
The complex-symmetric case is more like a byproduct. This is mostly about making the Hermitian case available to non-`RealOrComplex`. What you're saying, would that be a 3-arg `dotu` method?

PR #1412, opened 2025-08-11
Fix signature in generic `cholesky!`

dkarrasch:
The type alias included `SymTridiagonal`, which doesn't make sense here, (edit: and wasn't working anyways). This has not been released.


dkarrasch:
Doesn't need backporting. The addition of `SymTridiagonal` to `const SelfAdjoint` hasn't been done on the release-v1.12 branch.

PR #1418, opened 2025-08-14
Fix overflow and type issues with generic det of a triangular matrix

### LINKED ISSUE #1415: DimensionError for determinant of generic, triangular, unitful matrices
```jl
using Unitful
using LinearAlgebra
det([1 0; 0 1]*u"m")
```

leads to
```
ERROR: DimensionError: m and 1 m² are not dimensionally compatible.
Stacktrace:
 [1] convert(::Type{Quantity{Float64, 𝐋, Unitful.FreeUnits{(m,), 𝐋, nothing}}}, x::Quantity{Int64, 𝐋², Unitful.FreeUnits{(m²,), 𝐋², nothing}})
   @ Unitful ~/.julia/packages/Unitful/sxiHA/src/conversion.jl:133
 [2] det(A::Matrix{Quantity{Int64, 𝐋, Unitful.FreeUnits{(m,), 𝐋, nothing}}})
   @ LinearAlgebra ~/.julia/juliaup/julia-1.11.6+0.aarch64.apple.darwin14/share/julia/stdlib/v1.11/LinearAlgebra/src/generic.jl:1649
 [3] top-level scope
   @ REPL[24]:1
```

I suspect that it is because this type handling should be using `oneunit` instead of `one`
https://github.com/JuliaLang/LinearAlgebra.jl/blob/7e647cf4a6d35654118b7e732c7befc3c3deaa10/src/generic.jl#L1848

Although it isn't entirely clear to me why conversion is necessary, here is a potential fix that keeps the code similar but fixes this edge case
```jl
function mydet(A::AbstractMatrix{T}) where {T}
    if istriu(A) || istril(A)
        S = promote_type(T, typeof((oneunit(T)*zero(T) + zero(T)*zero(T))/one(T)))
        return convert(S, det(UpperTriangular(A)))
    end
    return det(lu(A; check = false))
end
mydet([1 0; 0 1]*u"m") # returns 1.0 m²
```

lxvm:
Fixes #1415


lxvm:
Instead of using the number type as a constructor, I switched back to using `convert` because this fixed some broken tests for minimal number types that did not define the constructor. I used `Base.Fix(convert, S)` rather than `y -> convert(S, y)` because the latter is not type stable.

test/generic.jl
```diff
@@ -96,6 +97,18 @@ n = 5 # should be odd
     @testset "det with nonstandard Number type" begin
         elty <: Real && @test det(Dual.(triu(A), zero(A))) isa Dual
     end
+    if elty <: Int
+        @testset "det no overflow - triangular" begin
+            A = diagm([typemax(elty), typemax(elty)])
+            @test det(A) ≈ det(float(A))
```
stevengj:
```suggestion
            @test det(A) == det(float(A))
```
Should be exactly equal, no?

test/generic.jl
```diff
@@ -96,6 +97,18 @@ n = 5 # should be odd
...
+        end
+    end
+    @testset "det with units - triangular" begin
+        for dim in 1:10
```
stevengj:
```suggestion
        for dim in 0:4
```
should suffice?

stevengj: Updated to test `for dim in 0:4` to make sure that the 0x0 case is correct…

test/generic.jl
lxvm: Yes

test/generic.jl
lxvm: Good catch, I needed to provide the `init` keyword to `prod`

stevengj: You still need to update the test by committing the suggestion above.

lxvm: Added, sorry!

stevengj (approved):

stevengj:
The codecov failure seems spurious.

PR #1419, opened 2025-08-20
Improve type-stability of `eigencopy_oftype`

topolarity:
Without the `where` clause, this can end up inferred / invoked as e.g. `eigencopy_oftype(::Symmetric{ComplexF64, ...}, ::Type)::Matrix` (eltype is lost)


ViralBShah:
@dkarrasch Is it simplest to just cherry pick this directly onto the release-1.12 branch, or should we do the backports?

dkarrasch:
Usually one would want to run a CI check before merging/cherry-picking into a release branch, but this is simple enough to be cherry-picked directly, I think.

PR #1429, opened 2025-08-25
Backports to release v1.12

dkarrasch:
- [x] #1428


ViralBShah:
Need to run BumpStdlibs to get this on to the Julia repo as 1.12 gets closer. I don't think there's anything else that needs to be backported for 1.12.

Issue #1438, opened 2025-09-10
missing hermitianpart(x::Number) = real(x)


stevengj:
It seems like this should be defined for numbers, since `hermitianpart` is just a generalization of the real part (indeed, in some literature it is called the ["operator real part"](https://github.com/JuliaLang/julia/pull/31836#issuecomment-967334233)):
```jl
julia> hermitianpart(3+4im)
ERROR: MethodError: no method matching hermitianpart(::Complex{Int64})
```
Should be easy for someone to add a `hermitianpart(x::Number) = real(x)`, update the docstring, and add a test.

(This came up in #1437 for generic code on `Hermitian(::Diagonal)`.  Note that the original PR for `hermitianpart` included a `::Number` method, but it was removed by @dkarrasch at the time because the application was unclear, but with the option of adding it back: https://github.com/JuliaLang/julia/pull/31836#discussion_r1094172213)

GianmarcoCuppari:
Hi @stevengj ! I've worked on implementing the requested hermitianpart method for numbers. Here's what I've done:

- Added the method: hermitianpart(x::Number) = real(x) in src/symmetric.jl
- Updated documentation: Added a note in the docstring explaining that for numbers, hermitianpart returns the real part
- Added tests: Included test cases to verify the behavior with complex numbers

This should resolve the MethodError you encountered with hermitianpart(3+4im) and support the generic code usage in #1437.
I'm new to open source contributions and wanted to make sure I'm on the right track before submitting the PR. Could you confirm this covers everything needed? I'd appreciate any feedback to ensure I'm following the project's conventions correctly.
Thanks for your patience as I learn the contribution process!

dkarrasch:
Hi @GianmarcoCuppari, thanks for taking up the initiative, that's much appreciated. Please submit a pull request with what you have. Then others can review the changes, give feedback or suggest concrete changes. A pull request doesn't have to be "perfect" right from the start, and working on the details is usually insightful and rewarding, at least that's my personal experience.

GianmarcoCuppari:
Thank you for the kind words and encouragement! 
I’ve just submitted the pull request.
Looking forward to the review and happy to make any adjustments if needed.

PR #1439, opened 2025-09-11
Ensure inverse of Symmetric{<:,Diagonal} returns same type

### LINKED ISSUE #1437: Inverse of `Symmetric{ , Diagonal}` returns a dense matrix
```julia
using LinearAlgebra

julia> A = I(2)
2×2 Diagonal{Bool, Vector{Bool}}:
 1  ⋅
 ⋅  1

julia> inv(A) # correct, diagonal matrix
2×2 Diagonal{Float64, Vector{Float64}}:
 1.0   ⋅ 
  ⋅   1.0

julia> inv(Symmetric(A)) # dense matrix?
2×2 Matrix{Float64}:
 1.0  0.0
 0.0  1.0
```

possible fix:

```julia
inv(A::Symmetric{<:Any, <:Diagonal}) = inv(A.data)
```

I'd be happy to open a PR if this seems reasonable.

vandenman:
Fixes https://github.com/JuliaLang/LinearAlgebra.jl/issues/1437

Does not add a test for 1438, but let me know if I should.

Feedback is welcome!

### DIFF
--- a/src/diagonal.jl
+++ b/src/diagonal.jl
@@ -970,6 +970,10 @@ function inv(D::Diagonal{T}) where T
     Diagonal(Di)
 end
 
+# Ensure doubly wrapped matrices use efficient diagonal methods and return a Symmetric/Hermitian type
+inv(A::Symmetric{<:Number,<:Diagonal}) = Symmetric(inv(A.data), sym_uplo(A.uplo))
+inv(A::Hermitian{<:Number,<:Diagonal}) = Hermitian(inv(real(A.data)), sym_uplo(A.uplo))
+
 function pinv(D::Diagonal{T}) where T
     Di = similar(D.diag, typeof(inv(oneunit(T))))
     for i = 1:length(D.diag)
--- a/src/symmetric.jl
+++ b/src/symmetric.jl
@@ -987,6 +987,7 @@ See also [`hermitianpart!`](@ref) for the corresponding in-place operation.
     This function requires Julia 1.10 or later.
 """
 hermitianpart(A::AbstractMatrix, uplo::Symbol=:U) = Hermitian(_hermitianpart(A), uplo)
+hermitianpart(a::Number) = real(a)
 
 """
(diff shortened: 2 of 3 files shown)


araujoms:
You do need to add a test for #1438.

You probably want to preserve the type of the number in `hermitianpart`, to match the behaviour when acting on 1x1 matrices. With something like `_hermitianpart(a::T) where {T<:Number} = T(real(a))`.

You probably want to do something similar for `inv(A::Hermitian{<:Number,<:Diagonal})`, it's weird for a complex matrix to suddenly become real.

stevengj:
Since #1438 is not needed for this PR, it should be a separate PR.

src/diagonal.jl
```diff
@@ -970,6 +970,10 @@ function inv(D::Diagonal{T}) where T
     Diagonal(Di)
 end
 
+# Ensure doubly wrapped matrices use efficient diagonal methods and return a Symmetric/Hermitian type
+inv(A::Symmetric{<:Number,<:Diagonal}) = Symmetric(inv(A.data), sym_uplo(A.uplo))
+inv(A::Hermitian{<:Number,<:Diagonal}) = Hermitian(inv(real(A.data)), sym_uplo(A.uplo))
```
dkarrasch, line 975: To avoid the issue of returning a real-eltype matrix for a complex-eltype argument, this should perhaps be

```suggestion
inv(A::Hermitian{T,S}) where {T<:Number, S<:Diagonal} = Hermitian{T,S}(inv(real(A.data)), A.uplo)
```
?

vandenman: I suppose that's possible, but a Hermitian Diagonal matrix is actually always a real Diagonal matrix, so we can deduce from the type alone that the inverse is always a real diagonal matrix. This also holds for quaternion and octonion valued matrices. If there is a number type where becomes a problem, then using `real` might also be a problem.

PR #1445, opened 2025-09-16
Add hermitianpart method for Number types

### LINKED ISSUE #1438: missing hermitianpart(x::Number) = real(x)

GianmarcoCuppari:
Fixes JuliaLang/LinearAlgebra.jl#1438

This PR implements `hermitianpart` for `Number` types as requested in the issue.

**Changes:**
- Added `hermitianpart(x::Number) = real(x)` method in `src/symmetric.jl`
- Updated docstring to document the behavior for numbers  
- Added test cases for complex number inputs in `test/symmetric.jl`

This resolves the `MethodError` when calling `hermitianpart` on complex numbers and enables the generic code usage mentioned in the issue.


test/symmetric.jl
```diff
@@ -1035,6 +1035,14 @@ end
...
         @test Aherm.uplo == LinearAlgebra.char_uplo(uplo)
     end
+    @testset "hermitianpart for numbers" begin
+    @test hermitianpart(3 + 4im) == 3
+    @test hermitianpart(5) == 5
+    @test hermitianpart(2.5) == 2.5
+    @test hermitianpart(-1 + 0im) == -1
+    @test typeof(hermitianpart(3 + 4im)) == Int
+    @test typeof(hermitianpart(2.5 + 3im)) == Float64
+end
```
dkarrasch: These lines should get indented by 4 spaces each.

src/symmetric.jl
```diff
@@ -987,6 +988,7 @@ See also [`hermitianpart!`](@ref) for the corresponding in-place operation.
     This function requires Julia 1.10 or later.
 """
 hermitianpart(A::AbstractMatrix, uplo::Symbol=:U) = Hermitian(_hermitianpart(A), uplo)
+hermitianpart(x::Number) = real(x)
```
dkarrasch, line 1020: Should this perhaps be

```suggestion
hermitianpart(x::Number) = float(real(x))
```

to mimick the `(x + x')/2` behavior of arrays

```julia
julia> hermitianpart([5;;])
1×1 Hermitian{Float64, Matrix{Float64}}:
 5.0
```

?

dkarrasch (commented):
Thanks for your PR!

test/symmetric.jl
```diff
@@ -1035,6 +1035,14 @@ end
         @test Aherm isa Hermitian
         @test Aherm.uplo == LinearAlgebra.char_uplo(uplo)
     end
+    @testset "hermitianpart for numbers" begin
+    @test hermitianpart(3 + 4im) == 3
+    @test hermitianpart(5) == 5
+    @test hermitianpart(2.5) == 2.5
+    @test hermitianpart(-1 + 0im) == -1
+    @test typeof(hermitianpart(3 + 4im)) == Int
```
dkarrasch: After the `float` change, this needs to be changed to

```suggestion
    @test typeof(hermitianpart(3 + 4im)) == Float64
```

src/symmetric.jl
```diff
@@ -987,6 +988,7 @@ See also [`hermitianpart!`](@ref) for the corresponding in-place operation.
     This function requires Julia 1.10 or later.
 """
 hermitianpart(A::AbstractMatrix, uplo::Symbol=:U) = Hermitian(_hermitianpart(A), uplo)
+hermitianpart(x::Number) = float(real(x))
```
```suggestion
hermitianpart(x::Number) = real(x)
```
I don't think `float` is necessary?

src/symmetric.jl
stevengj: I'm not sure I agree.  Why does it need to match the type returned for arrays?

stevengj: And even for arrays, the type is not always floating-point:
```jl
julia> hermitianpart([5//1;;])
1×1 Hermitian{Rational{Int64}, Matrix{Rational{Int64}}}:
 5
```

dkarrasch: It was more like an open question. The association that came to my mind is the norm of numbers, which we also promote to floating-point types, perhaps to mimick the `sqrt(x^2)` computation, without actually computing it.

stevengj: The `norm` of a `Complex` number has to be a floating-point type regardless.  And `norm(::Integer)` also arguably should be floating-point to avoid spurious overflow for fixed-width integers (consider `norm(Int8(-128))`).  Similarly for `norm(::Rational)`.   So there's a good argument for `norm` to be floating point even for scalars, especially since the generic `norm` on arrays is defined in terms of calling `norm` recursively on the elements.

This isn't the case for `hermitianpart`, since `real(x)` should always succeed without additional conversions.

src/symmetric.jl
dkarrasch: As per the discussion above, this should be the right solution. Sorry for the back and forth.

GianmarcoCuppari: I'll proceed with the modification!

For informational purposes:
real(x) was chosen for its advantages in preserving precision and avoiding unnecessary type conversions, rather than float(real(x)), which would force a conversion to Float64 and risk losing precision for rational numbers?

Thanks

araujoms:
I think we should have `hermitianpart(x::Number) = (x + conj(x))/2` in order to match the type behaviour of the matrix case. Then we have `hermitianpart(im) === 0.0 + 0.0im`, `hermitianpart(1) === 1.0` and `hermitianpart(1//1) === 1//1`.

GianmarcoCuppari:
I see the point, but I’d lean towards keeping hermitianpart(x::Number) = real(x) for scalars.
This way we avoid unnecessary operations and preserve the exact type (Int, Rational, etc.), which feels more natural for numbers.
Matrices already use (A + A')/2, so having slightly different behavior for arrays vs. scalars might actually be fine here.

What do you think @stevengj ? Could this make sense?

stevengj:
I think the general principle here is that we shouldn't change the type unless it is required to represent the output, since type conversions may incur a loss of data.

araujoms:
`real(x)` does change the type of `x`, that's what bothers me. If we use `hermitianpart` to handle both matrices and scalars - which is the whole point of this function - we want the typing between both cases to be consistent.

I think it would be the perfect solution for the sister PR #1439, which currently has an unsatisfactory behavior:
```julia
julia> Z = complex([1 0;0 2]);

julia> inv(Diagonal(Z))
2×2 Diagonal{ComplexF64, Vector{ComplexF64}}:
 1.0-0.0im      ⋅    
     ⋅      0.5-0.0im

julia> inv(Hermitian(Diagonal(Z)))
2×2 Hermitian{Float64, Diagonal{Float64, Vector{Float64}}}:
 1.0   ⋅ 
  ⋅   0.5
```

stevengj:
> `real(x)` does change the type of `x`

It doesn't change the underlying numeric type — it just extracts a component, like `a[i]` for an array, without doing any conversion or losing any data from that component.

> If we use hermitianpart to handle both matrices and scalars - which is the whole point of this function - we want the typing between both cases to be consistent.

Why?  You want the *meaning* between both cases to be consistent, so that you can use it in generic code.  But the return type is different anyway, so what does it matter if the matrix case (by necessity) *also* does a floating-point conversion?

araujoms:
Because we'd get type instability.

stevengj:
> Because we'd get type instability.

It's not a type instability to return different output types for different input types.

araujoms:
It's bound to happen with any non-trivial code. It's very common to have special cases for diagonal, symmetric, Hermitian matrices in LinearAlgebra, where we wrap the matrix, do the computation, and unwrap it again. I fixed a lot of type instabilities resulting from this: #1360

Letting the inverse of a complex matrix be a real matrix is just asking for trouble.

stevengj:
> It's bound to happen with any non-trivial code. 

*What* is bound to happen?   

`hermitianpart(::Number)` will return a `Number`, and `hermitianpart(::AbstractMatrix)` will return a `Hermitian` matrix.   This is not a type instability.  What does it matter if the latter is floating-point and the former is not (which is also not a type instability)?

What specific problem are you worried about?

> Letting the inverse of a complex matrix be a real matrix is just asking for trouble.

I agree that it's asking for trouble, because the inverse of a complex matrix is generally not real.  I don't understand what that has to do with the current PR.

dkarrasch:
I think it makes sense to have `hermitianpart(x::Number) = real(x)`. In a "recursive" setting (like where this was originally (thought to be) needed), the real number would get promoted automatically. Also, the automatic promotion already exists as

```julia
hermitian(A::Number, ::Symbol=:U) = convert(typeof(A), real(A))
```

GianmarcoCuppari:
Thanks for the clarification!  

Just to confirm, do you think this PR is now ready to merge, or are there any further changes you’d like me to make regarding the `hermitianpart` handling for numbers?

GianmarcoCuppari:
I just saw that the tests on this PR aren't working.

Do you think this could be due to code changes @dkarrasch ?

Also, I'm still not sure if you prefer to keep "real." Taking a closer look, it seems like having "real" makes the most sense.

Looking forward to hearing from you,
Thanks!

README.md
```diff
@@ -1,5 +1,24 @@
...
+**Original repository:** [JuliaLang/LinearAlgebra.jl](https://github.com/JuliaLang/LinearAlgebra.jl)
+
+---
+
+# LinearAlgebra
+
+This package is part of the Julia standard library (stdlib).
+
+LinearAlgebra.jl provides functionality for performing linear algebra operations in Julia.
+
```
dkarrasch: Can you please remove all these changes?

src/symmetric.jl
```diff
@@ -1003,19 +1003,26 @@ end
...
 Return the Hermitian part of the square matrix `A`, defined as `(A + A') / 2`, as a
 [`Hermitian`](@ref) matrix. For real matrices `A`, this is also known as the symmetric part
 of `A`; it is also sometimes called the "operator real part". The optional argument `uplo` controls the corresponding argument of the
 [`Hermitian`](@ref) view. For real matrices, the latter is equivalent to a
 [`Symmetric`](@ref) view.
 
+For scalar inputs `x`, the function returns the real part of `x`. Standard integer scalars
+are promoted to `Float64` to align with the behavior of 1×1 Hermitian matrices, while other numeric types
+(`Float32`, `BigInt`, `Rational`, `BigFloat`) are preserved.
+
```
dkarrasch: I'd suggest to leave this out, too.

src/symmetric.jl
```diff
@@ -1003,19 +1003,26 @@ end
...
 See also [`hermitianpart!`](@ref) for the corresponding in-place operation.
 
 !!! compat "Julia 1.10"
     This function requires Julia 1.10 or later.
 """
+
```
dkarrasch:
```suggestion
```

The docstring needs to be attached to the function definition, otherwise Julia doesn't know where it belongs.

src/symmetric.jl
```diff
@@ -1003,19 +1003,26 @@ end
...
 hermitianpart(A::AbstractMatrix, uplo::Symbol=:U) = Hermitian(_hermitianpart(A), uplo)
+hermitianpart(x::Number) = float(real(x))
```
dkarrasch:
```suggestion
hermitianpart(x::Number) = real(x)
```

I believe this will get promoted if necessary by other mechanisms, so there is no need to enforce it per se.

src/symmetric.jl
```diff
@@ -1052,4 +1059,4 @@ function Base.replace_in_print_matrix(A::HermOrSym,i::Integer,j::Integer,s::Abst
     ijminmax = minmax(i, j)
     inds = A.uplo == 'U' ? ijminmax : reverse(ijminmax)
     Base.replace_in_print_matrix(parent(A), inds..., s)
-end
+end
```
dkarrasch:
```suggestion
end

```

It's common style to have an empty line at the end. No need to modify this.

test/symmetric.jl
```diff
@@ -1035,6 +1035,16 @@ end
...
     end
+    @testset "hermitianpart for numbers" begin
+    @test hermitianpart(3 + 4im) == 3
+    @test hermitianpart(5) == 5.0
+    @test typeof(hermitianpart(5)) == Float64
+    @test hermitianpart(2.5) == 2.5
+    @test hermitianpart(-1 + 0im) == -1
+    @test typeof(hermitianpart(3 + 4im)) == 3.0
+    @test typeof(hermitianpart(3 + 4im)) == Float64
+    @test typeof(hermitianpart(2.5 + 3im)) == Float64
```
dkarrasch:
```suggestion
    @test hermitianpart(2.5) == 2.5
    @test hermitianpart(-1 + 0im) == -1
    @test typeof(hermitianpart(3 + 4im)) == 3.0
```

These type tests wouldn't be needed anymore.

dkarrasch (dismissed):
Here are a few comments. Once that is settled, we should be able to merge.

GianmarcoCuppari:
Apologies for the formatting mistakes!! I've applied all the suggested changes. Hopefully everything looks good now. Thanks for the review and guidance!

README.md
```diff
@@ -1,9 +1,3 @@
-# LinearAlgebra
-
-This package is part of the Julia standard library (stdlib).
-
-`LinearAlgebra.jl` provides functionality for performing linear algebra operations in Julia.
-
```
dkarrasch, line 6: Can you please undo this change. Eventually, there should be changes only in the two `symmetric.jl` files.

EDIT: Sorry, I see I myself marked too much for removal above.

dkarrasch:
Thanks @GianmarcoCuppari for your contribution and your patience, and sorry for the back and forth.

PR #1456, opened 2025-09-19
Make `dot` with Bool-arrays type-stable

### LINKED ISSUE #1430: 3-arg `dot` is type-unstable for matrices with eltype `Bool`
This is due to the fact that when we initialize the zero value, we use the eltypes or, alternatively, the first values, before we start to add up further `dot` products. We also have this inconsistency:

```julia
julia> using LinearAlgebra

julia> dot(true, true)
true

julia> dot([true], [true])
1
```

A minimal fix would be to define

```julia
dot(x::Bool, y::Bool) = x*y + false
```

which changes the return type for scalar arguments, but renders the matrix cases (both 2-arg and 3-arg) type-stable. What do you think?

dkarrasch:
Fixes #1430.

### DIFF
--- a/src/generic.jl
+++ b/src/generic.jl
@@ -996,7 +996,8 @@ function dot(x, y) # arbitrary iterables
     return s
 end
 
-dot(x::Number, y::Number) = conj(x) * y
+# the unary + is for type promotion in the Boolean case, mimicking the reduction in usual dot
+dot(x::Number, y::Number) = +(conj(x) * y)
 
 function dot(x::AbstractArray, y::AbstractArray)
     lx = length(x)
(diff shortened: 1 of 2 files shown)


stevengj (approved):

ViralBShah:
Do we want a 1.12 backport on this?

dkarrasch:
The "issue" is very old, so if this was backportable (or of interest), we should backport it also to the LTS. OTOH, this is so narrow, I'm not sure it's worth risking creating hassle.

ViralBShah:
Doesn’t seem worth the hassle for LTS for sure.

PR #1458, opened 2025-09-22
Avoid divide the first element for type check in normalize

### LINKED ISSUE #1182: Find a better solution for determining element type for the return array in e.g. `normalize`
Coming from: https://github.com/JuliaLang/julia/pull/55830

Making this an issue rather than PR because there seems to be some question about what the solution should actually be, and it might be broader than this one method.

In the `normalize` stack, the existing code computes the result on one element, finds the type, and then that becomes the element type for the output array. I believe this is done to be generic for operations on arrays of arrays. However, relying on the first element means the type of the result has a hidden dependency on element order, and the scalar lookup means the code can't run on a GPU.

albertomercurio:
Closes #1182

The previous implementation of `normalize` was computing the division between the first element of the array and the norm, in order to obtain the output type. This is both inefficient and doesn't work on arrays that don't support scalar indexing.

I have changed the type check with `float(Base.promote_eltype(a, nrm))`.

### DIFF
--- a/src/generic.jl
+++ b/src/generic.jl
@@ -2108,7 +2108,7 @@ NaN
 function normalize(a::AbstractArray, p::Real = 2)
     nrm = norm(a, p)
     if !isempty(a)
-        aa = copymutable_oftype(a, typeof(first(a)/nrm))
+        aa = copymutable_oftype(a, float(Base.promote_eltype(a, nrm)))
         return __normalize!(aa, nrm)
     else
         T = typeof(zero(eltype(a))/nrm)
--- a/src/generic.jl
+++ b/src/generic.jl
@@ -2108,7 +2108,7 @@ NaN
 function normalize(a::AbstractArray, p::Real = 2)
     nrm = norm(a, p)
     if !isempty(a)
-        aa = copymutable_oftype(a, float(Base.promote_eltype(a, nrm)))
+        aa = copymutable_oftype(a, float(Base.promote_op(/, eltype(a), typeof(nrm))))
         return __normalize!(aa, nrm)
     else
         T = typeof(zero(eltype(a))/nrm)
--- a/src/generic.jl
+++ b/src/generic.jl
@@ -2107,6 +2107,7 @@ NaN
 """
 function normalize(a::AbstractArray, p::Real = 2)
     nrm = norm(a, p)
+    T = promote_op(/, eltype(a), typeof(nrm))
     if !isempty(a)
         aa = copymutable_oftype(a, promote_op(/, eltype(a), typeof(nrm)))
         return __normalize!(aa, nrm)
--- a/src/generic.jl
+++ b/src/generic.jl
@@ -2108,7 +2108,7 @@ NaN
 function normalize(a::AbstractArray, p::Real = 2)
     nrm = norm(a, p)
     if !isempty(a)
-        aa = copymutable_oftype(a, Base.promote_op(/, eltype(a), typeof(nrm)))
+        aa = copymutable_oftype(a, promote_op(/, eltype(a), typeof(nrm)))
         return __normalize!(aa, nrm)
     else
         T = typeof(zero(eltype(a))/nrm)
--- a/src/generic.jl
+++ b/src/generic.jl
@@ -2111,7 +2111,6 @@ function normalize(a::AbstractArray, p::Real = 2)
         aa = copymutable_oftype(a, promote_op(/, eltype(a), typeof(nrm)))
         return __normalize!(aa, nrm)
     else
-        T = typeof(zero(eltype(a))/nrm)
         return T[]
     end
 end
--- a/src/generic.jl
+++ b/src/generic.jl
@@ -2109,7 +2109,7 @@ function normalize(a::AbstractArray, p::Real = 2)
     nrm = norm(a, p)
     T = promote_op(/, eltype(a), typeof(nrm))
     if !isempty(a)
-        aa = copymutable_oftype(a, promote_op(/, eltype(a), typeof(nrm)))
+        aa = copymutable_oftype(a, T)
         return __normalize!(aa, nrm)
     else
         return T[]
--- a/src/generic.jl
+++ b/src/generic.jl
@@ -2108,7 +2108,7 @@ NaN
 function normalize(a::AbstractArray, p::Real = 2)
     nrm = norm(a, p)
     if !isempty(a)
-        aa = copymutable_oftype(a, float(Base.promote_op(/, eltype(a), typeof(nrm))))
+        aa = copymutable_oftype(a, Base.promote_op(/, eltype(a), typeof(nrm)))
         return __normalize!(aa, nrm)
     else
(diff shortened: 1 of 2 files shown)


src/generic.jl
```diff
@@ -2109,7 +2109,7 @@ NaN
 function normalize(a::AbstractArray, p::Real = 2)
     nrm = norm(a, p)
     if !isempty(a)
-        aa = copymutable_oftype(a, typeof(first(a)/nrm))
+        aa = copymutable_oftype(a, float(Base.promote_eltype(a, nrm)))
```
stevengj: This won't work for arrays of arrays:
```jl
julia> a = [[1,2], [3,4]]; nrm = norm(a)
5.477225575051661

julia> Base.promote_eltype(a, nrm)
Any
```

stevengj: Maybe instead of calling `__normalize!`, which requires us to compute the type of the array, we should just call `map`, which does the type computation for us.  Something like:
```jl
function normalize(a::AbstractArray, p::Real = 2)
    nrm = norm(a, p)
    # duplicate logic from __normalize!, but applied to map:
    δ = inv(prevfloat(typemax(nrm)))
    if nrm ≥ δ # Safe to multiply with inverse
        return map(Base.Fix2(*, inv(nrm)), a)
    else # scale elements to avoid overflow
        εδ = eps(one(nrm))/δ
        result = map(Base.Fix2(*, εδ), a)
        return rmul!(result, inv(nrm*εδ))
    end
end
```
This duplicates the logic in `__normalize!`, but that seems easier than duplicating the logic in `map`?

stevengj:
> This is both inefficient and doesn't work on arrays that don't support scalar indexing.

`first(a)` just calls `a[first(eachindex(a))]` — why does this imply scalar indexing?    Any array, even ones that only support cartesian indices, should implement `eachindex`, no?

Do you have an example of an array type where this fails?

albertomercurio:
For example any GPU vector doesn't support it. The rest of the normalize function is supported.

araujoms:
Maybe instead we could use instead `promote_op`; which is not a nice function but used anyway everywhere in `LinearAlgebra`:
```julia
julia> a = [[1,2], [3,4]]; nrm = norm(a)
5.477225575051661

julia> Base.promote_op(/, eltype(a), typeof(nrm))
Vector{Float64} (alias for Array{Float64, 1})
```

albertomercurio:
Ok I have just implemented the `promote_op` version

araujoms:
Why did you put a `float` there?

albertomercurio:
Correct, I have removed it.

araujoms:
Could you please also remove `Base.`, and add @stevengj 's example to the tests?

albertomercurio:
Ok done

araujoms:
On a second thought, we don't want the returned type to be different in both branches, can you define `T = promote_op(/, eltype(a), typeof(nrm))` outside the `if` and use it in both?

Also, the tests are failing. There's no need to make it consistent with `normalize!`, as that is only defined for mutable types.

src/generic.jl
```diff
@@ -2108,7 +2108,7 @@ NaN
 function normalize(a::AbstractArray, p::Real = 2)
     nrm = norm(a, p)
```
araujoms, line 2109:
```suggestion
    nrm = norm(a, p)
    T = promote_op(/, eltype(a), typeof(nrm))
```

src/generic.jl
```diff
@@ -2108,7 +2108,7 @@ NaN
...
     if !isempty(a)
-        aa = copymutable_oftype(a, typeof(first(a)/nrm))
+        aa = copymutable_oftype(a, promote_op(/, eltype(a), typeof(nrm)))
```
araujoms:
```suggestion
        aa = copymutable_oftype(a, T)
```

src/generic.jl
```diff
@@ -2108,7 +2108,7 @@ NaN
...
         return __normalize!(aa, nrm)
     else
         T = typeof(zero(eltype(a))/nrm)
```
araujoms:
```suggestion
```

test/generic.jl
```diff
@@ -495,6 +495,18 @@ end
...
+@testset "normalize array of arrays" begin
+    a = [[1,2], [3,4]]
+    for Tr in (Int32, Int64, Float32, Float64)
+        for T in (Tr, Complex{Tr})
+            b = convert(Vector{Vector{T}}, a)
+            @test normalize(b) == normalize!(copy(b))
+            @test norm(normalize(b)) == 1.0
+            @inferred normalize(b)
+        end
+    end
```
araujoms:
```suggestion
    b = [Int[]]
    @test @inferred norm(normalize(a)) ≈ 1
    @test normalize(a) isa Vector{Vector{Float64}}
    @test normalize(b) isa Vector{Vector{Float64}}
```

araujoms (changes requested):

araujoms (approved):
@stevengj good for merging?

stevengj:
Currently, we have
```
julia> normalize(Any[1])
1-element Vector{Float64}:
 1.0
```
whereas the new version will produce `Vector{Any}`, right?   I guess this is fine?

A little more worrying, right now `normalize(Real[1])` produces `Vector{Float64}` (narrowing the type), but now it will produce `Any[]` (widening the type).

Are there other cases where `promote_op` will produce a different result?

stevengj:
A more robust option, which avoids relying on the (potentially fragile and version-dependent) inference behavior of `promote_op`, would be to duplicate some of the `__normalize!` logic but out-of-place:
```jl
function normalize(a::AbstractArray, p::Real = 2)
    nrm = norm(a, p)
    δ = inv(prevfloat(typemax(nrm)))
    if nrm ≥ δ # Safe to multiply with inverse
        return a * inv(nrm)
    else # scale elements to avoid overflow
        εδ = eps(one(nrm))/δ
        sa = a * εδ
        return ismutable(sa) ? rmul!(sa, inv(nrm*εδ)) : sa * inv(nrm*εδ)
    end
end
```
(There's a reason why we put so many warnings in the `Base.promote_op` docstring, so I'm especially wary of any PR that adds a new dependency on the behavior of `promote_op`.)

(See also #1630 for an improved version of this logic, which should be applied to both `normalize` and `normalize!`.)

test/generic.jl
```diff
@@ -495,6 +495,14 @@ end
                        [0.0 + 0.0im, 0.0 - 0.0im, NaN + NaN*im]))
 end
 
+@testset "normalize array of arrays" begin
```
stevengj, line 498: Should be combined with the `@testset "norm and normalize!" begin` tests.

stevengj (changes requested):
See also my comments regarding avoiding `promote_op`.

test/generic.jl
stevengj: Also, if we adopt the `normalize` algorithm I suggested above, we'll want to add a few more tests, since testing `normalize` will no longer test `normalize!` and vice versa.

araujoms:
Closed in favour of #1634

Issue #1461, opened 2025-10-03
1.12-rc3: Overload Array(::BunchKaufman) to fix Array(factorization(A))


dlfivefifty:
I ran into a missing overload in [MatrixFactorizations.jl](https://github.com/JuliaLinearAlgebra/MatrixFactorizations.jl/pull/72). I believe `Array` should be overloaded for `BunchKaufman` just like it is for `Cholesky`:

```julia
julia> A = Symmetric(randn(5,5)) + 100I;

julia> Array(cholesky(A))
5×5 Matrix{Float64}:
 100.064      -0.513683   0.0526112   -0.734536    0.402664
  -0.513683   99.7659    -0.520904    -0.025515   -0.299517
   0.0526112  -0.520904  98.3513       0.157766    0.239067
  -0.734536   -0.025515   0.157766   100.985      -1.34571
   0.402664   -0.299517   0.239067    -1.34571   100.884

julia> Array(cholesky(A)) ≈ A
true

julia> Array(bunchkaufman(A)) ≈ A
ERROR: MethodError: no method matching Array(::BunchKaufman{Float64, Matrix{Float64}, Vector{Int64}})
The type `Array` exists, but no method is defined for this combination of argument types when trying to construct it.

Closest candidates are:
  Array(::Tridiagonal)
   @ LinearAlgebra ~/Projects/julia-1.12/usr/share/julia/stdlib/v1.12/LinearAlgebra/src/tridiag.jl:624
  Array(::Schur)
   @ LinearAlgebra ~/Projects/julia-1.12/usr/share/julia/stdlib/v1.12/LinearAlgebra/src/schur.jl:446
  Array(::LDLt)
   @ LinearAlgebra ~/Projects/julia-1.12/usr/share/julia/stdlib/v1.12/LinearAlgebra/src/ldlt.jl:224
  ...

Stacktrace:
 [1] top-level scope
   @ REPL[10]:1
```

dlfivefifty:
Actually this didn't exist in Julia v1.11. I'm not sure why the tests for MatrixFactorizations.jl are now failing because of this on v1.12, but in any case it doesn't needed to be fixed urgently.

dlfivefifty:
I realised the issue. In Julia v1.11 we had:
```julia
julia> a = randn(5,5); A = a'a; F = factorize(A);

julia> typeof(F)
Cholesky{Float64, Matrix{Float64}}

julia> Array(F)
[...]

PR #1466, opened 2025-10-08
Overload array constructors for BunchKaufman (#1461)

dlfivefifty:
This adds overloads for array conversion for `BunchKaufman`. This fixes a regression in Julia v1.12-rc3:
```julia
julia> a = randn(5,5); A = a'a; F = factorize(A);

julia> typeof(F) # Was a Cholesky in v1.11
BunchKaufman{Float64, Matrix{Float64}, Vector{Int64}}

julia> Array(F) # Succeeded for Cholesky
ERROR: MethodError: no method matching Array(::BunchKaufman{Float64, Matrix{Float64}, Vector{Int64}})
The type `Array` exists, but no method is defined for this combination of argument types when trying to construct it.

Closest candidates are:
  Array(::QL)
   @ MatrixFactorizations ~/Projects/MatrixFactorizations.jl/src/ql.jl:204
  Array(::Eigen)
   @ LinearAlgebra ~/Projects/julia-1.12/usr/share/julia/stdlib/v1.12/LinearAlgebra/src/eigen.jl:682
  Array(::UniformScaling, ::Tuple{Int64, Int64})
   @ LinearAlgebra ~/Projects/julia-1.12/usr/share/julia/stdlib/v1.12/LinearAlgebra/src/uniformscaling.jl:437
  ...

Stacktrace:
 [1] top-level scope
   @ REPL[52]:1
```
See discussion here: https://github.com/JuliaLang/LinearAlgebra.jl/issues/1461

### DIFF
--- a/src/bunchkaufman.jl
+++ b/src/bunchkaufman.jl
@@ -219,7 +219,7 @@ Factorization{T}(B::BunchKaufman) where {T} = BunchKaufman{T}(B)
 
 AbstractMatrix(B::BunchKaufman) = B.uplo == 'U' ? B.P'B.U*B.D*B.U'B.P : B.P'B.L*B.D*B.L'B.P
 AbstractArray(B::BunchKaufman) = AbstractMatrix(B)
-Matrix(B::BunchKaufman) = Array(AbstractArray(B))
+Matrix(B::BunchKaufman) = convert(Array, AbstractArray(B))
 Array(B::BunchKaufman) = Matrix(B)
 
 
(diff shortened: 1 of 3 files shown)


src/bunchkaufman.jl
```diff
@@ -217,6 +217,12 @@ BunchKaufman{T}(B::BunchKaufman) where {T} =
     BunchKaufman(convert(Matrix{T}, B.LD), B.ipiv, B.uplo, B.symmetric, B.rook, B.info)
 Factorization{T}(B::BunchKaufman) where {T} = BunchKaufman{T}(B)
 
+AbstractMatrix(B::BunchKaufman) = B.uplo == 'U' ? B.P'B.U*B.D*B.U'B.P : B.P'B.L*B.D*B.L'B.P
+AbstractArray(B::BunchKaufman) = AbstractMatrix(B)
+Matrix(B::BunchKaufman) = Array(AbstractArray(B))
```
dkarrasch: Perhaps

```suggestion
Matrix(B::BunchKaufman) = convert(Matrix, AbstractArray(B))
```

to avoid another copy, when the inner call already returns a `Matrix`?

dlfivefifty: I made the change. Also in cholesky.jl (where I copied the code from)

dkarrasch (approved):

PR #1476, opened 2025-10-17
make checksquare fail on non-matrices

### LINKED ISSUE #1362: checksquare should error for generic arrays
Currently

1. it works for more than two dimensions,
2. regardless of what the trailing sizes are

```julia
julia> LinearAlgebra.checksquare(zeros(1,1,1))
1

julia> LinearAlgebra.checksquare(zeros(1,1,2))
1
```

Incidentally, the docstring should document that the error thrown is a `DimensionError`.

araujoms:
Closes #1362

### DIFF
--- a/src/LinearAlgebra.jl
+++ b/src/LinearAlgebra.jl
@@ -346,13 +346,7 @@ function checksquare(A)
     return sizeA[1]
 end
 
-function checksquare(A...)
-    sizes = Vector{Int}(undef, length(A))
-    for i in eachindex(A)
-        sizes[i] = checksquare(A[i])
-    end
-    return sizes
-end
+checksquare(A...) = [checksquare(a) for a in A]
 
 function char_uplo(uplo::Symbol)
     if uplo === :U
--- a/src/LinearAlgebra.jl
+++ b/src/LinearAlgebra.jl
@@ -326,7 +326,7 @@ StridedMatrixStride1{T} = StridedArrayStride1{T,2}
 """
     LinearAlgebra.checksquare(A)
 
-Check that a matrix is square, then return its common dimension.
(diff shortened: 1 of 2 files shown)


stevengj (approved):

src/LinearAlgebra.jl
```diff
@@ -340,16 +340,16 @@ julia> LinearAlgebra.checksquare(A, B)
...
-    m,n = size(A)
-    m == n || throw(DimensionMismatch(lazy"matrix is not square: dimensions are $(size(A))"))
-    m
+    sizeA = size(A)
+    length(sizeA) == 2 || throw(DimensionMismatch(lazy"input is not a matrix: dimensions are $sizeA"))
+    sizeA[1] == sizeA[2] || throw(DimensionMismatch(lazy"matrix is not square: dimensions are $sizeA"))
+    return sizeA[1]
 end
 
 function checksquare(A...)
```
stevengj: Side note: it seems pretty weird and sub-optimal that this function returns a vector and not a tuple.

i.e. it seems like it should really be:
```jl
checksquare(A...) = map(checksquare, A) # returns a tuple
```
but since it is public and documented maybe it can't be fixed?

That being said, a simpler implementation returning a vector would be
```jl
checksquare(A...) = Int[checksquare(a) for a in A]
```

araujoms: I agree. Personally I'm rather blasé about changing the return type of functions, but I'm afraid @jishnub would object. In any case, thanks for the one liner, I don't see how could I have missed it.

dkarrasch: The tuple-vs-vector issue related to `checksquare` has been discussed many times, and it was common sense that we can't change it. We might introduce an internal function, though, that returns tuples and use it inside LinearAlgebra.jl.

jishnub (approved):

Issue #1450, opened 2022-01-27
Inconsistency in `eigen` interface for specialised methods


st--:
Some of the implementations such as `eigen(::SymTridiagonal)` don't actually accept the keyword arguments of the other implementations, `scale`, `permute`, `sortby`. (`eigen(::Diagonal)` does accept even permute and scale though it just ignores them.)

Should these also wrap the Lapack result in `sorteig!` before passing it to the `Eigen` constructor? And have dummy kwargs for scale and permute?

_Originally posted by @st-- in https://github.com/JuliaLang/julia/issues/43904#issuecomment-1021963768_

stevengj:
This is documented in the `eigen` docstring:

> Some special matrix types (e.g. `Diagonal` or `SymTridiagonal`) may implement their own sorting convention and not accept a `sortby` keyword.

and 

> For **general nonsymmetric matrices** [*emphasis added*] it is possible to specify how the matrix is balanced before the eigenvector calculation. […documentation for `permute` and `scale`…]

For example, re-ordering the eigenvalues of a `Diagonal` matrix is a pretty awkward thing to do, because it makes the eigenvectors no longer the identity matrix.  See also the discussion in https://github.com/JuliaLang/julia/pull/21598

araujoms:
Fixed by #1477, #1595, #1626, and #1645

PR #1477, opened 2025-10-17
make eigvals(::Diagonal) accept sortby

araujoms:
Solves the `Diagonal` part of #1471 and #1450.

I thought it would be better to start with `Diagonal`, as it was the only part that required a bit of thinking. I think it also makes sense in itself, since `eigen` accepted `sortby`, just `eigvals` and `eigvecs` did not. Even though the docstring said none did.

If this is welcome I'll also do `SymTridiagonal`, and then we can document that eigenvalues are sorted by default.

### DIFF
--- a/src/diagonal.jl
+++ b/src/diagonal.jl
@@ -1007,16 +1007,16 @@ end
 _ortho_eltype(T) = Base.promote_op(/, T, T)
 _ortho_eltype(T::Type{<:Number}) = typeof(one(T)/one(T))
 
-# TODO Docstrings for eigvals, eigvecs, eigen all mention permute, scale, sortby as keyword args
-# but not all of them below provide them. Do we need to fix that?
 #Eigensystem
-eigvals(D::Diagonal{<:Number}; permute::Bool=true, scale::Bool=true) = copy(D.diag)
-eigvals(D::Diagonal; permute::Bool=true, scale::Bool=true) =
-    reduce(vcat, eigvals(x) for x in D.diag) #For block matrices, etc.
-function eigvecs(D::Diagonal{T}) where {T<:AbstractMatrix}
-    diag_vecs = [ eigvecs(x) for x in D.diag ]
+eigvals(D::Diagonal{<:Number}; permute::Bool=true, scale::Bool=true, sortby::Union{Function,Nothing}=eigsortby) = sorteig!(copy(D.diag), sortby)
+eigvals(D::Diagonal; permute::Bool=true, scale::Bool=true, sortby::Union{Function,Nothing}=eigsortby) =
+sorteig!(reduce(vcat, eigvals(x; sortby=nothing) for x in D.diag), sortby) #For block matrices, etc.
+function _eigen(D::Diagonal{T}) where {T<:AbstractMatrix}
+    facts = [eigen(x; sortby=nothing) for x in D.diag]
+    λ = reduce(vcat, f.values for f in facts)
+    diag_vecs = [f.vectors for f in facts]
     matT = promote_type(map(typeof, diag_vecs)...)
-    ncols_diag = [ size(x, 2) for x in D.diag ]
+    ncols_diag = [size(x, 2) for x in D.diag]
     nrows = size(D, 1)
     vecs = Matrix{Vector{eltype(matT)}}(undef, nrows, sum(ncols_diag))
     for j in axes(D, 2), i in axes(D, 1)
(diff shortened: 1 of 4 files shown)


ViralBShah:
cc @stevengj

src/diagonal.jl
```diff
@@ -1007,16 +1007,16 @@ end
...
-# but not all of them below provide them. Do we need to fix that?
 #Eigensystem
-eigvals(D::Diagonal{<:Number}; permute::Bool=true, scale::Bool=true) = copy(D.diag)
-eigvals(D::Diagonal; permute::Bool=true, scale::Bool=true) =
-    reduce(vcat, eigvals(x) for x in D.diag) #For block matrices, etc.
-function eigvecs(D::Diagonal{T}) where {T<:AbstractMatrix}
-    diag_vecs = [ eigvecs(x) for x in D.diag ]
+eigvals(D::Diagonal{<:Number}; permute::Bool=true, scale::Bool=true, sortby::Union{Function,Nothing}=eigsortby) = sorteig!(copy(D.diag), sortby)
+eigvals(D::Diagonal; permute::Bool=true, scale::Bool=true, sortby::Union{Function,Nothing}=eigsortby) =
+sorteig!(reduce(vcat, eigvals(x; sortby=nothing) for x in D.diag), sortby) #For block matrices, etc.
```
```suggestion
    sorteig!(reduce(vcat, eigvals(x; sortby=nothing) for x in D.diag), sortby) #For block matrices, etc.
```

dkarrasch (commented):
Just a minor comment. I like it in terms of consistency, but I'm a little afraid due to changing the default behavior.

araujoms:
I think one can hardly rely on eigenvalues *not* being sorted, making them sorted by default is harmless.

Another possibility would be to add the `sortby` keyword argument to `eigvals(::Diagonal)` with default value `nothing`. It would make the interface inconsistent, but at least it would be possible to pass `eigsortby` and be certain that the eigenvalues will be sorted.

dkarrasch:
Do we have a test for the eigenvectors with sorted eigenvalues? Currently, the default `eigvecs` is a `Diagonal`, so we can't sort inplace, and need to make sure this works correctly.

About the default sorting, @stevengj's comment is relevant here: https://github.com/JuliaLang/LinearAlgebra.jl/issues/1450#issuecomment-3307013409 Before v.12, the eigenvectors were returned as a dense matrix, for which it doesn't seem to make any difference whether the eigenvectors are the identity matrix or a permuted version thereof. With the new `Diagonal` return type, it starts to make a difference, so, as you say, we may consider making the default for `Diagonal`s no sorting, even though that would be inconsistent with the symmetric case. On the other hand, it would save memory and computational effort by default.

In any case, making sorting accessible is a clear (consistency) win!

araujoms:
There is already a test, line 422 of `test/diagonal.jl`.

Note that I didn't change the behaviour of `eigen(::Diagonal)`, just the default, so it will still return a `Diagonal` matrix of eigenvectors. I don't think that's a good idea, I think it should always return a dense matrix even when the matrix of eigenvectors is identity, for two reasons: 1 - it made the function type unstable, and 2 - the optimization is not really useful, as nobody calls `eigen` specifically on a diagonal matrix, this is done only in generic code. and for generic code it's better to avoid special cases.

dkarrasch:
Now I see clearer. If you change the default, it won't return a `Diagonal` anymore, since that is only if the sortby function is `nothing`. In all other cases, it will allocate a matrix and potentially permute the ones. Type instability is not an issue as long as the sortby function is a constant, like hard-coded or not given at all and then filled by the default value. Given the older discussions, I'd personally prefer to keep this "less controversial" and keep the default, at least for now, but make `eigvals` accept the sorting function as the title says.

araujoms:
Ah, I see, functions are types, so it is not actually type unstable. Just a trip hazard.

Is it less controversial, though? People already complained about it in the original PR https://github.com/JuliaLang/julia/pull/21598, and another person besides me filed an issue about it.

dkarrasch:
I guess differing opinions is what makes it controversial. 😂 I don't have a personal preference TBH.

If we change the default behavior, this needs to be listed in some sort of NEWS.md, together with the advice that if the user wants to benefit from low-memory `eigvecs`, the `sortby=nothing` keyword argument needs to be provided explicitly.

araujoms:
Sorry, I thought you meant keeping the default `sortby=nothing` is less controversial than changing it to `sortby=eigsortby`. Now I see you meant that that adding `sortby` to `eigvals(::Diagonal)` is less controversial than changing the default.

Fair enough, I want to do another PR for `SymTridiagonal` anyway, I can reverse the change here and leave it for there.

dkarrasch:
Yes, adding the keyword is nice, and an "uncontroversial" new feature if we, for the time being, use/leave the current default sorting function `nothing`. Adding the keyword also to `SymTridiagonal` is also nice.

Orthogonally to that, perhaps in an issue, we should discuss if we shift towards having all `eig*`, which accept a sorting function, towards using `eigsortby`. But I guess we should try to reach more people and get them involved.

araujoms:
I did open an issue to discuss that before writing the PR, that was #1471, which was closed as a duplicate of #1450, which got crickets.

dkarrasch (approved):

stevengj:
> I think one can hardly rely on eigenvalues not being sorted, making them sorted by default is harmless.

Except that for a `Diagonal` matrix there *is* a "natural" ordering of eigenvalues given by the order in which the diagonal elements appear along the diagonal, corresponding to eigenvectors $= I$.

araujoms:
Sure, viewed in isolation that is the natural answer. But what is the most useful way to program `eigen`/`eigvals`? Nobody calls them specifically for `Diagonal` matrices, so I don't think giving the natural answer is useful. I think the most useful is to have a consistent interface for generic matrices.

dkarrasch:
I guess it really depends on what people do after the eigendecomposition, and if there are any implicit assumptions when users don't specify an eigenvalue sorting. Like, is it error-prone or something?

PR #1480, opened 2025-10-24
Fix forgotten conj(alpha) in her2k documentation

Technici4n:
See BLAS doc: https://www.netlib.org/lapack/explore-html/d8/d94/group__her2k_ga9e41d3bfaa0d0eeb02e46477d23f679c.html


dkarrasch:
Nice catch! Thank you.

PR #1489, opened 2025-11-24
typo in Elementary Operations table for UpperHessenberg matrix type 

langou:
There is a typo at
[https://docs.julialang.org/en/v1/stdlib/LinearAlgebra/#Elementary-operations](https://docs.julialang.org/en/v1/stdlib/LinearAlgebra/#Elementary-operations)
for the row with UpperHessenberg, the supported operations should be MV and not MM.

This PR fixes the typo.


docs/src/index.md
```diff
@@ -194,7 +194,7 @@ as well as whether hooks to various optimized methods for them in LAPACK are ava
 | [`UnitUpperTriangular`](@ref) |     |     | MV  | MV  | [`inv`](@ref), [`det`](@ref), [`logdet`](@ref)                                |
 | [`LowerTriangular`](@ref)     |     |     | MV  | MV  | [`inv`](@ref), [`det`](@ref), [`logdet`](@ref)                                |
 | [`UnitLowerTriangular`](@ref) |     |     | MV  | MV  | [`inv`](@ref), [`det`](@ref), [`logdet`](@ref)                                |
-| [`UpperHessenberg`](@ref)     |     |     |     | MM  | [`inv`](@ref), [`det`](@ref)                                |
+| [`UpperHessenberg`](@ref)     |     |     |     | MV  | [`inv`](@ref), [`det`](@ref)                                |
```
stevengj:
```suggestion
| [`UpperHessenberg`](@ref)     |     |     |     |    | [`inv`](@ref), [`det`](@ref)                                |
```
Shouldn't it simply be blank?  I don't see any optimized MM or MV methods for `UpperHessenberg`.

In principle, an optimized MV probably should be added — it could call `BLAS.trmv!` to multiply by the upper-triangular part, plus a linear-time loop to add in the contribution of the lower diagonal.

Similarly, we could implement an optimized MM by calling `BLAS.trmm!` for the $O(m^3)$ triangular MM and then implement an $O(m^2)$ loop for the remaining contributions.  (Though I'm not sure how much demand there is for upper-Hessenberg MM.)

langou:
Hi @stevengj, 

(*) OKey doc, I had no idea whether there was an optimized implementation for mat-vec or mat-mat with Hessenberg matrices but the 'MM' was not making sense and was confusing to me. Thanks!

(*) I think it is a good idea to have a 'UpperHessenberg' type even though there is no optimized operation for it. It is still convenient and can be handy. Great.

(*) I think it is a good idea to keep the line 'UpperHessenberg' in "Elementary operations", even though the line is empty, maybe one day, there will be some optimized versions there.

(*) Quoting "In principle, an optimized MV probably should be added — it could call `BLAS.trmv!` to multiply by the upper-triangular part, plus a linear-time loop to add in the contribution of the lower diagonal. Similarly, we could implement an optimized MM by calling `BLAS.trmm!` for the $\mathcal{O}( m^3 )$ triangular MM and then implement an $\mathcal{O} ( m^2 )$ loop for the remaining contributions. (Though I'm not sure how much demand there is for upper-Hessenberg MM.)" I agree with all of this. I also note that for the matrix-vector product one could use BLAS.gbmv( ) to handle this. But the approach proposed by @stevengj sounds great.

(*) Related, I am not sure what is being done for the "norms" of these special matrices. For example, LAPACK has LAPACK.lantr() for triangular matrices and then LAPACK.lansy() for symmetric matrices, etc. If I call 'opnorm( Symmetric(A), 1)', does it call LAPACK.lansy()? If I call 'opnorm( UnitUpperTriangular(A), 1)', does it call LAPACK.lantr()? This could be considered as "optimized" elementary operation. It could be good to have a column "norm". I can open an issue and a PR with this but the column will most likely be empty! I am not sure I'll be able to implement the wrapper for LAPACK norm functions.

Cheers,
Julien.

stevengj:
> I think it is a good idea to have a 'UpperHessenberg' type even though there is no optimized operation for it. It is still convenient and can be handy. Great.

The optimized operation for it is `ldiv`.

stevengj:
Whoops, I was looking at the wrong column.  This *was* the column for `\`, and there *is* an optimized MV routine for this (it's not LAPACK, but rather a specialized function in Julia).  Sorry, I thought this was the column for `*`.

So "MV" is indeed appropriate.

(It also supports MM `H \ B`, but that is not at BLAS3-style optimization that exploits the matrix–matrix structure.)

docs/src/index.md
```diff
@@ -194,7 +194,7 @@ as well as whether hooks to various optimized methods for them in LAPACK are ava
 | [`UnitUpperTriangular`](@ref) |     |     | MV  | MV  | [`inv`](@ref), [`det`](@ref), [`logdet`](@ref)                                |
 | [`LowerTriangular`](@ref)     |     |     | MV  | MV  | [`inv`](@ref), [`det`](@ref), [`logdet`](@ref)                                |
 | [`UnitLowerTriangular`](@ref) |     |     | MV  | MV  | [`inv`](@ref), [`det`](@ref), [`logdet`](@ref)                                |
-| [`UpperHessenberg`](@ref)     |     |     |     | MM  | [`inv`](@ref), [`det`](@ref)                                |
+| [`UpperHessenberg`](@ref)     |     |     |     |    | [`inv`](@ref), [`det`](@ref)                                |
```
stevengj: Sorry, maybe this should be MV after all, since this is in the `\` column as noted above.  Also there is an optimized `logdet` method.

stevengj:
On the other hand, the table says that it is about "optimized methods for them in LAPACK", whereas the optimized method here is in Julia.    By that criterion, `UpperHessenberg` should currently list no LAPACK methods.

(We might want to re-think the description of the table and what this is for.)

langou:
> On the other hand, the table says that it is about "optimized methods for them in LAPACK", whereas the optimized method here is in Julia. By that criterion, UpperHessenberg should currently list no LAPACK methods.

(*) Maybe use "optimized" for LAPACK and then "specialized" for julia implementations. Or maybe call LAPACK something like "LAPACK optimized backend" and "specialized" sounds good. In that sense, 'logdet' would be "specialized" and not "optimized" indeed. It might be worth differentiating indeed. 

Maybe the table is about "optimized" and "specialized" functionality. Or maybe two tables.

> "So "MV" is indeed appropriate. (It also supports MM H \ B, but that is not at BLAS3-style optimization that exploits the matrix–matrix structure.)"

Comment: My understanding is that 'MM' does not makes sense in this table. Maybe the notation should be changed. If we want to say 'MM' (matrix-matrix), we need to say 'M'. If we want to say 'MV' (matrix-vector), we need to say 'V'. If we want to say 'MM' (matrix-vector) and 'MV' (matrix-vector), we need to say 'MV'. This is written in the "legend" just below the table "Elementary operations". I find this confusing-ish, and maybe using something like MM-MV-MS (instead of MVS) while more verbose would be clearer.

langou:
We are back with 'MV' in the '\' column. 🔁. Let me know. Happy to remove it too if that's the decision. In any case, 'MM' needs to be changed. (To 'MV' or to a blank.)

stevengj:
I mean, there is a specialized `ldiv!` method for `ldiv!(UpperHessenberg, Matrix)` that should be somewat faster than acting column-by-column (though only about 40% in a quick test on an 1000x1000 example).

~~That being said, I noticed another problem.  `UpperHessenberg \ MatrixOrVector` currently calls the generic fallback, doing an LU factorization of the Hessenberg matrix, instead of calling the specialized `ldiv!` method.  This needs to be fixed: https://github.com/JuliaLang/LinearAlgebra.jl/pull/1497~~ already fixed in https://github.com/JuliaLang/LinearAlgebra.jl/pull/1322

langou:
> I mean, there is a specialized ldiv! method for ldiv!(UpperHessenberg, Matrix)

Yes, I hear you. So do we want "MV" or nothing there? Is "specialized" part of this table or not?

stevengj:
I would say not...just leave it as MV.

The MM method it is specialized, but only slightly faster, whereas a BLAS3-style implementation is *much* faster.  (In fact, it's so much faster that it seems faster to just do `lu` factorization of the Hessenberg matrix and then call the triangular solver in BLAS when the right-hand-side is a matrix.   We could improve on this, I guess, by implementing a specialized qr method to get a triangular matrix in O(n^2) time, and using that when the right-hand-side is a matrix.

langou:
> I would say not...just leave it as MV.

OK. Cool. I agree. Leaving it as MV. Then I think the PR is ready for merge.

langou:
> The MM method it is specialized, but only slightly faster, whereas a BLAS3-style implementation is much faster. (In fact, it's so much faster that it seems faster to just do lu factorization of the Hessenberg matrix and then call the triangular solver in BLAS when the right-hand-side is a matrix. We could improve on this, I guess, by implementing a specialized qr method to get a triangular matrix in O(n^2) time, and using that when the right-hand-side is a matrix.

I am reading the code at 
https://github.com/JuliaLang/LinearAlgebra.jl/blob/189d49d728913b26d1ef58a373464e2d56839bc4/src/hessenberg.jl#L224

Two comments:

(*) the main difficulty is to do this with H as input only, which is nice, but indeed, that's a strong constrain. Transforming H to upper triangular would be fast, and then TRSM and done. But that would have H as inout. (H would be trashed.)

(*) with respect to current code, it seems that if we allow one array of size m, we can vectorize all the i-loops. That probably would help a tat with the performance.

PR #1485, opened 2025-11-17
Fix 3-arg `dot` for empty arrays

dkarrasch:
Fixes #1484. Should be backported.

EDIT: This fixes it for all matrix types (`HermOrSym` were also missing), and also makes sure that the exact zero is returned, to avoid things like

```julia
julia> dot(zero(ComplexF64), zero(Float64), zero(Float64))
0.0 - 0.0im
```

PR #1239, opened 2025-03-14
Revert "Use `BLAS.trsm!` instead of `LAPACK.trtrs!` in left-triangular solves"

vtjnash:
Reverts JuliaLang/LinearAlgebra.jl#1194 since it introduced a bug into BaseBenchmarks

PR #1490, opened 2025-11-26
Backports to v1.12

dkarrasch:
- [x] #1239
- [x] #1485
- [x] #1480
- [x] #1492 
- [x] #1494


dkarrasch:
Why can't I make a merge commit? It only offers "Squash and merge", but we don't squash backports commits, do we?

@KristofferC @ViralBShah

ViralBShah:
I personally have no view on it - if useful, we can turn them on. I am not sure how they interact with the rest of our processes.

PR #1497, opened 2025-11-28
Hessenberg \ and / should call ldiv! and rdiv!

stevengj:
Fixes a problem noted in https://github.com/JuliaLang/LinearAlgebra.jl/pull/1489#issuecomment-3590179742

> I noticed another problem.  `UpperHessenberg \ MatrixOrVector` currently calls the generic fallback, doing an LU factorization of the Hessenberg matrix, instead of calling the specialized `ldiv!` method.


stevengj:
Oh nevermind, it was already fixed by me a few months ago in https://github.com/JuliaLang/LinearAlgebra.jl/pull/1322, duh.  I just wasn't looking at the master branch; I knew this issue seemed familiar.

PR #1500, opened 2025-11-30
Reduce method definitions for cat of `AbstractQ`s

dkarrasch:
This is related to #1233. The remaining function definitions are necessary because the generic counterparts treat everything not-an-array as a scalar, so we absolutely have to teach it to treat `AbstractQ`s as arrays in concatenation. In the assignment of a subarray this even allows to avoid a temporary array.

### DIFF
--- a/src/abstractq.jl
+++ b/src/abstractq.jl
@@ -145,7 +145,6 @@ function copyto!(dest::PermutedDimsArray{T,2,perm}, src::AbstractQ) where {T,per
     end
     return dest
 end
-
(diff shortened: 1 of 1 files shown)


src/abstractq.jl
```diff
@@ -145,16 +145,15 @@ function copyto!(dest::PermutedDimsArray{T,2,perm}, src::AbstractQ) where {T,per
     end
     return dest
 end
+
```
dkarrasch:
```suggestion
```

Issue #1139, opened 2024-12-04
Add `NEWS.md` to keep track of/announce changes


ajinkya-k:
The [JuliaLang/julia](https://github.com/JuliaLang/julia) repo has a `NEWS.md` which also documented changes in `LinearAlgebra`. We might need one for this repo

inkydragon:
Or a `CHANGELOG.md`, like `Pkg.jl`: https://github.com/JuliaLang/Pkg.jl/blob/master/CHANGELOG.md

ajinkya-k:
@inkydragon That makes sense

ajinkya-k:
This one is kinda important but still not done. Can some of the maintainers take a look?

ViralBShah:
We'll need a volunteer for this if we want something better than just the commit log.

PR #1532, opened 2025-12-27
Added CHANGELOG.md for LinearAlgebra.jl

Sisira07:
This PR adds a CHANGELOG.md for LinearAlgebra.jl, covering versions 1.6 through 1.12. The content is with reference to [HISTORY.md](https://github.com/JuliaLang/julia/blob/master/HISTORY.md).
issue #1139


stevengj:
Should it be HISTORY.md to match Julia?

ViralBShah:
Yes that’s a good idea.

Sisira07:
If we consider this to be a HISTORY.md file, would you prefer that I also include the changes from versions before 1.6?

ViralBShah:
Going from 1.0 onwards would be great of it is easy to do. Otherwise 1.6 onwards is plenty far back.

jishnub:
Let's merge for now, and progressively update this if necessary?

PR #1533, opened 2026-01-01
Allow MulAddMul to be fused

yuyichao:
Is there any reason `muladd` wasn't used here previously? AFAICT normal blas implementation makes very liberal use of `fma` instructions when available so I don't see why not here.

There are probably other places where `muladd` can be used but this is probably the place that has the broadest impact.


ViralBShah:
cc @oscardssmith

oscardssmith:
seems reasonable as long as this isn't circular

KristofferC (approved):

PR #1534, opened 2026-01-01
Only loop over nonzero indices in matvecmul

jishnub:
This speeds up the operation for sparse arrays in general if a specialization is added by `SparseArrays.jl`. I'm choosing `FillArrays` somewhat arbitrarily in the example to ensure that we take the fallback method defined here.

```julia
julia> using LinearAlgebra, FillArrays, SparseArrays

julia> A = Fill(4, 100, 100);

julia> v = sparsevec([1, 100], [1, 100]);

julia> y = similar(A, size(A,1));

julia> @btime mul!($y, $A, $v);
  1.159 μs (0 allocations: 0 bytes)

julia> LinearAlgebra.nonzeroinds(v::SparseVector) = SparseArrays.nonzeroinds(v)

julia> @btime mul!($y, $A, $v);
  243.815 ns (0 allocations: 0 bytes)
```

The idea is that since we compute `A * v` as `sum(@view(A[:,i]) * v[i] for i in eachindex(v))`, we may limit the sum to the indices for which `v` has a non-zero element.

### DIFF
--- a/src/LinearAlgebra.jl
+++ b/src/LinearAlgebra.jl
@@ -551,6 +551,7 @@ include("exceptions.jl")
 include("generic.jl")
 
 include("blas.jl")
+include("sparse_interface.jl")
 include("matmul.jl")
 include("lapack.jl")
 
--- a/src/matmul.jl
+++ b/src/matmul.jl
@@ -13,6 +13,8 @@ StridedMaybeAdjOrTransVecOrMat{T} = Union{StridedVecOrMat{T}, AdjOrTrans{<:Any, 
 
 matprod(x, y) = x*y + x*y
 
+nonzeroinds(v::AbstractVector) = eachindex(v)
+
 # dot products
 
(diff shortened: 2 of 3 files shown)


dkarrasch:
Should we move ownership to `LinearAlgebra` and only extend the method in `SparseArrays`? Or is that too "breaking", like if other packages extend the `SparseArrays` method?

jishnub:
[Quite a few packages](https://juliahub.com/ui/Search?type=code&q=nonzeroinds) seem to extend the `SparseArrays` function and explicitly import the name from `SparseArrays`. These should largely work if the ownership is changed, but code quality tests such as `ExplicitImports` might induce breakages. Perhaps we may start here and slowly migrate the packages if this makes sense.

ViralBShah:
That would certainly make sense. Should be easy to create a pkgeval run in the julia repo.

PR #1537, opened 2026-01-10
Remove some re-exports from Base

ericphanson:
Base exports `copyto!`, `/`, `\`, and `convert`. LinearAlgebra currently re-exports all of these. I don't think it should really, it's non-standard to re-export what Base is already exporting. This causes slightly weird behavior in ExplicitImports: https://github.com/JuliaTesting/ExplicitImports.jl/issues/137.

xref https://github.com/JuliaLang/julia/issues/60318

### DIFF
--- a/src/LinearAlgebra.jl
+++ b/src/LinearAlgebra.jl
@@ -81,7 +81,6 @@ export
     condskeel,
     copy_adjoint!,
     copy_transpose!,
-    copyto!,
     copytrito!,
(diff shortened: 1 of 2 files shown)


dkarrasch (approved):

ViralBShah:
Actually, just after I hit merge, I realized that the better thing to do here would be to comment these out with a note, instead of deleting them.

PR #1539, opened 2026-01-11
Function to obtain non-zero rows/cols

jishnub:
This incorporates more of the sparse matrix querying functions into this package. With this,
```julia
julia> using LinearAlgebra, SparseArrays

julia> S = sprand(10, 10, 0.1);

julia> A = rand(10,10);

julia> using Test: GenericArray # to hit the fallback

julia> B = GenericArray(A);

julia> C = zero(A);

julia> @btime mul!($C, $S, $B);
  2.314 μs (0 allocations: 0 bytes)

julia> LinearAlgebra.nzrows(S::SparseMatrixCSC, k::Integer) = @view rowvals(S)[nzrange(S,k)]

julia> @btime mul!($C, $S, $B);
 1.069 μs (0 allocations: 0 bytes)
```


dkarrasch:
Nice!

PR #1540, opened 2026-01-12
Comment-out re-exports from Base

ericphanson:
This re-does #1537 addressing @ViralBShah's review comment that they should be commented out instead. I was also a bit more thoroughly, using this list:
```julia
julia> names(LinearAlgebra) ∩ names(Base)
8-element Vector{Symbol}:
 :/
 :\
 :adjoint
 :convert
 :copyto!
 :kron
 :kron!
 :transpose
```

(on 3-day old nightly).

### DIFF
--- a/src/LinearAlgebra.jl
+++ b/src/LinearAlgebra.jl
@@ -81,6 +81,7 @@ export
     condskeel,
     copy_adjoint!,
     copy_transpose!,
+    copyto!,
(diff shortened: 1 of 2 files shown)


dkarrasch (approved):
Thanks for (carefully) double-checking!

ViralBShah (approved):

PR #1541, opened 2026-01-13
Optimize `copytrito!(B, A, 'L')`

putianyi889:
Before this PR
```julia
else # uplo == 'L'
    LAPACK.lacpy_size_check(size(B), (m, m < n ? m : n))
    for j in axes(A,2), i in axes(A,1)[j:end]
        # extract the parents for LowerTriangular matrices
        Bv, Av = lowertridata(B), lowertridata(A)
        @inbounds Bv[i,j] = Av[i,j]
    end
end
```
The loop bound is not optimal for rectangular matrices. Example:
```julia
julia> A = zeros(3,10000);

julia> @btime copytrito!(B,A,'L');
  4.714 μs (0 allocations: 0 bytes)

julia> A = zeros(3,100000);

julia> @btime copytrito!(B,A,'L');
  46.600 μs (0 allocations: 0 bytes)
```

This PR changes the loop bound to be the diagonal length.

### DIFF
--- a/src/generic.jl
+++ b/src/generic.jl
@@ -2147,19 +2147,20 @@ function copytrito!(B::AbstractMatrix, A::AbstractMatrix, uplo::AbstractChar)
     BLAS.chkuplo(uplo)
     B === A && return B
     m,n = size(A)
+    d = min(m,n)
     A = Base.unalias(B, A)
     if uplo == 'U'
-        LAPACK.lacpy_size_check(size(B), (n < m ? n : m, n))
+        LAPACK.lacpy_size_check(size(B), (d, n))
         # extract the parents for UpperTriangular matrices
         Bv, Av = uppertridata(B), uppertridata(A)
         for j in axes(A,2), i in axes(A,1)[begin : min(j,end)]
             @inbounds Bv[i,j] = Av[i,j]
         end
     else # uplo == 'L'
-        LAPACK.lacpy_size_check(size(B), (m, m < n ? m : n))
+        LAPACK.lacpy_size_check(size(B), (m, d))
         # extract the parents for LowerTriangular matrices
         Bv, Av = lowertridata(B), lowertridata(A)
-        for j in axes(A,2), i in axes(A,1)[j:end]
+        for j in axes(A,2)[1:d], i in axes(A,1)[j:end]
             @inbounds Bv[i,j] = Av[i,j]
         end
     end
--- a/src/lapack.jl
+++ b/src/lapack.jl
@@ -7179,19 +7179,27 @@ for (fn, elty) in ((:dlacpy_, :Float64),
             chkstride1(A, B)
             m, n = size(A)
             m1, n1 = size(B)
+            minmn = min(m, n)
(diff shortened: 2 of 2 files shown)


dkarrasch:
What are the improved runtimes?

putianyi889:
> What are the improved runtimes?

Good question. I gave a bad example because `Matrix` is redirected to a `LAPACK` routine, which indicates a deeper problem.

Anyway, if the general implementation is forced to apply to `Matrix`,
```julia
function LinearAlgebra.copytrito!(B::Matrix, A::Matrix, uplo::AbstractChar)
    Base.require_one_based_indexing(A, B)
    LinearAlgebra.BLAS.chkuplo(uplo)
    B === A && return B
    m,n = size(A)
    d = min(m,n)
    A = Base.unalias(B, A)
    if uplo == 'U'
        LinearAlgebra.LAPACK.lacpy_size_check(size(B), (d, n))
        # extract the parents for UpperTriangular matrices
        Bv, Av = LinearAlgebra.uppertridata(B), LinearAlgebra.uppertridata(A)
        for j in axes(A,2), i in axes(A,1)[begin : min(j,end)]
            @inbounds Bv[i,j] = Av[i,j]
        end
    else # uplo == 'L'
        LinearAlgebra.LAPACK.lacpy_size_check(size(B), (m, d))
        # extract the parents for LowerTriangular matrices
        Bv, Av = LinearAlgebra.lowertridata(B), LinearAlgebra.lowertridata(A)
        for j in axes(A,2)[1:d], i in axes(A,1)[j:end]
            @inbounds Bv[i,j] = Av[i,j]
        end
    end
    return B
end
```

```julia
julia> A = zeros(3,10000);

julia> @btime copytrito!(B,A,'L');
  43.939 ns (0 allocations: 0 bytes)

julia> A = zeros(3,100000);

julia> @btime copytrito!(B,A,'L');
  49.848 ns (0 allocations: 0 bytes)

julia> A = zeros(10000,3); # heavy lower-triangular part for testing

julia> @btime copytrito!(B,A,'L');
  4.171 μs (0 allocations: 0 bytes)

julia> A = zeros(100000,3);

julia> @btime copytrito!(B,A,'L');
  132.400 μs (0 allocations: 0 bytes)
```

putianyi889:
The third comment of https://github.com/Reference-LAPACK/lapack/issues/1183#issuecomment-3748656316 gives a workaround. Shall we use it before LAPACK is updated?

dkarrasch:
> The third comment of [Reference-LAPACK/lapack#1183 (comment)](https://github.com/Reference-LAPACK/lapack/issues/1183#issuecomment-3748656316) gives a workaround. Shall we use it before LAPACK is updated?

Yes.

dkarrasch:
Seems ready, doesn't it? I checked that there is no overhead in the wide rectangular case for the LAPACK case.

putianyi889:
> Seems ready, doesn't it? I checked that there is no overhead in the wide rectangular case for the LAPACK case.

Agree.

PR #1544, opened 2026-01-30
Efficient powermod version for matrices

Shmuma:
Function implements efficient `powermod` operation for integer matrices.

This is my first julia contribution, so I don't have lots of experience :).

### DIFF
--- a/src/matmul.jl
+++ b/src/matmul.jl
@@ -1525,7 +1525,7 @@ julia> powermod([big(1) big(1); big(1) big(0)], big(12341234), 1_000_000_007)
 ```
 """
 function Base.powermod(x::Matrix{T}, p::M, m::K)::Matrix{T} where {T<:Integer, M<:Integer, K<:Integer}
-    p <= 0 && return I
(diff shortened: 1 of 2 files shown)


stevengj:
`mod(::Matrix{Int64}, ::Int64)` is not defined, so I'm not sure a `powermod` method makes sense?

It seems like you want a matrix of modular numbers ala https://github.com/scheinerman/Mods.jl, in which case the ordinary exponentiation function should work?

src/matmul.jl
```diff
@@ -1503,3 +1503,49 @@ end
...
+        if isodd(p)
+            if res_set
+                mul!(res_tmp, res, x_p)
+                res .= res_tmp .% m
+            else
+                res .= x_p
+                res_set = true
+            end
+        end
+        mul!(x_p_tmp, x_p, x_p)
```
stevengj, line 1546: This is vulnerable to overflow.

Shmuma: Yes, you're right, `Mod{p}` solves my problem way more elegantly.
Thanks a lot!

PR #1554, opened 2026-02-27
Trigger CI on master

DilumAluthge:

PR #1556, opened 2026-02-28
Avoid construction in convert

jishnub:
This follows the `convert` implementation for `AbstractArray`s and short-circuits if the input is already of the correct type. There's no actual construction required in this case.

Should fix the failure on master seen in https://github.com/JuliaLang/LinearAlgebra.jl/pull/1554

### DIFF
--- a/src/bidiag.jl
+++ b/src/bidiag.jl
@@ -249,6 +249,7 @@ AbstractMatrix{T}(A::Bidiagonal{T}) where {T} = copy(A)
 
 function convert(::Type{T}, A::AbstractMatrix) where T<:Bidiagonal
     checksquare(A)
+    A isa T && return A
     isbanded(A, -1, 1) || throw(InexactError(:convert, T, A))
     iszero(diagview(A, 1)) ? T(A, :L) :
         iszero(diagview(A, -1)) ? T(A, :U) : throw(InexactError(:convert, T, A))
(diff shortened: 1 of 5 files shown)


dkarrasch (approved):
Do you have any idea why this wasn't an issue before?

jishnub:
The method was seemingly broken in https://github.com/JuliaLang/LinearAlgebra.jl/pull/1212. It was working correctly on v1.12 before that PR was merged. I'm not entirely certain why this test didn't appear before. The Transpose seems to work even without the `convert`:
```julia
julia> B
4×4 Bidiagonal{Int64, UnitRange{Int64}}:
 1  1  ⋅  ⋅
 ⋅  2  2  ⋅
 ⋅  ⋅  3  3
 ⋅  ⋅  ⋅  4

julia> Transpose(B)
4×4 transpose(::Bidiagonal{Int64, UnitRange{Int64}}) with eltype Int64:
 1  ⋅  ⋅  ⋅
 1  2  ⋅  ⋅
 ⋅  2  3  ⋅
 ⋅  ⋅  3  4

julia> convert(typeof(B), B)
ERROR: MethodError: Cannot `convert` an object of type Vector{Int64} to an object of type UnitRange{Int64}

julia> VERSION
v"1.14.0-DEV.1818"
```
I can replicate the failure on using julia's commit 248531f9b6 but this works on e8947b0ecfc, so I wonder if this was just a regression on nightly.

Edit: Looks like https://github.com/JuliaLang/julia/commit/3e744a7861263bc0df7033d2d1981e3173ecebad was the fix

PR #1557, opened 2026-02-28
faster hessenberg eigvals

stevengj:
This is a port of https://github.com/JuliaLang/julia/pull/41420, which implements faster `eigvals` methods for `UpperHessenberg` matrices — about twice as fast for a 1000x1000 matrix on my machine, single-threaded.  (By adding a new `eigvals!` method.)

(The PR is simpler than the original because `LAPACK.hseqr!` was added in https://github.com/JuliaLang/julia/pull/47872, which was merged in Julia 1.10.0.)

One could also accelerate other cases, including `eigen`, as noted in that PR, but I don't have time to work on that and `eigvals` is useful in itself.  For example, to find roots of polynomials, one finds eigenvalues of a [companion matrix](https://en.wikipedia.org/wiki/Companion_matrix), and companion matrices are upper-Hessenberg (or transposes thereof); similarly for "colleague" matrices of orthogonal-polynomial bases.

### DIFF
--- a/test/hessenberg.jl
+++ b/test/hessenberg.jl
@@ -320,4 +320,14 @@ end
     @test U == U2
 end
 
+@testset "eigensolvers" begin
+    for T in (Float32, Float64, ComplexF32, ComplexF64)
+        H = UpperHessenberg(randn(T, 5,5))
+        λ = eigvals(H)
+        F = eigen(H)
+        @test λ ≈ eigvals(Matrix(H)) ≈ F.values
+        @test H * F.vectors ≈ F.vectors * Diagonal(λ)
+    end
+end
+
 end # module TestHessenberg
--- a/test/hessenberg.jl
+++ b/test/hessenberg.jl
@@ -324,8 +324,9 @@ end
     for T in (Float32, Float64, ComplexF32, ComplexF64)
         H = UpperHessenberg(randn(T, 5,5))
         λ = eigvals(H)
+        λ2 = @invoke eigvals!(copy(H)::UpperHessenberg) # test fallback
         F = eigen(H)
-        @test λ ≈ eigvals(Matrix(H)) ≈ F.values
+        @test λ ≈ eigvals(Matrix(H)) ≈ F.values ≈ λ2
         @test H * F.vectors ≈ F.vectors * Diagonal(λ)
     end
 end
--- a/test/hessenberg.jl
+++ b/test/hessenberg.jl
@@ -321,13 +321,16 @@ end
 end
 
 @testset "eigensolvers" begin
-    for T in (Float32, Float64, ComplexF32, ComplexF64)
+    for T in (Float16, Float32, Float64, ComplexF16, ComplexF32, ComplexF64)
         H = UpperHessenberg(randn(T, 5,5))
         λ = eigvals(H)
-        λ2 = @invoke eigvals!(copy(H)::UpperHessenberg) # test fallback
         F = eigen(H)
-        @test λ ≈ eigvals(Matrix(H)) ≈ F.values ≈ λ2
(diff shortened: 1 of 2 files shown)


stevengj:
~~(Not sure if we should add a check for upper-Hessenberg to the `eigvals` polyalgorithm?  We are checking for upper-triangular anyway, and so checking for upper-Hessenberg is only an additional O(n) cost.)~~  Nevermind, we don't check for triangular matrices right now, only Hermitian.

stevengj:
cc @dlfivefifty, who implemented a `hesseneigvals` function in ApproxFun.jl (https://github.com/JuliaApproximation/ApproxFunBase.jl/blob/57b908420ded32a56bea399fd260987907c009c6/src/LinearAlgebra/hesseneigs.jl#L2-L30) for the same reason.

dlfivefifty:
I assume the call to `LAPACK.gebal` fixes the column scaling? If I remember right, my `hesseneigs` had issues with inaccuracies when compared to a dense solver since I didn't rebalance the columns.

stevengj:
> I assume the call to `LAPACK.gebal` fixes the column scaling?

That's what it's supposed to do: the documentation says that it applies "a diagonal similarity transformation to rows and columns ... to make the rows and columns as close in norm as possible."

But if you have a test case where it was inaccurate without the scaling that would be great.

dlfivefifty:
No I wasn't saying there's a problem. I didn't know or call `LAPACK.gebal` so my version would be much less accurate. Good to see there's a simple solution!

stevengj:
CI failure seems unrelated:
```
Error in testset LinearAlgebra/cholesky:
Error During Test at none:1
  Got exception outside of a @test
```

test/hessenberg.jl
```diff
@@ -320,4 +320,15 @@ end
...
+@testset "eigensolvers" begin
+    for T in (Float32, Float64, ComplexF32, ComplexF64)
+        H = UpperHessenberg(randn(T, 5,5))
+        λ = eigvals(H)
+        λ2 = @invoke eigvals!(copy(H)::UpperHessenberg) # test fallback
+        F = eigen(H)
+        @test λ ≈ eigvals(Matrix(H)) ≈ F.values ≈ λ2
+        @test H * F.vectors ≈ F.vectors * Diagonal(λ)
+    end
+end
```
araujoms, line 335:
```suggestion
@testset "eigensolvers" begin
    for T in (Float16, Float32, Float64, ComplexF16, ComplexF32, ComplexF64)
        H = UpperHessenberg(randn(T, 5,5))
        λ = eigvals(H)
        F = eigen(H)
        @test λ ≈ eigvals(Matrix(H)) ≈ F.values
        @test H * F.vectors ≈ F.vectors * Diagonal(λ)
        if T <: BlasFloat
            λ2 = @invoke eigvals!(copy(H)::UpperHessenberg) # test fallback
            @test λ ≈ λ2
        end
    end
end
```
This also tests the `Float16` and `ComplexF16` conversions.

test/hessenberg.jl
```diff
@@ -320,4 +320,18 @@ end
...
 end
 
+@testset "eigensolvers" begin
+    for T in (Float16, Float32, Float64, ComplexF16, ComplexF32, ComplexF64)
+        H = UpperHessenberg(randn(T, 5,5))
+        λ = eigvals(H)
+        F = eigen(H)
+        @test λ ≈ eigvals(Matrix(H)) ≈ F.values
+        @test H * F.vectors ≈ F.vectors * Diagonal(λ)
+        if T <: BlasFloat
```
araujoms:
```suggestion
        if T <: LinearAlgebra.BlasFloat
```

araujoms (approved):

stevengj:
Should be good to merge?

PR #1558, opened 2026-03-03
throw error when sqrt function is inaccurate due to < n-1 nonzero eigenvalues

### LINKED ISSUE #1548: incorrect sqrt([0 1; 0 0]): zero matrix instead of error
This is wrong:
```jl
julia> sqrt([0 1; 0 0])
2×2 Matrix{Float64}:
 0.0  0.0
 0.0  0.0
```
In fact, this matrix does not have a square root, as explained in [Björk and Hammarling (1983)](https://doi.org/10.1016/0024-3795(83)80010-X):

Image

Our answer for the second example here is also wrong:
```jl
julia> sqrt([0 1 0; 0 0 0; 0 0 0])
3×3 Matrix{Float64}:
 0.0  0.0  0.0
 0.0  0.0  0.0
 0.0  0.0  0.0
```
whereas a correct square root is:
```jl
julia> [0 0 1; 0 0 0; 0 1 0]^2 == [0 1 0; 0 0 0; 0 0 0]
true
```
cc @aravindh-krishnamoorthy who looked at this algorithm recently for https://github.com/JuliaLang/julia/pull/50661.

trevorgloe:
Fixes #1548

Now 
```julia
sqrt([0 1; 0 0])
```
prints
```julia
┌ Warning: Matrix has less than 1 nonzero eigenvalues. Square root may be inaccurate or matrix may not have a square root.
└ @ LinearAlgebra ~/Documents/julia_dev/LinearAlgebra.jl/src/triangular.jl:2685
```

### DIFF
--- a/src/dense.jl
+++ b/src/dense.jl
@@ -996,7 +996,7 @@ log(A::AdjointAbsMat) = adjoint(log(parent(A)))
 log(A::TransposeAbsMat) = transpose(log(parent(A)))
 
 """
-    sqrt(A::AbstractMatrix)
+    sqrt(A::AbstractMatrix; check=true)
 
 If `A` has no negative real eigenvalues, compute the principal matrix square root of `A`,
 that is the unique matrix ``X`` with eigenvalues having positive real part such that
@@ -1016,7 +1016,7 @@ and then the complex square root of the triangular factor.
 If a real square root exists, then an extension of this method [^H87] that computes the real
 Schur form and then the real square root of the quasi-triangular factor is instead used.
 
-When an $n \times n$ matrix has fewer than $n-1$ nonzero eigenvalues, the square root may not exist, so a warning is printed. Eigenvalues of magnitude less than `eps(max|λ|)` are treated as zero.
+When an n \times n matrix has fewer than n-1 nonzero eigenvalues, the square root may not exist. When the `check` flag is true, the algorithm will verify `X^2≈A` and throw and error if not.
 
 [^BH83]:
 
@@ -1045,7 +1045,7 @@ julia> sqrt(A)
 """
 sqrt(::AbstractMatrix)
 
-function sqrt(A::AbstractMatrix{T}) where {T<:Union{Real,Complex}}
+function sqrt(A::AbstractMatrix{T}; check=true) where {T<:Union{Real,Complex}}
     if checksquare(A) == 0
         return copy(float(A))
     elseif isdiag(A)
@@ -1055,33 +1055,33 @@ function sqrt(A::AbstractMatrix{T}) where {T<:Union{Real,Complex}}
             return applydiagonal(sqrt, A)
         end
     elseif ishermitian(A)
-        return _safe_parent(sqrt(Hermitian(A)))
+        return _safe_parent(sqrt(Hermitian(A))) # dont need to check for hermitian matrices
     elseif istriu(A)
-        return triu!(parent(sqrt(UpperTriangular(A))))
+        return triu!(parent(sqrt(UpperTriangular(A), check=check)))
     elseif isreal(A)
         SchurF = schur(real(A))
         if istriu(SchurF.T)
-            sqrtA = SchurF.Z * sqrt(UpperTriangular(SchurF.T)) * SchurF.Z'
+            sqrtA = SchurF.Z * sqrt(UpperTriangular(SchurF.T), check=check) * SchurF.Z'
         else
             # real sqrt exists whenever no eigenvalues are negative
             is_sqrt_real = !any(x -> isreal(x) && real(x) < 0, SchurF.values)
             # sqrt_quasitriu uses LAPACK functions for non-triu inputs
             if typeof(sqrt(zero(T))) <: BlasFloat && is_sqrt_real
-                sqrtA = SchurF.Z * sqrt_quasitriu(SchurF.T, SchurF.values) * SchurF.Z'
+                sqrtA = SchurF.Z * sqrt_quasitriu(SchurF.T, SchurF.values, check=check) * SchurF.Z'
             else
                 SchurS = Schur{Complex}(SchurF)
-                sqrtA = SchurS.Z * sqrt(UpperTriangular(SchurS.T)) * SchurS.Z'
+                sqrtA = SchurS.Z * sqrt(UpperTriangular(SchurS.T), check=check) * SchurS.Z'
             end
         end
         return eltype(A) <: Complex ? complex(sqrtA) : sqrtA
     else
         SchurF = schur(A)
-        return SchurF.vectors * sqrt(UpperTriangular(SchurF.T)) * SchurF.vectors'
+        return SchurF.vectors * sqrt(UpperTriangular(SchurF.T), check=check) * SchurF.vectors'
     end
 end
 
-sqrt(A::AdjointAbsMat) = adjoint(sqrt(parent(A)))
-sqrt(A::TransposeAbsMat) = transpose(sqrt(parent(A)))
+sqrt(A::AdjointAbsMat; check=true) = adjoint(sqrt(parent(A), check=check))
+sqrt(A::TransposeAbsMat, check=true) = transpose(sqrt(parent(A), check=check))
 
 """
     cbrt(A::AbstractMatrix{<:Real})
--- a/src/dense.jl
+++ b/src/dense.jl
@@ -1016,6 +1016,8 @@ and then the complex square root of the triangular factor.
 If a real square root exists, then an extension of this method [^H87] that computes the real
 Schur form and then the real square root of the quasi-triangular factor is instead used.
 
+When there are less than n-1 nonzero eigenvalues, the square root may not exist so a warning is printed. Eigenvalues of magnitude less than `eps(max|λ|)` are treated as zero.
+
 [^BH83]:
 
     Åke Björck and Sven Hammarling, "A Schur method for the square root of a matrix",
@@ -1065,7 +1067,7 @@ function sqrt(A::AbstractMatrix{T}) where {T<:Union{Real,Complex}}
             is_sqrt_real = !any(x -> isreal(x) && real(x) < 0, SchurF.values)
             # sqrt_quasitriu uses LAPACK functions for non-triu inputs
             if typeof(sqrt(zero(T))) <: BlasFloat && is_sqrt_real
-                sqrtA = SchurF.Z * sqrt_quasitriu(SchurF.T) * SchurF.Z'
+                sqrtA = SchurF.Z * sqrt_quasitriu(SchurF.T, SchurF.values) * SchurF.Z'
             else
                 SchurS = Schur{Complex}(SchurF)
                 sqrtA = SchurS.Z * sqrt(UpperTriangular(SchurS.T)) * SchurS.Z'
--- a/src/dense.jl
+++ b/src/dense.jl
@@ -1016,7 +1016,7 @@ and then the complex square root of the triangular factor.
 If a real square root exists, then an extension of this method [^H87] that computes the real
 Schur form and then the real square root of the quasi-triangular factor is instead used.
 
-When there are less than n-1 nonzero eigenvalues, the square root may not exist so a warning is printed. Eigenvalues of magnitude less than `eps(max|λ|)` are treated as zero.
+When an $n \times n$ matrix has fewer than $n-1$ nonzero eigenvalues, the square root may not exist, so a warning is printed. Eigenvalues of magnitude less than `eps(max|λ|)` are treated as zero.
 
 [^BH83]:
 
--- a/src/dense.jl
+++ b/src/dense.jl
@@ -1016,7 +1016,7 @@ and then the complex square root of the triangular factor.
 If a real square root exists, then an extension of this method [^H87] that computes the real
 Schur form and then the real square root of the quasi-triangular factor is instead used.
 
-When an $n \times n$ matrix has fewer than $n-1$ nonzero eigenvalues, the square root may not exist. In this case, and when the `check` flag is true, the algorithm will verify `X^2≈A` and throw and error if not. 
+When an $n \times n$ matrix has fewer than $n-1$ nonzero eigenvalues, the square root may not exist. In this case, and when the `check` flag is true, the algorithm will verify `X^2≈A` and throw an error if not. 
 
 [^BH83]:
 
--- a/src/dense.jl
+++ b/src/dense.jl
@@ -1016,7 +1016,7 @@ and then the complex square root of the triangular factor.
 If a real square root exists, then an extension of this method [^H87] that computes the real
 Schur form and then the real square root of the quasi-triangular factor is instead used.
 
-When an n \times n matrix has fewer than n-1 nonzero eigenvalues, the square root may not exist. When the `check` flag is true, the algorithm will verify `X^2≈A` and throw and error if not.
+When an $n \times n$ matrix has fewer than $n-1$ nonzero eigenvalues, the square root may not exist. In this case, and when the `check` flag is true, the algorithm will verify `X^2≈A` and throw and error if not. 
 
 [^BH83]:
 
--- a/src/dense.jl
+++ b/src/dense.jl
@@ -1057,7 +1057,7 @@ function sqrt(A::AbstractMatrix{T}; check=true) where {T<:Union{Real,Complex}}
     elseif ishermitian(A)
         return _safe_parent(sqrt(Hermitian(A))) # dont need to check for hermitian matrices
     elseif istriu(A)
-        return triu!(parent(sqrt(UpperTriangular(A), check=check)))
+        return triu!(parent(sqrt(UpperTriangular(A); check)))
     elseif isreal(A)
         SchurF = schur(real(A))
         if istriu(SchurF.T)
--- a/src/dense.jl
+++ b/src/dense.jl
@@ -1045,7 +1045,7 @@ julia> sqrt(A)
 """
 sqrt(::AbstractMatrix)
 
-function sqrt(A::AbstractMatrix{T}; check=true) where {T<:Union{Real,Complex}}
+function sqrt(A::AbstractMatrix{T}; check::Bool=true) where {T<:Union{Real,Complex}}
     if checksquare(A) == 0
         return copy(float(A))
     elseif isdiag(A)
@@ -1061,7 +1061,7 @@ function sqrt(A::AbstractMatrix{T}; check=true) where {T<:Union{Real,Complex}}
     elseif isreal(A)
         SchurF = schur(real(A))
         if istriu(SchurF.T)
-            sqrtA = SchurF.Z * sqrt(UpperTriangular(SchurF.T), check=check) * SchurF.Z'
+            sqrtA = SchurF.Z * sqrt(UpperTriangular(SchurF.T); check) * SchurF.Z'
         else
             # real sqrt exists whenever no eigenvalues are negative
             is_sqrt_real = !any(x -> isreal(x) && real(x) < 0, SchurF.values)
@@ -1070,18 +1070,18 @@ function sqrt(A::AbstractMatrix{T}; check=true) where {T<:Union{Real,Complex}}
                 sqrtA = SchurF.Z * sqrt_quasitriu(SchurF.T, SchurF.values, check=check) * SchurF.Z'
             else
                 SchurS = Schur{Complex}(SchurF)
-                sqrtA = SchurS.Z * sqrt(UpperTriangular(SchurS.T), check=check) * SchurS.Z'
+                sqrtA = SchurS.Z * sqrt(UpperTriangular(SchurS.T); check) * SchurS.Z'
             end
         end
         return eltype(A) <: Complex ? complex(sqrtA) : sqrtA
     else
         SchurF = schur(A)
-        return SchurF.vectors * sqrt(UpperTriangular(SchurF.T), check=check) * SchurF.vectors'
+        return SchurF.vectors * sqrt(UpperTriangular(SchurF.T); check) * SchurF.vectors'
     end
 end
 
-sqrt(A::AdjointAbsMat; check=true) = adjoint(sqrt(parent(A), check=check))
-sqrt(A::TransposeAbsMat, check=true) = transpose(sqrt(parent(A), check=check))
+sqrt(A::AdjointAbsMat; check::Bool=true) = adjoint(sqrt(parent(A); check))
+sqrt(A::TransposeAbsMat, check::Bool=true) = transpose(sqrt(parent(A); check))
 
 """
     cbrt(A::AbstractMatrix{<:Real})
--- a/src/dense.jl
+++ b/src/dense.jl
@@ -1016,7 +1016,7 @@ and then the complex square root of the triangular factor.
 If a real square root exists, then an extension of this method [^H87] that computes the real
 Schur form and then the real square root of the quasi-triangular factor is instead used.
 
-When an ``n × n`` non-Hermitian matrix has fewer than ``n - 1`` nonzero eigenvalues, the square root may not exist. In this case, and when the `check` flag is true, the algorithm will verify `X^2≈A` and throw an error if not. 
+When a non-Hermitian matrix has two or more null eigenvalues, the square root may not exist. In this case, and when the `check` flag is true, the algorithm will verify `X^2≈A` and throw an error if not. 
 
 [^BH83]:
 
--- a/src/dense.jl
+++ b/src/dense.jl
@@ -1081,7 +1081,7 @@ function sqrt(A::AbstractMatrix{T}; check::Bool=true) where {T<:Union{Real,Compl
 end
 
 sqrt(A::AdjointAbsMat; check::Bool=true) = adjoint(sqrt(parent(A); check))
-sqrt(A::TransposeAbsMat, check::Bool=true) = transpose(sqrt(parent(A); check))
+sqrt(A::TransposeAbsMat; check::Bool=true) = transpose(sqrt(parent(A); check))
 
 """
     cbrt(A::AbstractMatrix{<:Real})
--- a/src/dense.jl
+++ b/src/dense.jl
@@ -1016,7 +1016,7 @@ and then the complex square root of the triangular factor.
 If a real square root exists, then an extension of this method [^H87] that computes the real
 Schur form and then the real square root of the quasi-triangular factor is instead used.
 
-When an $n \times n$ matrix has fewer than $n-1$ nonzero eigenvalues, the square root may not exist. In this case, and when the `check` flag is true, the algorithm will verify `X^2≈A` and throw an error if not. 
+When an $n \times n$ non-Hermitian matrix has fewer than $n-1$ nonzero eigenvalues, the square root may not exist. In this case, and when the `check` flag is true, the algorithm will verify `X^2≈A` and throw an error if not. 
 
 [^BH83]:
 
--- a/src/dense.jl
+++ b/src/dense.jl
@@ -1067,7 +1067,7 @@ function sqrt(A::AbstractMatrix{T}; check::Bool=true) where {T<:Union{Real,Compl
             is_sqrt_real = !any(x -> isreal(x) && real(x) < 0, SchurF.values)
             # sqrt_quasitriu uses LAPACK functions for non-triu inputs
             if typeof(sqrt(zero(T))) <: BlasFloat && is_sqrt_real
-                sqrtA = SchurF.Z * sqrt_quasitriu(SchurF.T, SchurF.values, check=check) * SchurF.Z'
+                sqrtA = SchurF.Z * sqrt_quasitriu(SchurF.T, SchurF.values; check) * SchurF.Z'
             else
                 SchurS = Schur{Complex}(SchurF)
                 sqrtA = SchurS.Z * sqrt(UpperTriangular(SchurS.T); check) * SchurS.Z'
--- a/src/dense.jl
+++ b/src/dense.jl
@@ -1055,7 +1055,7 @@ function sqrt(A::AbstractMatrix{T}; check::Bool=true) where {T<:Union{Real,Compl
             return applydiagonal(sqrt, A)
         end
     elseif ishermitian(A)
-        return _safe_parent(sqrt(Hermitian(A))) # dont need to check for hermitian matrices
+        return _safe_parent(sqrt(Hermitian(A); check))
     elseif istriu(A)
         return triu!(parent(sqrt(UpperTriangular(A); check)))
     elseif isreal(A)
--- a/src/dense.jl
+++ b/src/dense.jl
@@ -1016,7 +1016,7 @@ and then the complex square root of the triangular factor.
 If a real square root exists, then an extension of this method [^H87] that computes the real
 Schur form and then the real square root of the quasi-triangular factor is instead used.
 
-When an $n \times n$ non-Hermitian matrix has fewer than $n-1$ nonzero eigenvalues, the square root may not exist. In this case, and when the `check` flag is true, the algorithm will verify `X^2≈A` and throw an error if not. 
+When an ``n × n`` non-Hermitian matrix has fewer than ``n - 1`` nonzero eigenvalues, the square root may not exist. In this case, and when the `check` flag is true, the algorithm will verify `X^2≈A` and throw an error if not. 
 
 [^BH83]:
 
--- a/src/symmetric.jl
+++ b/src/symmetric.jl
@@ -989,7 +989,8 @@ function log(A::SelfAdjoint)
 end
 
 # sqrt has rtol kwarg to handle matrices that are semidefinite up to roundoff errors
-# note that `check` is passed to the function for consistancy with the dense sqrt function, but it is not used as the matrix should always at least have a complex square root (just from eigendecomposition)
+# note a `check` keyword argument is accepted for consistency with the generic `sqrt` function,
+# but it is not used here, as a self-adjoint matrix is diagonalizable and hence always has a sqrt.
 function sqrt(A::SelfAdjoint; check=true, rtol = eps(real(float(eltype(A)))) * size(A, 1))
     F = eigen(A)
     λ₀ = -maximum(abs, F.values) * rtol # treat λ ≥ λ₀ as "zero" eigenvalues up to roundoff
--- a/src/symmetric.jl
+++ b/src/symmetric.jl
@@ -991,7 +991,7 @@ end
 # sqrt has rtol kwarg to handle matrices that are semidefinite up to roundoff errors
 # note a `check` keyword argument is accepted for consistency with the generic `sqrt` function,
 # but it is not used here, as a self-adjoint matrix is diagonalizable and hence always has a sqrt.
-function sqrt(A::SelfAdjoint; check=true, rtol = eps(real(float(eltype(A)))) * size(A, 1))
+function sqrt(A::SelfAdjoint; check::Bool=true, rtol = eps(real(float(eltype(A)))) * size(A, 1))
     F = eigen(A)
     λ₀ = -maximum(abs, F.values) * rtol # treat λ ≥ λ₀ as "zero" eigenvalues up to roundoff
     if all(λ -> λ ≥ λ₀, F.values)
--- a/src/symmetric.jl
+++ b/src/symmetric.jl
@@ -989,7 +989,8 @@ function log(A::SelfAdjoint)
 end
 
 # sqrt has rtol kwarg to handle matrices that are semidefinite up to roundoff errors
-function sqrt(A::SelfAdjoint; rtol = eps(real(float(eltype(A)))) * size(A, 1))
+# note that `check` is passed to the function for consistancy with the dense sqrt function, but it is not used as the matrix should always at least have a complex square root (just from eigendecomposition)
+function sqrt(A::SelfAdjoint; check=true, rtol = eps(real(float(eltype(A)))) * size(A, 1))
     F = eigen(A)
     λ₀ = -maximum(abs, F.values) * rtol # treat λ ≥ λ₀ as "zero" eigenvalues up to roundoff
     if all(λ -> λ ≥ λ₀, F.values)
--- a/src/triangular.jl
+++ b/src/triangular.jl
@@ -2650,7 +2650,7 @@ end
 
 # End of auxiliary functions for matrix logarithm and matrix power
 
-sqrt(A::UpperTriangular) = sqrt_quasitriu(A)
+sqrt(A::UpperTriangular) = sqrt_quasitriu(A, diagview(A)) # matrix is upper triangular, so eigenvalues are just the diagonals
 function sqrt(A::UnitUpperTriangular{T}) where T
     B = A.data
     t = typeof(sqrt(zero(T)))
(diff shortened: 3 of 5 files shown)


src/triangular.jl
```diff
@@ -2676,6 +2676,15 @@ sqrt(A::UnitLowerTriangular) = copy(transpose(sqrt(copy(transpose(A)))))
 # square root of upper triangular or real upper quasitriangular matrix
 function sqrt_quasitriu(A0; blockwidth = eltype(A0) <: Complex ? 512 : 256)
     n = checksquare(A0)
+    if isa(eltype(A0), AbstractFloat)
+        nonzero_eig = sum([abs(e)>floatmin(eltype(A0)) for e in diag(A0)]) # if its a float type, check if the float is greater than the smallest representable float
```
stevengj: This seems like a weird test to me.  Why is `floatmin` bad but `nextfloat(floatmin(...))` is okay?  (By the way, `floatmin` is the smallest non-subnormal, not the smallest "representable" nonzero magnitude.)

Isn't the correct measure of "small" here some multiple of `eps()` times some norm of the matrix?

src/triangular.jl
```diff
@@ -2676,6 +2676,15 @@ sqrt(A::UnitLowerTriangular) = copy(transpose(sqrt(copy(transpose(A)))))
...
+    else
+        nonzero_eig = sum([!iszero(e) for e in diag(A0)]) # check if there are less than n-1 nonzero eigenvalues
```
stevengj: No point in allocating here.  Can just do:

```suggestion
        nonzero_eig = count(!iszero, diag(A0)) # check if there are less than n-1 nonzero eigenvalues
```

src/triangular.jl
```diff
@@ -2676,6 +2676,15 @@ sqrt(A::UnitLowerTriangular) = copy(transpose(sqrt(copy(transpose(A)))))
...
+    end
+    if (nonzero_eig < n - 1)
+        @warn "Matrix has less than $(n - 1) nonzero eigenvalues. Square root may be inaccurate or matrix may not have a square root."
```
```suggestion
        @warn "Matrix has fewer than n-1=$(n - 1) nonzero eigenvalues. Square root may be inaccurate or matrix may not have a square root."
```

src/triangular.jl
trevorgloe: That's a fair point. I was pondering a good way to do this test. It seems from [this](https://github.com/JuliaLang/julia/issues/23376) discussion that comparing with zero using `isapprox()` can be tricky, so my thinking was that if the number on the diagonal is `floatmin()`, its probably supposed to be zero, or at least close enough to give a warning. But maybe this is not correct. 

I could make it `abs(e)>=floatmin(eltype(A0))` so there will be warning if the eigenvalue is smaller than the smallest non-subnormal.

trevorgloe: But now that I think about it, `eps()` times the norm of the matrix does seem like the best measure of "small" in this case. I'll change the test to that.

stevengj:
Note that the matrix here is only quasi-upper-triangular, so the diagonal entries are not quite eigenvalues.  For example, the diagonal block `[0 -1; 1 0]` represents the eigenvalues $\pm i$, even though it has 0's on the diagonal.

Maybe we should
1. modify `sqrt_quasitriu` to take an `eigvals::AbstractVector` argument.   This can either be `SchurF.values` if it comes from a Schur factorization or just a `diagview` of the matrix if it is upper triangular.    
2. when checking for zero eigenvalues, use a tolerance of `atol = eps(normInf(eigvals))` based on the maximum |λ| (noting that `normInf` always returns a float type), i.e. use `num_zeros = count(eigval -> abs(eigval) < atol, eigvals)`

src/triangular.jl
```diff
@@ -2676,6 +2676,15 @@ sqrt(A::UnitLowerTriangular) = copy(transpose(sqrt(copy(transpose(A)))))
 # square root of upper triangular or real upper quasitriangular matrix
 function sqrt_quasitriu(A0; blockwidth = eltype(A0) <: Complex ? 512 : 256)
     n = checksquare(A0)
+    if isa(eltype(A0), AbstractFloat)
+        nonzero_eig = count(x->abs(x)>=eps(eltype(A0))*norm(A0), diag(A0)) # if its a float, check if the eigenvalue is greater than eps*(2-norm of matrix)
+    else
```
stevengj: note that this `else` branch will be triggered if the `eltype` is `Complex`, which isn't right.

src/triangular.jl
```diff
@@ -2676,6 +2676,15 @@ sqrt(A::UnitLowerTriangular) = copy(transpose(sqrt(copy(transpose(A)))))
 # square root of upper triangular or real upper quasitriangular matrix
 function sqrt_quasitriu(A0; blockwidth = eltype(A0) <: Complex ? 512 : 256)
     n = checksquare(A0)
+    if isa(eltype(A0), AbstractFloat)
+        nonzero_eig = count(x->abs(x)>=eps(eltype(A0))*norm(A0), diag(A0)) # if its a float, check if the eigenvalue is greater than eps*(2-norm of matrix)
```
stevengj: This calls `norm` repeatedly, once per element, so it is $O(n^3)$

trevorgloe:
The PR has been updated accordingly. Will it always be guaranteed that `generic_normInf(evals)` will be a `float` type that we can take `eps` of? It seemed to be the case for any custom types I defined when I tested it, but I just worry that someone could be using a weird numeric type in the matrix that wouldn't allow this.

src/dense.jl
```diff
@@ -1016,6 +1016,8 @@ and then the complex square root of the triangular factor.
 If a real square root exists, then an extension of this method [^H87] that computes the real
 Schur form and then the real square root of the quasi-triangular factor is instead used.
 
+When there are less than n-1 nonzero eigenvalues, the square root may not exist so a warning is printed. Eigenvalues of magnitude less than `eps(max|λ|)` are treated as zero.
```
```suggestion
When an $n \times n$ matrix has fewer than $n-1$ nonzero eigenvalues, the square root may not exist, so a warning is printed. Eigenvalues of magnitude less than `eps(max|λ|)` are treated as zero.
```

src/triangular.jl
```diff
@@ -2674,8 +2674,15 @@ sqrt(A::UnitLowerTriangular) = copy(transpose(sqrt(copy(transpose(A)))))
 # Auxiliary functions for matrix square root
 
 # square root of upper triangular or real upper quasitriangular matrix
-function sqrt_quasitriu(A0; blockwidth = eltype(A0) <: Complex ? 512 : 256)
+# A0 is triangular or quasitriangular matrix, evals is the eigenvalues
+function sqrt_quasitriu(A0, evals::AbstractVector; blockwidth = eltype(A0) <: Complex ? 512 : 256)
     n = checksquare(A0)
+    atol = eps(generic_normInf(evals)) # should work for any numeric data type
+    nonzero_eig = count(x->abs(x)>=atol, evals) # count number of eigenvalues that are \approx 0
```
```suggestion
    nonzero_eig = count(x -> abs(x) > atol, evals) # count eigenvalues ≉ 0
```
Maybe should be `>`, not `≥`, in case `atol == 0` (though I don't think that should happen since `eps(0.0) > 0`).

stevengj:
An alternative to a warning here would be to explicitly check `sqrt(A)^2 ≈ A` when `nonzero_eig < n-1` and throw an exception if this fails.  The additional time for the matmul is not very significant and it seems a lot safer to throw an exception than to give the wrong result with a warning.

trevorgloe:
> An alternative to a warning here would be to explicitly check `sqrt(A)^2 ≈ A` when `nonzero_eig < n-1` and throw an exception if this fails. The additional time for the matmul is not very significant and it seems a lot safer to throw an exception than to give the wrong result with a warning.

Throwing an exception feels a little extreme here, as if the algorithm fails I think it might be better to let the user decide how to handle it or how to error. It could be something along the lines of returning a flag. I'm thinking about how some optimization libraries in python handle the possibility of the algorithm failing. 

I also think about the case like this
```julia
A = [1e-100 1 0; 0 0 0; 0 0 0]
sqrt(A)^2 # is A
B = [0 1 0; 0 0 0; 0 0 0]
sqrt(B)^2 # fails
```
Even though the algorithm succeeds in the first case, the warning kind of lets the user know that they are within the dangerous regime where the algorithm might fail. But I guess there are still edge cases where a small perturbation will take you from being in the regime where there are n-1 nonzero eigenvalues to the algorithm failing, which the warning wouldn't alert you of. Having the warning for cases like the one above may still be helpful for catching some of the those dangerous situations.

stevengj:
> Throwing an exception feels a little extreme here, as if the algorithm fails I think it might be better to let the user decide how to handle it or how to error.

A printed warning is much harder to detect programmatically than a thrown exception.  You can always catch the exception.

> It could be something along the lines of returning a flag

That would be a breaking change.

Of course, so is throwing an exception, but arguably returning a completely incorrect result is a bug, and we don't maintain backward compatibility with bugs.

Another option would be to add a new argument to `sqrt(A; check=true)` that adds the check and the exception, maybe defaulting to `true` for safety, so if you want to ignore problems you have to pass `check=false`.  (Similar to the `check` argument for `lu` or `cholesky`.)

trevorgloe:
> A printed warning is much harder to detect programmatically than a thrown exception. You can always catch the exception.

You're right, it would probably make more sense to throw an error and let the user catch that if they want.

> That would be a breaking change.
> 
> Of course, so is throwing an exception, but arguably returning a completely incorrect result is a bug, and we don't maintain backward compatibility with bugs.

Opps, yes, I didn't think about that. Probably best to not do that then. 

> Another option would be to add a new argument to `sqrt(A; check=true)` that adds the check and the exception, maybe defaulting to `true` for safety, so if you want to ignore problems you have to pass `check=false`. (Similar to the `check` argument for `lu` or `cholesky`.)

I think in light of the fact that a user could just catch the error if the function errors, the `check` flag might be a bit redundant. If the user doesn't want the function to error, they could just catch the exception. But maybe it doesn't hurt to have a redundant way to stop the function from erroring.

stevengj:
> But maybe it doesn't hurt to have a redundant way to stop the function from erroring.

Passing `check=false` not only stops the function from throwing, it also allows you to inspect whatever output it returns (even if it's garbage), whereas if it throws an error you don't get to see what the return value would have been.

trevorgloe:
Ok, I've updated the PR so it will error when the `sqrt` algorithm fails. When it does fail, it checks if the number of nonzero eigenvalues is less than `n-1` so it can report that as the reason why. I also added a small test :).

Question: I was getting an error with the change in the docstring that was made, specifically, where it was 
```
$n \times n$ matrix has fewer than $n-1$
```
The `$` was causing an error on my end and I wasn't able to test the code with it there. Sorry for removing it. Is this a problem on my end? 

Also I found that `A^2` would error when `A` was populated with `Furlong` elements, but `A*A` would run just fine. Is that a bug with the `^` operator?

trevorgloe:
Shoot, I'm realizing that I misunderstood the check. The algorithm shouldn't ever fail when the number of nonzero eigenvalues are at least `n-1`, so the test shouldn't need to happen in that case. My bad. Let me fix this then I'll update the PR.

src/triangular.jl
```diff
@@ -2668,20 +2669,15 @@ function sqrt(A::UnitUpperTriangular{T}) where T
...
 # square root of upper triangular or real upper quasitriangular matrix
 # A0 is triangular or quasitriangular matrix, evals is the eigenvalues
-function sqrt_quasitriu(A0, evals::AbstractVector; blockwidth = eltype(A0) <: Complex ? 512 : 256)
+function sqrt_quasitriu(A0, evals::AbstractVector; blockwidth = eltype(A0) <: Complex ? 512 : 256, check=true)
     n = checksquare(A0)
-    atol = eps(generic_normInf(evals)) # should work for any numeric data type
-    nonzero_eig = count(x -> abs(x) > atol, evals) # count eigenvalues ≉ 0
-    if (nonzero_eig < n - 1)
-        @warn "Matrix has fewer than n-1=$(n - 1) nonzero eigenvalues. Square root may be inaccurate or matrix may not have a square root."
-    end
```
stevengj: I think it should only perform a check if `nonzero_eig < n - 1` (and that this should be documented).

(And it probably shouldn't use the default tolerance for `isapprox` … requiring half of the digits to match seems too stringent … we are just trying to check for catastrophic failure here.)

trevorgloe:
PR is updated. I made the check require that about 1/4 of the digits match, which seemed reasonable to me. 

I am still confused a bit about why the `^` operator doesn't seem to work with `Furlong` type elements, but the alternate path I used when the `eltype` is not real or complex works with the test suite. 

I also added back in the `$` to the docstring, I'm assuming that the errors I'm seeing on my end are just something weird about my setup and nothing actually wrong with the docstring there.

src/triangular.jl
```diff
@@ -2705,14 +2705,21 @@ function sqrt_quasitriu(A0, evals::AbstractVector; blockwidth = eltype(A0) <: Co
...
-        if !test
-            throw_sqrterror(evals, n)
+        atol = eps(generic_normInf(evals)) # should work for any numeric data type
+        nonzero_eig = count(x -> abs(x) > atol, evals) # count eigenvalues ≉ 0
+        if (nonzero_eig < n - 1) # in the regime where the algorithm could fail
+            if eltype(A0) <: Real || eltype(A0) <: Complex
+                rtol_test = eps(float(one(eltype(A0))))^(1/4) # check that 1/4 of the digits match
+                test = isapprox(R^2,A0,rtol=rtol_test)
+            else
+                test = generic_normInf(R*R - A0) <= (eps(generic_normInf(A0)))^(1/4) # when eltype is not real or complex, use the norm
```
stevengj: We might as well use this in both cases, rather than call `isapprox`?

stevengj: You can save an allocation by replacing `generic_normInf(R*R - A0)` with `generic_normInf(R*R .-= A0)`

src/triangular.jl
```diff
@@ -2705,14 +2705,21 @@ function sqrt_quasitriu(A0, evals::AbstractVector; blockwidth = eltype(A0) <: Co
...
+            end
+            if !test
+                throw(ArgumentError("Matrix has fewer than n-1=$(n - 1) nonzero eigenvalues. Square root may be inaccurate or matrix may not have a square root."))
```
stevengj: The message should reflect the check failure, and mention the possibility of passing `check=true` to ignore the error.

src/dense.jl
```diff
@@ -1016,6 +1016,8 @@ and then the complex square root of the triangular factor.
 If a real square root exists, then an extension of this method [^H87] that computes the real
 Schur form and then the real square root of the quasi-triangular factor is instead used.
 
+When an $n \times n$ matrix has fewer than $n-1$ nonzero eigenvalues, the square root may not exist. In this case, and when the `check` flag is true, the algorithm will verify `X^2≈A` and throw and error if not.
```
```suggestion
When an $n \times n$ matrix has fewer than $n-1$ nonzero eigenvalues, the square root may not exist. In this case, and when the `check` flag is true, the algorithm will verify `X^2≈A` and throw an error if not. 
```

src/triangular.jl
stevengj: (You don't need the extra parens around `eps(...)`.  And should use `^(1//4)` rather than `^(1/4)` to avoid accidental promotion to `Float64`.

test/dense.jl
```diff
@@ -1000,7 +1000,7 @@ end
     @test sqrt(x2)^2 ≈ x2
 
     x3 = [-1 -eps() 0 0; eps() -1 0 0; 0 0 -1 -eps(); 0 0 eps() Inf]
-    @test all(isnan, sqrt(x3))
+    @test all(isnan, sqrt(x3, check=false))
```
stevengj: I don't think we should be hitting the check in this case since the eigenvalues are NaN, not zero. 

The problem is that `NaN > ...` returns `false`.

An easy fix would be to instead count:
```jl
zero_eig = count(x -> abs(x) <= atol, evals) # count eigenvalues ≈ 0
```
and then check `zero_eig > 1`

src/triangular.jl
trevorgloe: Fixed, did not know about that last part.

src/triangular.jl
trevorgloe: Fixed. I figured the wording "failed to produce matrix with X^2=A..." would work there.

test/dense.jl
trevorgloe: Fixed this. Although I'm not sure if you think `check=false` should be removed from this test. I don't think it should really matter now as the NaNs shouldn't show up at zeros.

stevengj: Yes, I would remove `check=false` from this test as it is not needed.
```suggestion
    @test all(isnan, sqrt(x3))
```

src/triangular.jl
```diff
@@ -2702,6 +2705,19 @@ function sqrt_quasitriu(A0; blockwidth = eltype(A0) <: Complex ? 512 : 256)
         R = zeros(Tc, size(A0))
     end
     _sqrt_quasitriu!(R, A; blockwidth=blockwidth, n=n)
+
+    # check that the algorithm worked
+    if check
+        atol = eps(generic_normInf(evals)) # should work for any numeric data type
+        zero_eig = count(x -> abs(x) <= atol, evals) # count eigenvalues = 0
```
```suggestion
        zero_eig = count(x -> abs(x) <= atol, evals) # count eigenvalues ≈ 0
```

src/triangular.jl
```diff
@@ -2702,6 +2705,19 @@ function sqrt_quasitriu(A0; blockwidth = eltype(A0) <: Complex ? 512 : 256)
...
+        if (zero_eig > 1) # in the regime where the algorithm could fail
+            test = generic_normInf(R*R .-= A0) <= eps(generic_normInf(A0))^(1//4) # when eltype is not real or complex, use the norm
```
```suggestion
            test = generic_normInf(R*R .-= A0) <= eps(generic_normInf(A0))^(1//4)
```

stevengj: `eps(generic_normInf(A0))^(1//4)` is not correct here (the units are not correct … imagine a dimensionful matrix $A$): I think you want to take the 1/4 root of the *relative* `eps()` and then multiply it by $\Vert A_0 \Vert_\infty$.

In particular: 
```suggestion
            test = generic_normInf(R*R .-= A0) <= eps(typeof(atol))^(1//4) * generic_normInf(A0)
```

(This requires about 1/4 of the significant digits to match.)

trevorgloe: I see, its the relative tolerance that should be raised to the power of 1/4, I was using an absolute tolerance. My bad there.

trevorgloe: I think I did something that made the suggestion outdated, so I just committed the change myself.

src/dense.jl
```diff
@@ -1053,33 +1055,33 @@ function sqrt(A::AbstractMatrix{T}) where {T<:Union{Real,Complex}}
             return applydiagonal(sqrt, A)
         end
     elseif ishermitian(A)
-        return _safe_parent(sqrt(Hermitian(A)))
+        return _safe_parent(sqrt(Hermitian(A))) # dont need to check for hermitian matrices
     elseif istriu(A)
-        return triu!(parent(sqrt(UpperTriangular(A))))
+        return triu!(parent(sqrt(UpperTriangular(A), check=check)))
```
stevengj: This can be shortened:
```suggestion
        return triu!(parent(sqrt(UpperTriangular(A); check)))
```
(and similarly below).

src/dense.jl
```diff
@@ -996,7 +996,7 @@ log(A::AdjointAbsMat) = adjoint(log(parent(A)))
 log(A::TransposeAbsMat) = transpose(log(parent(A)))
 
 """
-    sqrt(A::AbstractMatrix)
+    sqrt(A::AbstractMatrix; check=true)
```
stevengj, line 999: Note that the `SelfAdjoint` case (in `symmetric.jl`) takes an `rtol` keyword but no `check`

stevengj: (Though possibly in a future PR we can implement `rtol` for the quasi-triu case as well, in order to still give a real result for a slightly negative eigenvalue.)

src/triangular.jl
```diff
@@ -2702,6 +2705,19 @@ function sqrt_quasitriu(A0; blockwidth = eltype(A0) <: Complex ? 512 : 256)
...
     _sqrt_quasitriu!(R, A; blockwidth=blockwidth, n=n)
+
+    # check that the algorithm worked
+    if check
+        atol = eps(generic_normInf(evals)) # should work for any numeric data type
+        zero_eig = count(x -> abs(x) <= atol, evals) # count eigenvalues ≈ 0
+        if (zero_eig > 1) # in the regime where the algorithm could fail
+            test = generic_normInf(R*R .-= A0) <= eps(typeof(atol))^(1//4) * generic_normInf(A0)
+            if !test
+                throw(ArgumentError("Failed to produce matrix with X^2≈A. Set `check=false` to ignore. Matrix has fewer than n-1=$(n - 1) nonzero eigenvalues so square root may be inaccurate or matrix may not have a square root."))
```
```suggestion
                throw(ArgumentError("Failed to produce matrix with X^2≈A. Pass `check=false` to ignore. Matrix has fewer than n-1=$(n - 1) nonzero eigenvalues so square root may be inaccurate or matrix may not have a square root."))
```

src/triangular.jl
```diff
@@ -2702,6 +2705,19 @@ function sqrt_quasitriu(A0; blockwidth = eltype(A0) <: Complex ? 512 : 256)
         R = zeros(Tc, size(A0))
     end
     _sqrt_quasitriu!(R, A; blockwidth=blockwidth, n=n)
+
+    # check that the algorithm worked
+    if check
+        atol = eps(generic_normInf(evals)) # should work for any numeric data type
```
stevengj: Maybe we should use
```suggestion
        atol = eps(generic_normInf(evals)) * size(A, 1) # should work for any numeric data type
```
to match the default `rtol` for zero eigenvalues in symmetric.jl

src/dense.jl
stevengj: Maybe the `SelfAdjoint` case should take a `check` argument just for consistency, and ignore it?  (Zero eigenvalues are fine in that case.)

test/dense.jl
```diff
@@ -1013,6 +1013,11 @@ end
     @test sqrt(x6)^2 ≈ x6
 end
 
+@testset "issue #1548" begin
+    A = [0 1; 0 0]
+    @test_throws ArgumentError sqrt(A)
```
stevengj: Might be good to also add a couple tests with zero eigenvalues that *should* work:
```jl
@test iszero(sqrt(zeros(5,5)))
A = let X = rand(5,5); X * Diagonal([0,0,1,2,3]) / X; end
@test sqrt(A)^2 ≈ A
```

src/triangular.jl
trevorgloe: Why does symmetric.jl use this tolerance? Wouldn't this be misleadingly large for very sparse matrices?

stevengj:
> Will it always be guaranteed that `generic_normInf(evals)` will be a float type 

You can always call `float(...)` on the result to be sure.  But I don't think this is needed — [it already calls `float`](https://github.com/JuliaLang/LinearAlgebra.jl/blob/19ae2d7722a8d9a411007bf5150a8aaa28e60592/src/generic.jl#L566), and this is an undocumented internal function that shouldn't be overridden by external packages.

src/dense.jl
trevorgloe: I agree. I think it makes sense for the same function call to work whether its a dense matrix or symmetric matrix.

src/triangular.jl
stevengj: I think the general principle here is that the error in the eigenvalues (and Schur factorization) etcetera grows with the size of the matrix, due to roundoff accumulation.   Similar tolerances are used in things like the pseudo-inverse.

I'm not sure what sparsity has to do with it?

src/symmetric.jl
```diff
@@ -989,7 +989,8 @@ function log(A::SelfAdjoint)
 end
 
 # sqrt has rtol kwarg to handle matrices that are semidefinite up to roundoff errors
-function sqrt(A::SelfAdjoint; rtol = eps(real(float(eltype(A)))) * size(A, 1))
+# note that `check` is passed to the function for consistancy with the dense sqrt function, but it is not used as the matrix should always at least have a complex square root (just from eigendecomposition)
```
```suggestion
# note a `check` keyword argument is accepted for consistency with the generic `sqrt` function,
# but it is not used here, as a self-adjoint matrix is diagonalizable and hence always has a sqrt.
```

src/triangular.jl
```diff
@@ -2650,8 +2650,9 @@ end
 
 # End of auxiliary functions for matrix logarithm and matrix power
 
-sqrt(A::UpperTriangular) = sqrt_quasitriu(A)
-function sqrt(A::UnitUpperTriangular{T}) where T
+sqrt(A::UpperTriangular; check=true) = sqrt_quasitriu(A, diagview(A), check=check) # matrix is upper triangular, so eigenvalues are just the diagonals
```
```suggestion
sqrt(A::UpperTriangular; check=true) = sqrt_quasitriu(A, diagview(A); check) # matrix is upper triangular, so eigenvalues are just the diagonals
```
You can just pass keywords through by putting them with no equals sign after a semicolon, and it is nice to use this abbreviated syntax here and elsewhere.   (Especially if later on we also pass through an `rtol` keyword too.)

src/symmetric.jl
```diff
@@ -989,7 +989,8 @@ function log(A::SelfAdjoint)
...
+function sqrt(A::SelfAdjoint; check=true, rtol = eps(real(float(eltype(A)))) * size(A, 1))
```
```suggestion
function sqrt(A::SelfAdjoint; check::Bool=true, rtol = eps(real(float(eltype(A)))) * size(A, 1))
```

src/triangular.jl
```diff
@@ -2650,8 +2650,9 @@ end
...
+# shouldn't need to do a check for UnitUpperTriangular because the eigenvalues are all 1, flag included so the function call lines up
+function sqrt(A::UnitUpperTriangular{T}; check=true) where T
```
stevengj, line 2655:
```suggestion
function sqrt(A::UnitUpperTriangular{T}; check::Bool=true) where T 
```
I think we should declare `check` as `Bool` everywhere since it must be `true` or `false`.

At least in user-visible methods; the type can be omitted in internal methods, I guess.

src/triangular.jl
trevorgloe: I was thinking of something like a small matrix padded with a lot of zeros, in which case I would think that the number of zeros you pad with should not increase the tolerance. But I suppose those zeros could still contribute to the roundoff error unless is stored in a sparse format, in which case it probably wouldn't call this method anyway.

src/triangular.jl
trevorgloe: I'll add it to all the methods. It certainly shouldn't hurt to have it for internal methods.

src/triangular.jl
trevorgloe: Gotcha. I updated all the cases I could find in the PR to be like this. I didn't know of this way to pass keywords. Its quite clean.

src/triangular.jl
stevengj: Defaults are hard, but generically I would want the default tolerance to grow with the matrix size.

After this PR lands, I'd like to update the general `sqrt_quasitriu` algorithm to accept an `rtol` parameter, similar to the self-adjoint case, so that it can return a real square root for matrices with slightly negative eigenvalues due to roundoff errors.  At that point we can use the same `rtol` for the `check=true` eigenvalue test, and the user will be able to adjust it as needed.

src/dense.jl
```diff
@@ -1016,6 +1016,8 @@ and then the complex square root of the triangular factor.
 If a real square root exists, then an extension of this method [^H87] that computes the real
 Schur form and then the real square root of the quasi-triangular factor is instead used.
 
+When an $n \times n$ matrix has fewer than $n-1$ nonzero eigenvalues, the square root may not exist. In this case, and when the `check` flag is true, the algorithm will verify `X^2≈A` and throw an error if not.
```
```suggestion
When an $n \times n$ non-Hermitian matrix has fewer than $n-1$ nonzero eigenvalues, the square root may not exist. In this case, and when the `check` flag is true, the algorithm will verify `X^2≈A` and throw an error if not. 
```

src/dense.jl
```diff
@@ -1053,33 +1055,33 @@ function sqrt(A::AbstractMatrix{T}) where {T<:Union{Real,Complex}}
...
         if istriu(SchurF.T)
-            sqrtA = SchurF.Z * sqrt(UpperTriangular(SchurF.T)) * SchurF.Z'
+            sqrtA = SchurF.Z * sqrt(UpperTriangular(SchurF.T); check) * SchurF.Z'
         else
             # real sqrt exists whenever no eigenvalues are negative
             is_sqrt_real = !any(x -> isreal(x) && real(x) < 0, SchurF.values)
             # sqrt_quasitriu uses LAPACK functions for non-triu inputs
             if typeof(sqrt(zero(T))) <: BlasFloat && is_sqrt_real
-                sqrtA = SchurF.Z * sqrt_quasitriu(SchurF.T) * SchurF.Z'
+                sqrtA = SchurF.Z * sqrt_quasitriu(SchurF.T, SchurF.values, check=check) * SchurF.Z'
```
```suggestion
                sqrtA = SchurF.Z * sqrt_quasitriu(SchurF.T, SchurF.values; check) * SchurF.Z'
```

src/dense.jl
```diff
@@ -1053,33 +1055,33 @@ function sqrt(A::AbstractMatrix{T}) where {T<:Union{Real,Complex}}
             return applydiagonal(sqrt, A)
         end
     elseif ishermitian(A)
-        return _safe_parent(sqrt(Hermitian(A)))
+        return _safe_parent(sqrt(Hermitian(A))) # dont need to check for hermitian matrices
```
stevengj: Now that the self-adjoint case accepts a `check` keyword, we might as well pass it through:
```suggestion
        return _safe_parent(sqrt(Hermitian(A); check))
```

src/triangular.jl
```diff
@@ -2650,8 +2650,9 @@ end
 
 # End of auxiliary functions for matrix logarithm and matrix power
 
-sqrt(A::UpperTriangular) = sqrt_quasitriu(A)
-function sqrt(A::UnitUpperTriangular{T}) where T
+sqrt(A::UpperTriangular; check::Bool=true) = sqrt_quasitriu(A, diagview(A); check=check) # matrix is upper triangular, so eigenvalues are just the diagonals
```
```suggestion
sqrt(A::UpperTriangular; check::Bool=true) = sqrt_quasitriu(A, diagview(A); check) # matrix is upper triangular, so eigenvalues are just the diagonals
```

stevengj (approved):

src/dense.jl
```diff
@@ -1016,6 +1016,8 @@ and then the complex square root of the triangular factor.
 If a real square root exists, then an extension of this method [^H87] that computes the real
 Schur form and then the real square root of the quasi-triangular factor is instead used.
 
+When an $n \times n$ non-Hermitian matrix has fewer than $n-1$ nonzero eigenvalues, the square root may not exist. In this case, and when the `check` flag is true, the algorithm will verify `X^2≈A` and throw an error if not.
```
```suggestion
When an ``n × n`` non-Hermitian matrix has fewer than ``n - 1`` nonzero eigenvalues, the square root may not exist. In this case, and when the `check` flag is true, the algorithm will verify `X^2≈A` and throw an error if not. 
```

stevengj:
LGTM.  @araujoms?

src/dense.jl
```diff
@@ -1016,6 +1016,8 @@ and then the complex square root of the triangular factor.
 If a real square root exists, then an extension of this method [^H87] that computes the real
 Schur form and then the real square root of the quasi-triangular factor is instead used.
 
+When an ``n × n`` non-Hermitian matrix has fewer than ``n - 1`` nonzero eigenvalues, the square root may not exist. In this case, and when the `check` flag is true, the algorithm will verify `X^2≈A` and throw an error if not.
```
araujoms:
```suggestion
When a non-Hermitian matrix has two or more null eigenvalues, the square root may not exist. In this case, and when the `check` flag is true, the algorithm will verify `X^2≈A` and throw an error if not. 
```

src/triangular.jl
```diff
@@ -2702,6 +2705,19 @@ function sqrt_quasitriu(A0; blockwidth = eltype(A0) <: Complex ? 512 : 256)
...
     _sqrt_quasitriu!(R, A; blockwidth=blockwidth, n=n)
+
+    # check that the algorithm worked
+    if check
+        atol = eps(generic_normInf(evals)) * size(A, 1) # should work for any numeric data type
+        zero_eig = count(x -> abs(x) <= atol, evals) # count eigenvalues ≈ 0
+        if (zero_eig > 1) # in the regime where the algorithm could fail
+            test = generic_normInf(R*R .-= A0) <= eps(typeof(atol))^(1//4) * generic_normInf(A0)
+            if !test
+                throw(ArgumentError("Failed to produce matrix with X^2≈A. Pass `check=false` to ignore. Matrix has fewer than n-1=$(n - 1) nonzero eigenvalues so square root may be inaccurate or matrix may not have a square root."))
```
araujoms:
```suggestion
                throw(ArgumentError("Failed to produce matrix with X^2≈A. Pass `check=false` to ignore. Matrix has two or more null eigenvalues so square root may be inaccurate or matrix may not have a square root."))
```

araujoms (commented):
I think talking about the zero eigenvalues leads to more straightforward language than talking about nonzero eigenvalues.

Also, is `sqrt_quasitriu` ever called on a matrix where the eigenvalues are not simply the diagonals? If not then the extra argument is superfluous.

trevorgloe:
> Also, is `sqrt_quasitriu` ever called on a matrix where the eigenvalues are not simply the diagonals? If not then the extra argument is superfluous.

I believe `sqrt_quasitriu` will sometimes be called on a quasitriangular matrix, if its a real matrix with complex conjugate eigenvalues. For example if you do `schur(U)` where 
```julia
theta = 0.2
U = [cos(theta) -sin(theta); sin(theta) cos(theta)]
```
The `T` factor will be 
```
[0.980067 -0.198669
0.198669 0.980067]
```
(which is actually just `U`).

araujoms:
I see, nevermind then.

src/dense.jl
stevengj: I don't think we should use the word "null" as a synonym for "zero" here.  "Null" has too many other meanings in computer science.   Why not just "zero"?

araujoms: It's because the first number that appears in the sentence, two, is counting something, so the reader expects the next one to also do so, but it's not a count, it's a value. Using null breaks the pattern and lessens confusion.

It's a subtle effect, though, zero is fine.

src/triangular.jl
```diff
@@ -2668,14 +2669,16 @@ function sqrt(A::UnitUpperTriangular{T}) where T
...
-sqrt(A::UnitLowerTriangular) = copy(transpose(sqrt(copy(transpose(A)))))
+sqrt(A::LowerTriangular; check::Bool=true) = copy(transpose(sqrt(copy(transpose(A)); check)))
+sqrt(A::UnitLowerTriangular; check::Bool=true) = copy(transpose(sqrt(copy(transpose(A)); check)))
 
 # Auxiliary functions for matrix square root
 
 # square root of upper triangular or real upper quasitriangular matrix
-function sqrt_quasitriu(A0; blockwidth = eltype(A0) <: Complex ? 512 : 256)
+# A0 is triangular or quasitriangular matrix, evals is the eigenvalues
+function sqrt_quasitriu(A0, evals::AbstractVector; blockwidth = eltype(A0) <: Complex ? 512 : 256, check::Bool=true)
```
aravindh-krishnamoorthy, line 2679: Wouldn't it be better to perform/restrict the check to the public function `sqrt`, which already has access to `evals` from Schur decomposition?

`sqrt_quasitriu` is a private function (which is unlikely to be called directly by non-experts). Also, the check seems to completely depend on `evals` and not on any internal/intermediate value in the `sqrt_quasitru` function.

src/dense.jl
```diff
@@ -1053,33 +1055,33 @@ function sqrt(A::AbstractMatrix{T}) where {T<:Union{Real,Complex}}
...
         SchurF = schur(A)
-        return SchurF.vectors * sqrt(UpperTriangular(SchurF.T)) * SchurF.vectors'
+        return SchurF.vectors * sqrt(UpperTriangular(SchurF.T); check) * SchurF.vectors'
     end
 end
 
-sqrt(A::AdjointAbsMat) = adjoint(sqrt(parent(A)))
-sqrt(A::TransposeAbsMat) = transpose(sqrt(parent(A)))
+sqrt(A::AdjointAbsMat; check::Bool=true) = adjoint(sqrt(parent(A); check))
+sqrt(A::TransposeAbsMat, check::Bool=true) = transpose(sqrt(parent(A); check))
```
araujoms:
```suggestion
sqrt(A::TransposeAbsMat; check::Bool=true) = transpose(sqrt(parent(A); check))
```
Should be a keyword argument.

src/triangular.jl
araujoms: The problem is that you would need to put the check in two places, in the general `sqrt` and in `sqrt(::UpperTriangular)`.

stevengj: There are multiple `sqrt` functions where you'd have to put the check, currently.  Currently at least two, `sqrt(::AbstractMatrix)` and `sqrt(::UpperTriangular)`.  In the future we might want to add `sqrt(::Schur)` as well, and perhaps others.

So in this case it makes more sense to me to put the check once, in the lower-level function that all of these higher-level functions have to call.

src/dense.jl
stevengj: I don't have strong feelings either way.

src/triangular.jl
aravindh-krishnamoorthy: I see. Thanks!

araujoms (approved):

src/triangular.jl
```diff
@@ -2702,6 +2705,19 @@ function sqrt_quasitriu(A0; blockwidth = eltype(A0) <: Complex ? 512 : 256)
...
     _sqrt_quasitriu!(R, A; blockwidth=blockwidth, n=n)
+
+    # check that the algorithm worked
+    if check
+        atol = eps(generic_normInf(evals)) * size(A, 1) # should work for any numeric data type
+        zero_eig = count(x -> abs(x) <= atol, evals) # count eigenvalues ≈ 0
+        if (zero_eig > 1) # in the regime where the algorithm could fail
+            test = generic_normInf(R*R .-= A0) <= eps(typeof(atol))^(1//4) * generic_normInf(A0)
+            if !test
+                throw(ArgumentError("Failed to produce matrix with X^2≈A. Pass `check=false` to ignore. Matrix has two or more null eigenvalues so square root may be inaccurate or matrix may not have a square root."))
```
stevengj: Should this be a `DomainError` instead?

araujoms: I think so, the argument is valid, it's just outside the domain of the sqrt function.

stevengj:
```suggestion
                throw(DomainError("Failed to produce matrix with X^2≈A. Pass `check=false` to ignore. Matrix has two or more null eigenvalues so square root may be inaccurate or matrix may not have a square root."))
```

test/dense.jl
```diff
@@ -1013,6 +1013,21 @@ end
     @test sqrt(x6)^2 ≈ x6
 end
 
+@testset "issue #1548" begin
+    # check that error is thrown when matrix has no square root
+    A = [0 1; 0 0]
+    @test_throws ArgumentError sqrt(A)
```
stevengj:
```suggestion
    @test_throws DomainError sqrt(A)
```

stevengj:
Should be good to merge?

PR #1559, opened 2026-03-05
use muladd in generic_syrk

araujoms:
Gives a nice speedup for generic types which support it, while not affecting those which don't.


stevengj (approved):

araujoms:
@dkarrasch Would you mind if I backport this to 1.13? It's a stretch to call it a bugfix, but it's such a tiny change that I don't think it could possibly cause any trouble.

ViralBShah:
I think it is fine to backport.

PR #1561, opened 2026-03-13
Fix promotion in triangular linalg with mixed BLASFloats

### LINKED ISSUE #1560: performance of \(::UpperOrLowerTriangular{Float64}, ::Matrix{ComplexF64}) regression since v1.10
MWE:

```julia
using Random
using LinearAlgebra
Random.seed!(1234)
G = rand(2000, 2000);
Q = rand(2000, 2000);
Z = G + im * Q;
function f(A, b)
    
    F = lu(A; check=false)

    L, U, P = F
    y = L \ (b[P, :])
    return U \ y
end

@time f(G, Z);
```

In Julia 1.10

```julia
julia> @time f(G, Z);
  8.785467 seconds (13.07 k allocations: 275.582 MiB, 0.35% gc time, 0.25% compilation time)
```

In Julia 1.9

```julia
julia> @time f(G, Z);
  0.711327 seconds (18 allocations: 396.760 MiB, 1.92% gc time)
```

tracking to this fallback implement:
https://github.com/JuliaLang/julia/blob/95f30e51f4158dd3421cd5b8098849d24e97e497/stdlib/LinearAlgebra/src/triangular.jl#L1124

the LAPACK routine is not hit because `eltype(C) !== eltype(A)`
https://github.com/JuliaLang/julia/blob/95f30e51f4158dd3421cd5b8098849d24e97e497/stdlib/LinearAlgebra/src/triangular.jl#L830

dkarrasch:
To have triangular linalg be capable of unitful elements, we don't want/can't promote all factors. In the mixed BLAS-case, however, no-promotion prevents the BLAS pathway. I believe this was introduced by me in https://github.com/JuliaLang/julia/pull/50058. :(

Fixes #1560.


stevengj (approved):

dkarrasch:
Introducing methods for `BlasFloat`s elements (like in the first commit) required many methods for disambiguation, so I went with a different solution: if the target type is `BlasFloat`, it promotes both factors and performs the mul/solve in BLAS. This would include cases like `Int` times `BlasFloat`. Is that too much? Should we allow the BLAS route only if both factors are themselves of BlasFloat eltype?

dkarrasch:
Actually, I think this was the behavior that we used to have pre v1.10. Out of place mul/solve simply promoted both "factors" to the target type, so with one `Float64` matrix and another `Int`, for instance, the result was a `Float64` matrix, computed via the BLAS path.

@stevengj Any new thoughts after the rework after your approval?

PR #1562, opened 2026-03-16
Generalize zero in Givens algorithm

BambOoxX:
This PR fixes errors for eigen and singular value decompositions with [`Measurmements.jl`](https://github.com/JuliaPhysics/Measurements.jl) numbers.
See https://github.com/JuliaLinearAlgebra/GenericLinearAlgebra.jl/issues/163 and https://github.com/JuliaPhysics/Measurements.jl/issues/189 for a detailed history.

PR #1568, opened 2026-04-01
resolve test ambiguities with Base

araujoms:
x-ref: https://github.com/JuliaLang/LinearAlgebra.jl/pull/1562#issuecomment-4167844390

### DIFF
--- a/test/givens.jl
+++ b/test/givens.jl
@@ -107,8 +107,16 @@ end
 struct MockMeasurement{T<:AbstractFloat} <: AbstractFloat
     data::T
 end
+#these methods are only needed for preventing ambiguities
+MockMeasurement{T}(z::Complex) where {T<:AbstractFloat} =
+    isreal(z) ? MockMeasurement{T}(real(z))::MockMeasurement{T} :
+        throw(InexactError(nameof(MockMeasurement{T}), MockMeasurement{T}, z))
+MockMeasurement{T}(r::Rational{P}) where {P,T<:AbstractFloat} = MockMeasurement{T}(T(r))
+MockMeasurement{T}(x::MockMeasurement{T}) where {T<:AbstractFloat} = x
+MockMeasurement{T}(c::AbstractChar) where {T<:AbstractFloat} = error("undefined")
+MockMeasurement{T}(x::Base.TwicePrecision) where {T<:AbstractFloat} = error("undefined")
+#these methods are actually needed
 import Base: promote_rule, floatmin, eps, ==, <, -, +, <=, sqrt
-MockMeasurement(x::Real) = MockMeasurement(float(x))
 one(::Type{<:MockMeasurement{T}}) where {T<:Real} = one(T)
 promote_rule(::Type{MockMeasurement{T}}, ::Type{<:Real}) where {T} = MockMeasurement{T}
 for f in (:floatmin, :eps, :oneunit)
--- a/test/givens.jl
+++ b/test/givens.jl
@@ -114,7 +114,7 @@ MockMeasurement{T}(z::Complex) where {T<:AbstractFloat} =
 MockMeasurement{T}(r::Rational{P}) where {P,T<:AbstractFloat} = MockMeasurement{T}(T(r))
 MockMeasurement{T}(x::MockMeasurement{T}) where {T<:AbstractFloat} = x
 MockMeasurement{T}(c::AbstractChar) where {T<:AbstractFloat} = error("undefined")
-MockMeasurement{T}(x::Base.TwicePrecision) where {T<:AbstractFloat} = error("undefined")
+MockMeasurement{T}(x::Base.TwicePrecision) where {T<:AbstractFloat} = MockMeasurement(T(x))
 #these methods are actually needed
 import Base: promote_rule, floatmin, eps, ==, <, -, +, <=, sqrt
 one(::Type{<:MockMeasurement{T}}) where {T<:Real} = one(T)
--- a/test/givens.jl
+++ b/test/givens.jl
@@ -108,12 +108,10 @@ struct MockMeasurement{T<:AbstractFloat} <: AbstractFloat
     data::T
 end
 #these methods are only needed for preventing ambiguities
-MockMeasurement{T}(z::Complex) where {T<:AbstractFloat} =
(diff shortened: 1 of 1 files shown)


test/givens.jl
```diff
@@ -107,8 +107,16 @@ end
 struct MockMeasurement{T<:AbstractFloat} <: AbstractFloat
     data::T
 end
+#these methods are only needed for preventing ambiguities
+MockMeasurement{T}(z::Complex) where {T<:AbstractFloat} =
+    isreal(z) ? MockMeasurement{T}(real(z))::MockMeasurement{T} :
+        throw(InexactError(nameof(MockMeasurement{T}), MockMeasurement{T}, z))
+MockMeasurement{T}(r::Rational{P}) where {P,T<:AbstractFloat} = MockMeasurement{T}(T(r))
+MockMeasurement{T}(x::MockMeasurement{T}) where {T<:AbstractFloat} = x
+MockMeasurement{T}(c::AbstractChar) where {T<:AbstractFloat} = error("undefined")
```
stevengj: Why is an `AbstractChar` method needed?  That's not a type of number.

test/givens.jl
```diff
@@ -107,8 +107,16 @@ end
...
+MockMeasurement{T}(x::Base.TwicePrecision) where {T<:AbstractFloat} = error("undefined")
```
stevengj: Why wouldn't this do a lossy conversion to `T`?

test/givens.jl
```diff
@@ -107,8 +107,16 @@ end
 struct MockMeasurement{T<:AbstractFloat} <: AbstractFloat
     data::T
 end
+#these methods are only needed for preventing ambiguities
+MockMeasurement{T}(z::Complex) where {T<:AbstractFloat} =
+    isreal(z) ? MockMeasurement{T}(real(z))::MockMeasurement{T} :
+        throw(InexactError(nameof(MockMeasurement{T}), MockMeasurement{T}, z))
```
stevengj: Can't it just do `MockMeasurement(T(z))`, since `T(z)` already checks whether `z` is real?

test/givens.jl
stevengj: Oh, I see, there is a `(::Type{T})(x::AbstractChar) where T<:Union{AbstractChar, Number}` constructor defined.

stevengj: But in that case it should be 
```
MockMeasurement{T}(c::AbstractChar) where {T<:AbstractFloat} = MockMeasurement(T(c))
```
for consistency, no?

test/givens.jl
stevengj: i.e.
```suggestion
MockMeasurement{T}(x::Base.TwicePrecision) where {T<:AbstractFloat} = MockMeasurement(T(x))
```

stevengj:
(Aside: it really bugs me that we have to define the method as `MockMeasurement{T}(c::AbstractChar) where {T<:AbstractFloat}` instead of `MockMeasurement{T}(c::AbstractChar) where {T}`, since the `<:AbstractFloat` should be implied by the `MockMeasurement` type, but it seems that Julia's dispatch system can't figure that out.)

test/givens.jl
araujoms: I don't think it matters, because this method will never be called.

test/givens.jl
araujoms: Sure.

test/givens.jl
araujoms: Indeed.

araujoms:
I had indeed first defined the methods as `MockMeasurement{T}(c::AbstractChar) where {T}`, and couldn't understand why it didn't work. Well, nothing we can do about that here.

DilumAluthge:
@stevengj @dkarrasch Would this fix the CI failures in https://github.com/JuliaLang/julia/pull/61450?

test/givens.jl
```diff
@@ -107,8 +107,14 @@ end
 struct MockMeasurement{T<:AbstractFloat} <: AbstractFloat
     data::T
 end
+#these methods are only needed for preventing ambiguities
+MockMeasurement{T}(x::MockMeasurement{T}) where {T<:AbstractFloat} = x
+MockMeasurement{T}(z::Complex) where {T<:AbstractFloat} = MockMeasurement(T(z))
+MockMeasurement{T}(r::Rational{P}) where {P,T<:AbstractFloat} = MockMeasurement(T(r))
```
dkarrasch, line 113: Seems like this should be enough?

```suggestion
MockMeasurement{T}(r::Rational) where {T<:AbstractFloat} = MockMeasurement(T(r))
```

dkarrasch (commented):
I assume we need to disambiguate some of the following methods:

```julia
julia> methods(AbstractFloat, (Any,))
# 19 methods for type constructor:
[...]
 [15] (::Type{T})(x::Rational{S}) where {S, T<:AbstractFloat}
     @ rational.jl:162
 [16] (::Type{T})(z::Complex) where T<:Real
     @ complex.jl:44
 [17] (::Type{T})(x::Base.TwicePrecision) where T<:Number
     @ twiceprecision.jl:265
 [18] (::Type{T})(x::T) where T<:Number
     @ boot.jl:965
 [19] (::Type{T})(x::AbstractChar) where T<:Union{AbstractChar, Number}
     @ char.jl:52
```

So, this looks good to me, except for one comment for which I don't think the type parameter makes a difference anyway.

test/givens.jl
araujoms: I also thought it wouldn't make a difference, but I tested it, and it does.

dkarrasch: Interesting. I would have thought this is even "bad style" in the spirit of `Aqua.jl`. Was it "unbound parameter" or something like this? If anyone bothers to explain, I'd be happy to learn.

araujoms: I don't think `Aqua` complains about this, because the code is meaningful: `foo(z::Complex{T}) where {T}` is a more specific method than `foo(z::Complex)`, and will be the one called if we declare both.

Therefore it makes sense that `MockMeasurement{T}(r::Rational) where {T<:AbstractFloat}` doesn't help. It is a less specific method, and therefore doesn't help disambiguating `(::Type{T})(x::Rational{S}) where {S, T<:AbstractFloat}`.

dkarrasch (approved):

PR #1570, opened 2026-04-03
Deprecate implicit vector pinvs

antoine-levitt:
Try to reland https://github.com/JuliaLang/julia/pull/50955


dkarrasch (commented):
The `@deprecate` macro has a different syntax. It is usually used to deprecate one `old_name` function name in favor of a `new_name`. I'm not sure if one could "abuse" this here like:

```julia
@deprecate (\)(a::AbstractVector, b::AbstractArray) \(a, b)
```

Here, we are not even planning to rename the function call.

antoine-levitt:
So there's no mechanism to say "we want to remove this method completely "?

DilumAluthge:
You maybe could call `Base.depwarn` directly, with a custom message?

PR #1572, opened 2026-04-12
Simplify `matprod_dest` methods

dkarrasch:
The deleted methods complicate the extension of the `matprod_dest` approach in other packages such as `SparseArrays.jl`. Since the fallback uses 3-arg `similar`, I don't think we need special methods to handle `StructuredMatrix`es.


ViralBShah:
I am not sure I will be able to meaningfully review the change. It seems right directionally, and we can adapt to it in SparseArrays.jl for sure. I am not sure about the impact elsewhere.

src/LinearAlgebra.jl
```diff
@@ -670,17 +670,15 @@ _pushzero(A) = (B = similar(A, length(A)+1); @inbounds B[begin:end-1] .= A; @inb
 _droplast!(A) = deleteat!(A, lastindex(A))
 
 # destination type for matmul
-matprod_dest(A::StructuredMatrix, B::StructuredMatrix, TS) = similar(B, TS, size(B))
-matprod_dest(A, B::StructuredMatrix, TS) = similar(A, TS, size(A))
```
dkarrasch, line 674: For reviewers: the default behavior is to call 3-arg `similar` for the second array argument. This PR boils down to whether we believe this switch here is to ever make a difference. `StructuredMatrix` is a union over the banded matrix types in LinearAlgebra plus the triangular and `UpperHessenberg` wrapper types.

araujoms (approved):
I don't think it's at all controversial. The `similar` fallback works fine, so it's better to not have specific methods for `StructuredMatrix`.

I mean, the change here is mostly innocuous, the real point is whatever you'll do with it in `SparseArrays`.

dkarrasch:
It's the one line commented on above that technically might change behavior, because it replaces the `similar(A,...)` call by the fallback, which calls `similar(B, ...)`. But in the realm of LinearAlgebra.jl, this will almost always yield plain `Matrix`. So, then let's go with it.

The intended change over at SparseArrays.jl is to return dense arrays for products of dense and (wrapped versions of) sparse arrays. That should speed up mixed products significantly, because writing into the result array is much, much faster.

PR #1573, opened 2026-04-13
Add sizehint to syevr! to avoid potentially unnecessary allocation

### LINKED ISSUE #1425: Symmetric eigen memory usage for eigenvectors.
Commit 57e9a0d to reduce allocations in `syevr!` seems to have a potential negative impact on the method
```
eigen(A::RealHermSymComplexHerm, vl::Real, vh::Real)
```
when the interval is such that a small number of eigenvalues/vectors are returned.  The resizing and reshaping of the storage passed to LAPACK in `syevr!` means that space for an $n\times n$ eigenvector matrix is retained even if only a few eigenvectors are returned.  This can be painful for large matrices.

With 1.12.0-rc1 the following code illustrates this:
```
using LinearAlgebra

n = 2000
A = Hermitian(randn(n,n))
e, V = eigen(A)
println("All eigenvectors:")
@show size(V)
@show Base.summarysize(V)

println("\nEigenvectors in [-1,1]:")
e1, V1 = eigen(A, -1, 1)
@show size(V1)
@show Base.summarysize(V1)

println("\nEigenvectors in [-1,1], copied:")
V1 = copy(V1)
@show size(V1)
@show Base.summarysize(V1)
```

The output is:
```
All eigenvectors:
size(V) = (2000, 2000)
Base.summarysize(V) = 32000048

Eigenvectors in [-1,1]:
size(V1) = (2000, 30)
Base.summarysize(V1) = 32000048

Eigenvectors in [-1,1], copied:
size(V1) = (2000, 30)
Base.summarysize(V1) = 480048
```

trevorgloe:
Fixes #1425 

Adding `sizehint!` into the reshape command in syevr! seems to avoid a potentially large allocation that happens temporarily in `eigen` when a range is given and only a small subset of eigenvectors are returned. It seems that the actual array of eigenvectors returned is not taking up the whole size that `summarysize` suggests but the `MemoryRef` is not updated correctly. This fix ensures that `summarysize` is actually reflecting the correct size of the array returned. From the tests, this doesn't seem to have any adverse effects otherwise.

### DIFF
--- a/src/lapack.jl
+++ b/src/lapack.jl
@@ -5429,7 +5429,7 @@ for (syev, syevr, syevd, sygvd, elty) in
                 end
             end
             zm = jobz == 'V' ? m[] : 0
-            resize!(W, m[]), reshape(resize!(Z, ldz * zm), ldz, zm)
+            resize!(W, m[]), reshape(sizehint!(resize!(Z, ldz * zm), ldz * zm), ldz, zm)
         end
         syevr!(jobz::AbstractChar, A::AbstractMatrix{$elty}) =
             syevr!(jobz, 'A', 'U', A, 0.0, 0.0, 0, 0, -1.0)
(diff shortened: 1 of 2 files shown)


ViralBShah:
Test failure appears related.

trevorgloe:
Oops! In hindsight it was probably a bad idea to assume that `summarysize` would return 96 for a 3x2 matrix of `Float64` on every machine (thats what it returns on my machine). I think the new test should capture that bad behavior, part of which is that the size will look different once you copy the array, which I believe is an artifact of the fact that the memoryRef is not updated, even though the array is.

Issue #855, opened 2021-05-27
qr factorisation method for nullspace


gszep:
I've opened this feature request based off the back of a thread on slack with @andreasnoack . Since QR is not as reliable as SVD in determining the rank, we could have an additional kwarg where the user provides an expected dimension of the nullspace. Roughly
```julia
function nullspace(A::AbstractMatrix; nullity::Int=0, kwargs...)

    if iszero(nullity) # default to SVD
        return nullspace(A;kwargs...)
   
    else 
        return qr(A').Q[:,end+1-nullity:end]
    end
end
```
A small test
```julia
A = randn(3)randn(3)'
nullity = 2 # expected nullity

A*nullspace(A)
A*qr(A').Q[:,end+1-nullity:end]
```

andreasnoack:
Ref https://github.com/JuliaLang/LinearAlgebra.jl/issues/514

PR #1576, opened 2026-04-18
Add SVD algorithm kwarg to `nullspace`

### LINKED ISSUE #1571: Add SVD-algorithm kwarg to `nullspace`
Presently, `nullspace` is essentially a simple wrapper over `svd`, picking out the relevant rows of `Vt`:

https://github.com/JuliaLang/LinearAlgebra.jl/blob/17325b74ec807f5237228512db1131ddfe4048e3/src/dense.jl#L1881-L1888

By default, `svd` uses a divide-and-conquer algorithm (`LinearAlgebra.DivideAndConquer()` -> `gesdd`), but it would be nice to be able to choose other algorithms - namely the QR iteration algorithm (`LinearAlgebra.QRIteration()`  -> `gesvd`). 
The purpose of this issue is therefore to suggest that an `alg` or `svd_alg` argument is added to `nullspace`, which is simply forwarded to `svd`.

The motivation is that the divide-and-conquer approach sometimes fails to converge the SVD, especially for very large and/or nearly rank-deficient problems.

trevorgloe:
Fixes #1571

Now you can call `nullspace` with an `alg` argument, which is passed through to `svd`. 
```julia
A = [1 1 1 0; 1 1 1 0; 0 0 1 1; 0 1 0 1]
qr_null = nullspace(A, alg=LinearAlgebra.QRIteration())
div_null = nullspace(A, alg=LinearAlgebra.DivideAndConquer())
```

### DIFF
--- a/src/dense.jl
+++ b/src/dense.jl
@@ -1840,8 +1840,9 @@ end
 ## Basis for null space
 
 """
-    nullspace(M; atol::Real=0, rtol::Real=atol>0 ? 0 : n*ϵ, alg::Algorithm=default_svd_algorithm(A))
+    nullspace(M; atol::Real=0, rtol::Real=atol>0 ? 0 : n*ϵ)
     nullspace(M, rtol::Real) = nullspace(M; rtol=rtol) # to be deprecated in Julia 2.0
+    nullspace(S::SVD{<:Any, T}; atol::Real=0, rtol::Real=min(n,m)*ϵ) where {T}
 
 Computes a basis for the nullspace of `M` by including the singular
 vectors of `M` whose singular values have magnitudes smaller than `max(atol, rtol*σ₁)`,
@@ -1851,12 +1852,7 @@ By default, the relative tolerance `rtol` is `n*ϵ`, where `n`
 is the size of the smallest dimension of `M`, and `ϵ` is the [`eps`](@ref) of
 the element type of `M`.
 
-The desired algorithm, `alg` is passed through to `svd`. The available algorithms will
-be the same as that of ['svd'](@ref).
-
-!!! compat "Julia 1.3"
-    The `alg` keyword argument requires Julia 1.3 or later.
-
+If a specific svd algorithm is required, the function may be called passing an SVD object rather than a matrix. 
 
 # Examples
 ```jldoctest
@@ -1883,12 +1879,42 @@ julia> nullspace(M, atol=0.95)
  0.0
  0.0
  1.0
+
+julia> A = [1 1; 2 0; -1 1]*[1 2 3; 3 2 1]
+3×3 Matrix{Int64}:
+ 4  4   4
+ 2  4   6
+ 2  0  -2
+
+julia> F = svd(A)
+SVD{Float64, Float64, Matrix{Float64}, Vector{Float64}}
+U factor:
+3×3 Matrix{Float64}:
+ -0.673105    0.462165  -0.57735
+ -0.736799   -0.351843   0.57735
+  0.0636942   0.814008   0.57735
+singular values:
+3-element Vector{Float64}:
+ 10.027069108300799
+  3.384949792442987
+  5.768888059150692e-16
+Vt factor:
+3×3 Matrix{Float64}:
+ -0.402773  -0.562439  -0.722106
+  0.819212   0.130367  -0.558477
+ -0.408248   0.816497  -0.408248
+
+julia> nullspace(F)
+3×1 Matrix{Float64}:
+ -0.4082482904638629
+  0.816496580927726
+ -0.4082482904638631
 ```
 """
-function nullspace(A::AbstractVecOrMat; atol::Real=0, rtol::Real = (min(size(A, 1), size(A, 2))*eps(real(float(oneunit(eltype(A))))))*iszero(atol), alg::Algorithm = default_svd_alg(A))
+function nullspace(A::AbstractVecOrMat; atol::Real=0, rtol::Real = (min(size(A, 1), size(A, 2))*eps(real(float(oneunit(eltype(A))))))*iszero(atol))
     m, n = size(A, 1), size(A, 2)
     (m == 0 || n == 0) && return Matrix{eigtype(eltype(A))}(I, n, n)
-    SVD = svd(A; full=true, alg)
+    SVD = svd(A; full=true)
     tol = max(atol, SVD.S[1]*rtol)
     indstart = sum(s -> s .> tol, SVD.S) + 1
     return copy((@view SVD.Vt[indstart:end,:])')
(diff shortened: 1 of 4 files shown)


src/dense.jl
```diff
@@ -1851,6 +1851,9 @@ By default, the relative tolerance `rtol` is `n*ϵ`, where `n`
 is the size of the smallest dimension of `M`, and `ϵ` is the [`eps`](@ref) of
 the element type of `M`.
 
+The desired algorithm, `alg` is passed through to `svd`. The available algorithms will
+be the same as that of ['svd'](@ref).
+
```
dkarrasch, line 1854: I think we need a compat entry here, saying something like "the keyword argument `alg` requires Julia version 1.14" or something like that. There are examples in this or other files.

stevengj:
I can imagine an `alg` argument where it doesn't use SVD at all, but instead uses a pivoted QR factorization.  (See also #855 and #1052).

stevengj:
Wouldn't a cleaner and more powerful API be to just talk an existing `SVD` object as an argument to `nullspace`, instead of the matrix?

That way, the caller could construct the SVD however they wanted, and it would also enable an existing SVD to be re-used.

Code-wise, it would just require a bit of refactoring of the existing method.

trevorgloe:
> Wouldn't a cleaner and more powerful API be to just talk an existing `SVD` object as an argument to `nullspace`, instead of the matrix?
> 
> That way, the caller could construct the SVD however they wanted, and it would also enable an existing SVD to be re-used.
> 
> Code-wise, it would just require a bit of refactoring of the existing method.

I was actually thinking the same thing. It seems a bit inefficient to compute the whole SVD here if the user may already have one. I will try my hand at making two methods: one for taking in an actual matrix (at way something like a pivoted QR can be used rather than an SVD), and one taking in an SVD object. 
This would also keep the change from breaking existing uses of the function.

trevorgloe:
Ok, I've fixed the test that seemed to be giving a problem. The test was to find the nullspace of 
```julia
A1 = [1.0 1.0; 2.0 0.0; -1.0 1.0] * [1.0 1.0 1.0; 1.0 -1.0 1.0]
```
and compare it to `[-1/sqrt(2); 0; 1/sqrt(2)]`, which should be the nullspace (at least theoretically). For some reason this test would only fail on the macos-aarch64 test suite. Anybody know why this would happen?

stevengj:
> For some reason this test would only fail on the macos-aarch64 test suite. Anybody know why this would happen?

The nullspace basis is only defined up to an arbitrary rotation.   In this case, for a 1d nullspace, the sign is arbitrary, and on MacOS it flips the sign, and returns a vector `≈ [1/√2, 0, -1/√2]`

stevengj (changes requested):
Instead of passing through an `alg` parameter, just refactor to add a method where you can pass an `SVD` object directly.

trevorgloe:
> Instead of passing through an `alg` parameter, just refactor to add a method where you can pass an `SVD` object directly.

Sorry this is quite delayed...forgot about this one. I believe the requested change is in there. In line 642 onward in svd.jl theres a new method 
```julia
nullspace(S::SVD{<:Any, T}; atol::Real=0, rtol::Real=min(n,m)*ϵ) where {T}
```
It was easier to put in the svd.jl file because then the svd type is well defined right there. I think I was getting some issues having it in dense.jl.

src/svd.jl
```diff
@@ -639,3 +639,53 @@ AbstractMatrix(F::SVD) = (F.U * Diagonal(F.S)) * F.Vt
 AbstractArray(F::SVD) = AbstractMatrix(F)
 Matrix(F::SVD) = Array(AbstractArray(F))
 Array(F::SVD) = Matrix(F)
+
+"""
+    nullspace(S::SVD{<:Any, T}; atol::Real=0, rtol::Real=min(n,m)*ϵ) where {T}
```
stevengj: I don't think it needs a whole separate docstring.  The ordinary `nullspace` docstring can just add a line or two saying that you can pass an SVD object instead of a matrix.

src/dense.jl
```diff
@@ -1838,7 +1838,7 @@ end
 ## Basis for null space
 
 """
-    nullspace(M; atol::Real=0, rtol::Real=atol>0 ? 0 : n*ϵ)
+    nullspace(M; atol::Real=0, rtol::Real=atol>0 ? 0 : n*ϵ, alg::Algorithm=default_svd_algorithm(A))
```
stevengj: I think you can get rid of the `alg` keyword now that you can pass an `SVD` object directly.

src/dense.jl
```diff
@@ -1876,10 +1883,10 @@ julia> nullspace(M, atol=0.95)
...
 """
-function nullspace(A::AbstractVecOrMat; atol::Real=0, rtol::Real = (min(size(A, 1), size(A, 2))*eps(real(float(oneunit(eltype(A))))))*iszero(atol))
+function nullspace(A::AbstractVecOrMat; atol::Real=0, rtol::Real = (min(size(A, 1), size(A, 2))*eps(real(float(oneunit(eltype(A))))))*iszero(atol), alg::Algorithm = default_svd_alg(A))
     m, n = size(A, 1), size(A, 2)
     (m == 0 || n == 0) && return Matrix{eigtype(eltype(A))}(I, n, n)
-    SVD = svd(A; full=true)
+    SVD = svd(A; full=true, alg)
     tol = max(atol, SVD.S[1]*rtol)
     indstart = sum(s -> s .> tol, SVD.S) + 1
     return copy((@view SVD.Vt[indstart:end,:])')
```
stevengj: Now you can just call the `SVD` version, no?

```suggestion
return nullspace(SVD; rtol, atol)
```

src/svd.jl
```diff
@@ -639,3 +639,53 @@ AbstractMatrix(F::SVD) = (F.U * Diagonal(F.S)) * F.Vt
...
+3×1 Matrix{Float64}:
+ -0.4082482904638629
+  0.816496580927726
+ -0.4082482904638631
+```
+"""
+
+function nullspace(F::SVD; atol::Real=0, rtol::Real = (min(size(F.U, 1), size(F.V, 1))*eps(real(float(oneunit(eltype(F))))))*iszero(atol))
+    r = rank(F; atol, rtol)
+    indstart = r + 1 # nullspace starts after the numerical rank + 1
```
stevengj, line 646: Is there a reason why you changed the `indstart` algorithm to use `rank`?

trevorgloe: This was because the rank function was doing the exact same thing that I would've been doing counting the singular values above the tolerance. Rather than just writing the same check again, I figured it would be better to use the existing function in svd.jl

PR #1578, opened 2026-04-21
Improving linear solves: reduction of methods, destination array type mechanism

dkarrasch:
Division by a triangular matrix is like multiplying by a triangular matrix, so we should be using the `matprod_dest` mechanism throughout. (EDIT: While this is technically true, inverses of sparse triangular matrices may still be "dense", so multiplication and division with triangular matrices are still different).

This fixes the inconsistency that left-division of a sparse matrix by a triangular matrix yields a dense matrix (hard-coded in SparseArrays.jl), whereas right-division yields a sparse matrix (because it is using the fallback). EDIT: This inconsistency will be resolved once the corresponding `mat[l/r]div_dest` methods are defined in SparseArrays.jl.

### DIFF
--- a/src/diagonal.jl
+++ b/src/diagonal.jl
@@ -606,8 +606,10 @@ function (*)(Da::Diagonal, Db::Diagonal, Dc::Diagonal)
     return Diagonal(Da.diag .* Db.diag .* Dc.diag)
 end
 
-# flipped order in matprod_dest is on purpose, to put the vector argument on second place
-/(A::AbstractVecOrMat, D::Diagonal) = _rdiv!(matprod_dest(D, A, promote_op(/, eltype(A), eltype(D))), A, D)
+matrdiv_dest(A, D::Diagonal) = similar(A, promote_op(/, eltype(A), eltype(D)))
+matrdiv_dest(A::HermOrSym, D::Diagonal) = similar(A, promote_op(/, eltype(A), eltype(D)), size(A))
+
+/(A::AbstractVecOrMat, D::Diagonal) = _rdiv!(matrdiv_dest(A, D), A, D)
 
 rdiv!(A::AbstractVecOrMat, D::Diagonal) = @inline _rdiv!(A, A, D)
 # avoid copy when possible via internal 3-arg backend
@@ -628,7 +630,10 @@ function _rdiv!(B::AbstractVecOrMat, A::AbstractVecOrMat, D::Diagonal)
     B
 end
 
-\(D::Diagonal, B::AbstractVecOrMat) = ldiv!(matprod_dest(D, B, promote_op(\, eltype(D), eltype(B))), D, B)
+matldiv_dest(D::Diagonal, B) = similar(B, promote_op(\, eltype(D), eltype(B)))
+matldiv_dest(D::Diagonal, B::HermOrSym) = similar(B, promote_op(\, eltype(D), eltype(B)), size(B))
+
+\(D::Diagonal, B::AbstractVecOrMat) = ldiv!(matldiv_dest(D, B), D, B)
 
 ldiv!(D::Diagonal, B::AbstractVecOrMat) = @inline ldiv!(B, D, B)
 function ldiv!(B::AbstractVecOrMat, D::Diagonal, A::AbstractVecOrMat)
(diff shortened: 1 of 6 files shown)


dkarrasch:
I was in the flow, and added a second mechanism which greatly reduces the number of `/` and `\` methods:

```julia
methods(\, (AbstractMatrix, AbstractMatrix), LinearAlgebra)
# 16 methods for generic function "\" from Base:
# vs 37 methods in v1.12

methods(/, (AbstractMatrix, AbstractMatrix), LinearAlgebra)
# 34 methods for generic function "/" from Base:
# vs 58 methods in v1.12
```

and some `mul` from 43 to 34. All of the reduced ones existed only for a postprocessing reason, which I moved to a helper function which gets called from the fallback promotion method.

araujoms:
What is the goal here? Left and right division do preserve triangularity, but `LinearAlgebra` currently respects that.

Now they don't preserve sparseness. I think the output of both left and right division should be dense, is that what you intend?

Right now the PR shouldn't change anything for `LinearAlgebra` itself other than simplification, right?

timholy:
I haven't read it in detail, but presumably it's at least in part "ambiguity engineering": designing the code base to make it easier for packages to extend functions for custom types without triggering ambiguities. I've done this kind of work myself, and I can attest to the fact that this can be a very high-level and complex design task. Kudos to @dkarrasch for the enormous progress made over the years: it's the kind of progress that can in some circumstances be game-changing for package developers, yet on first blush it's easy to miss its significance.

Perhaps we should have an "ambiguities" label?

dkarrasch:
Thanks @timholy for the appreciation!

And yes, @araujoms, this PR went a little sideways to the original goal. I would like to achieve the following all at once:

* provide a mechanism that allows to preserve outer wrappers like the triangular ones,
* eltype determination (e.g. no division in the `Unit*Triangular` integer cases),
* smart choice of destination array type (this is where multiplication and solves may differ),
* without the need to overload top-level functions like `*` and `\`.

My test case is `SparseArrays.jl` as a downstream dependency. So, indeed, as for LinearAlgebra, this does not intend to change anything functionally.

araujoms:
That's a great initiative, should help reducing the loading time of `SparseArrays`. Also, I'm always a fan of PRs that have net negative LOC.

As far as I can tell `matprod_dest` only accepted a type as a third argument to handle the complications of `rdiv` and `ldiv`; since now this is offloaded to `matldiv_dest` and `matrdiv_dest` can't it be simplified? Perhaps it would make sense having a unified `matop_dest(op, A, B)` instead of all three functions, I'm afraid the interface is getting rather messy.

Also, are `matldiv_dest` and `matrdiv_dest` intended to become `public`?

dkarrasch:
One issue with a single `*_dest` function is that it inherits all the ambiguities. Like left-division by `UpperTriangular` is handled by `matldiv_dest(A, B::UpperTriangular)`, and right-division by `matrdiv_dest(A::UpperTriangular, B)`. If we had a single function, this required another method `matop_dest(A::UpperTriangular, B::UpperTriangular)`, which may not be too bad after all.

How about we merge this as is and explore merging all the `*_dest` functions in another PR? Depending on the outcome we would decide about whether/what to declare public.

dkarrasch:
I'm just a little afraid of "spoiling" this PR with experimental commits.

araujoms:
I don't see why there would be any ambiguity. To be clear, I'm proposing to replace `matldiv_dest(A, B::UpperTriangular)` with `matop_dest(\, A, B::UpperTriangular)` and `matrdiv_dest(A::UpperTriangular, B)` with `matop_dest(/, A::UpperTriangular, B)`.

src/diagonal.jl
```diff
@@ -717,14 +717,14 @@ function ldiv!(T::Tridiagonal, D::Diagonal, S::Union{SymTridiagonal,Tridiagonal}
...
 
-function (/)(S::SymTridiagonal, D::Diagonal)
+function matrdiv_dest(S::SymTridiagonal, D::Diagonal)
     T = promote_op(\, eltype(D), eltype(S))
     du = similar(S.ev, T, max(length(S.dv)-1, 0))
     d  = similar(S.dv, T, length(S.dv))
     dl = similar(S.ev, T, max(length(S.dv)-1, 0))
-    _rdiv!(Tridiagonal(dl, d, du), S, D)
+    return Tridiagonal(dl, d, du)
 end
```
araujoms:
```suggestion
matrdiv_dest(S::SymTridiagonal, D::Diagonal) = matldiv_dest(D, S)
```

dkarrasch: Nice catch!

dkarrasch: Hm, except for when this is unitful. But I see there has been a typo in the `T` computation.

dkarrasch:
> I'm proposing to replace `matldiv_dest(A, B::UpperTriangular)` with `matop_dest(\, A, B::UpperTriangular)` and `matrdiv_dest(A::UpperTriangular, B)` with `matop_dest(/, A::UpperTriangular, B)`.

I like that. This concept could be merged with the existing `matprod_dest` by defining

```julia
matprod_dest(A, B, _) = matop_dest(*, A, B)
```

src/diagonal.jl
araujoms: I don't think it was a typo, but an assumption that `promote_op(\, eltype(D), eltype(S)) == promote_op(/, eltype(S), eltype(D))`. It is true for unitful.

araujoms:
You probably mean
```julia
matprod_dest(A, B, _) = matop_dest(matprod, A, B)
```
But in any case, I think this can be left for another PR.

araujoms (approved):

src/diagonal.jl
dkarrasch: Yes, unitful is the wrong example. Maybe arrays of (non-square) arrays may be tricky? But given the existing assumption, we can go with your proposal.

PR #1581, opened 2026-04-26
make set_num_threads and get_num_threads public

### LINKED ISSUE #1579: Export or make public `BLAS.set_num_threads`
This is not currently public, but is documented in [LinearAlgebra](https://docs.julialang.org/en/v1/stdlib/LinearAlgebra/#LinearAlgebra.BLAS.set_num_threads).

stevengj:
Fixes #1579

### DIFF
--- a/src/blas.jl
+++ b/src/blas.jl
@@ -84,6 +84,8 @@ export
     trsm!,
     trsm
 
+public set_num_threads, get_num_threads
+
 using ..LinearAlgebra: libblastrampoline, BlasReal, BlasComplex, BlasFloat, BlasInt, DimensionMismatch, checksquare, chkstride1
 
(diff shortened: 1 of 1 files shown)


jishnub:
There's an existing list of public names in the `LinearAlgebra.jl` file, and it might be better to add to that.

stevengj:
This is in the `LinearAlgebra.BLAS` submodule, which has its own `export` list and hence should have its own `public` list.

jishnub (approved):

dkarrasch:
Should stuff like this be backported to v1.13?

jishnub:
I think it should be fine to backport this. It's strictly non breaking and addresses a mismatch with the documentation.

PR #1582, opened 2026-04-27
Equality/`hash` edge cases for `QRCompactWY`

jishnub:
This ensures the we don't accidentally end up with cases where two objects are equal but their hashes differ, or we divide by zero if the matrices are empty.

Co-authored-by: Copilot <copilot@github.com>

### DIFF
--- a/test/qr.jl
+++ b/test/qr.jl
@@ -517,7 +517,7 @@ end
             # this should have worked before, Q is square, and R is 0 × 0:
             A_zero_cols = rand(F, dimA, 0)
             qr_zero_cols = qr(A_zero_cols)
-            @test size(qr_zero_cols.Q) == (dimA, dimA)
+            @test size(qr_zero_cols.Q) == (dimA, dimA)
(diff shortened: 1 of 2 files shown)


test/qr.jl
```diff
@@ -542,4 +542,23 @@ end
...
+        F, G = qr(A), qr(A)
+        @test F == G
+        @test isequal(F, G)
+        @test hash(F) == hash(G)
+    end
+
+    @testset "QRCompactWY equality checks T shape" begin
+        # This is a contrived example, but ensures that we don't violate the hash contract
+        F = LinearAlgebra.QRCompactWY(zeros(1, 1), reshape([1.0], 1, 1))
+        G = LinearAlgebra.QRCompactWY(zeros(1, 1), reshape([1.0 99.0], 1, 2))
```
araujoms, line 556: Why is `G` even valid, though? Shouldn't the constructor just throw an error?

jishnub: Perhaps it should

PR #1584, opened 2026-04-28
Make `generic_matmatmul!` etc `public`

timholy:
To avoid ambiguities in dispatching on `mul!`, it makes sense to allow LinearAlgebra to unwrap `Adjoint` and `Transpose` wrappers first. This means that packages should be free to specialize `generic_matmatmul`, `generic_matmatmul!`, and `generic_matvecmul!` for their own types. This PR marks those functions as `public` so that they can be extended by external packages.

While I have merge privileges, I will defer the decision about whether to merge to others, as this introduces a stability contract that needs to be maintained until the next breaking release.


jishnub:
I'd go with just the in-place functions. `generic_matmatmul` seems to generate the destination and call the in-place methods with it. If one wishes to alter the destination type, they should specialize `matprod_dest` instead.

I think this PR is fine otherwise, as I think the intent was to make these public in https://github.com/JuliaLang/julia/pull/49806.

timholy:
Done. I also added some tests to guard this as public API. (Tests were written by AI, triggered by my direct instructions to do this together with an explanatory comment.)

I'll still defer to an active maintainer to make the merge.

jishnub (approved):

araujoms:
CI is failing locally because of this PR:
```julia
Docstrings: Test Failed at /home/mateus/julia/LinearAlgebra.jl/test/runtests.jl:17
  Expression: isempty(Docs.undocumented_names(LinearAlgebra))
   Evaluated: isempty([:generic_matmatmul!, :generic_matvecmul!])
```
I don't know why it's not failing on github, but in any case public names should be documented.

topolarity:
> I don't know why it's not failing on github, but in any case public names should be documented.

Looks like the Docstrings tests are being accidentally skipped:

> The repo's Buildkite runs .ci/run_tests.jl, which calls Base.runtests(["LinearAlgebra"]) — julia's harness. Julia's harness (test/choosetests.jl:246-249) sees LinearAlgebra's testgroups file and expands the test set into the individual group files (matmul.jl, triangular.jl, …), executing them directly — it never executes runtests.jl itself, and the Docstrings (and versioninfo) testsets live only there, after the include loop.

(analysis from Claude Fable)


looks like `Base.runtests(["LinearAlgebra"])` skips these but `Pkg.test("LinearAlgebra")` hits them

PR #1586, opened 2026-05-06
buildkite: add per-job timeouts and automatic retries

ViralBShah:
## Summary
- Add `timeout_in_minutes: 90` to all four Buildkite jobs so a hung test can't hold a shared agent indefinitely.
- Add `retry.automatic` for `exit_status: -1` and `signal_reason: agent_stop` (limit 2) so transient agent disconnects self-heal instead of turning PRs red.

## Motivation
Recent dependabot PR #1585 went red because the `macos-aarch64` agent dropped ~4 seconds into the job — a clear infra flake, not a test failure. With these settings Buildkite would have retried automatically and the PR would have stayed green. The `timeout_in_minutes` is defensive: today there's no upper bound on how long a stuck job can hold an agent.

## Test plan
- [ ] Confirm the four jobs still run on this PR's Buildkite build
- [ ] Confirm the `retry` block parses and shows up in the Buildkite UI for each step

### DIFF
--- a/.buildkite/runtests.yml
+++ b/.buildkite/runtests.yml
@@ -6,9 +6,9 @@ steps:
     retry:
       automatic:
         - exit_status: -1
-          limit: 2
+          limit: 1
         - signal_reason: agent_stop
-          limit: 2
+          limit: 1
     plugins:
       # Julia installation outside the sandbox
       - JuliaCI/julia#v1:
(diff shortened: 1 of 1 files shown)


DilumAluthge:
Can you set a max number of retries? Otherwise in theory we can get caught in an infinite loop of retries.

ViralBShah:
Doesn't the `limit` do exactly that?

DilumAluthge:
Ah, I missed that, yes it does.

DilumAluthge (approved):

dkarrasch:
I assume this can be closed now that we move away from buildkite?

PR #1587, opened 2026-05-06
unify mat*_dest functions

araujoms:
Follow-up to #1578. The goal is to have a more consistent interface for packages using the mat_dest mechanism.

### DIFF
--- a/src/matmul.jl
+++ b/src/matmul.jl
@@ -58,7 +58,7 @@ function (*)(A::StridedMaybeAdjOrTransMat{T}, x::StridedVector{S}) where {T<:Bla
 end
 function (*)(A::AbstractMatrix, x::AbstractVector)
     matmul_size_check(size(A), size(x))
-    mul!(matop_dest(matprod, A, x), A, x)
+    mul!(matop_dest(*, A, x), A, x)
 end
 
 # these will throw a DimensionMismatch unless B has 1 row (or 1 col for transposed case):
@@ -118,7 +118,7 @@ julia> [1 1; 0 1] * [1 0; 1 1]
 mul(A::AbstractMatrix, B::AbstractMatrix) = _mul(A, B)
 function _mul(A::AbstractMatrix, B::AbstractMatrix)
     matmul_size_check(size(A), size(B))
-    mul!(matop_dest(matprod, A, B), A, B)
+    mul!(matop_dest(*, A, B), A, B)
 end
 
 """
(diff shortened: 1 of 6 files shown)


dkarrasch:
Ha, I started working on the same.

araujoms:
Sorry, ich konnte mich nicht aufhalten. Ich hoffe du hast nicht viel Zeit damit verbracht.

src/matmul.jl
```diff
@@ -119,41 +118,33 @@ julia> [1 1; 0 1] * [1 0; 1 1]
...
-"""
-    matrdiv_dest(A, B)
-
-Return an appropriate `AbstractArray` that may be used to store the result of `A / B`.
-
-!!! compat
-    This function requires at least Julia 1.14
-"""
-matrdiv_dest(A, B) = similar(A, promote_op(/, eltype(A), eltype(B)), size(A))
+matop_dest(::typeof(matprod), A, B) = similar(B, promote_op(matprod, eltype(A), eltype(B)), (size(A, 1), size(B, 2)))
```
dkarrasch: Since we don't just pass through the function argument, I thought it was easier to just have

```suggestion
matop_dest(::typeof(*), A, B) = similar(B, promote_op(matprod, eltype(A), eltype(B)), (size(A, 1), size(B, 2)))
```

dkarrasch (commented):
LGTM.

src/matmul.jl
araujoms: Not only easier, but actually necessary for other packages to use a sensible interface, since `matprod` is a private function.

dkarrasch (approved):

araujoms:
Is there any package other than than `SparseArrays` using `matprod_dest`? We need to give a heads-up, otherwise it'll mysteriously stop working.

dkarrasch:
Why? The only thing that doesn't work is when a package calls `matprod_dest` with a `T` other than the `promote_op(matprod, ...)`. I.e. you can't influence the `*` like that anymore. But that would be strange, given that one [sh/c]ould use either `similar` or an array constructor.

araujoms:
No, that still works, because I put a conversion there.

What will stop working is when a package extends `matprod_dest` with its own types to determine the output; because now `_mul` uses `matop_dest`:
```julia
function _mul(A::AbstractMatrix, B::AbstractMatrix)
    matmul_size_check(size(A), size(B))
    mul!(matop_dest(*, A, B), A, B)
end
```

dkarrasch:
True. I think there is website like a package catalogue where one could search all registered packages for usage of specific functions. I forgot the name, but shouldn't be hard to find.

dkarrasch:
This all uses of `matprod`: https://juliahub.com/ui/Search?q=matprod&type=code There are no results for `matprod_dest`.

araujoms:
We might as well remove it completely then.

dkarrasch:
You're brave! This is a public function, so I'm not sure. I personally broke enough in here (`AbstractQ` calling). :sweat_smile: But I'd be in favor.

araujoms:
`AbstractQ` was definitely worth it, I had to deal with the interface both before and after it.

I'll make a PR then to get people's comments then.

PR #1588, opened 2026-05-07
Move dispatch in matrix operations to lower-level functions

dkarrasch:
This is parallel effort to #1587, for mutual inspiration. I tried to make use of the mechanisms to further reduce high-level function overloads.

This seems to currently increase runtime of `Diagonal * *Triangular` for a reason I don't understand.

I won't have time next week to work on it.

### DIFF
--- a/src/matmul.jl
+++ b/src/matmul.jl
@@ -125,25 +125,26 @@ end
     matprod_dest(A, B, T)
 
 Return an appropriate `AbstractArray` with element type `T` that may be used to store the result of `A * B`.
-This function is only kept for backwards compatibility, use instead `matop_dest`.
+This function is only kept for backwards compatibility, use `matop_dest` instead.
 
-!!! compat
-    This function requires at least Julia 1.11
+!!! compat "Julia 1.11"
+    This function requires at least Julia 1.11.
 """
 matprod_dest(A, B, T) = convert(AbstractArray{T}, matop_dest(*, A, B))
 
 """
     matop_dest(op, A, B)
 
-Return an appropriate `AbstractArray` that may be used to store the result of `op(A, B)`.
+Return an appropriate `AbstractArray` that may be used to store the result of `op(A, B)`,
+where `op` is one of `*`, `/`, or `\\`.
 
-!!! compat
-    This function requires at least Julia 1.14
+!!! compat "Julia 1.14"
+    This function requires at least Julia 1.14.
 """
 matop_dest(::typeof(\), A, B) = similar(B, promote_op(\, eltype(A), eltype(B)), size(B))
 matop_dest(::typeof(/), A, B) = similar(A, promote_op(/, eltype(A), eltype(B)), size(A))
 matop_dest(::typeof(*), A, B) = similar(B, promote_op(matprod, eltype(A), eltype(B)), (size(A, 1), size(B, 2)))
-matop_dest(::typeof(*), A, b::AbstractVector) = similar(b, promote_op(matprod, eltype(A), eltype(b)), axes(A,1))
+matop_dest(::typeof(*), A, b::AbstractVector) = similar(b, promote_op(matprod, eltype(A), eltype(b)), axes(A, 1))
 
 const Ops = Union{typeof(*), typeof(\), typeof(/)}
 
@@ -154,8 +155,6 @@ Post-processing of `C`, which is assumed to be the result of `op(A, B)`.
 """
 postop_proc(::Ops, C, _, _) = C
 
-matop_dest(::typeof(\), A, B) = similar(B, promote_op(\, eltype(A), eltype(B)), size(B))
-matop_dest(::typeof(/), A, B) = similar(A, promote_op(/, eltype(A), eltype(B)), size(A))
 # optimization for dispatching to BLAS, e.g. *(::Matrix{Float32}, ::Matrix{Float64})
 # but avoiding the case *(::Matrix{<:BlasComplex}, ::Matrix{<:BlasReal})
 # which is better handled by reinterpreting rather than promotion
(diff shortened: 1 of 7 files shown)


araujoms:
That's more like a superset; I did the minimal fusion, you also generalised `postop_proc` and used it to simplify more code. I'd suggest to merge mine and rebase yours. That's not just to give me less work, but also would make your PR smaller.

dkarrasch:
I suggest to use this as a grab bag of ideas for other PRs: in little chunks, checking for performance regressions. Resolving the merge conflicts would take probably more time then to quickly come up with new PRs with small changes. The ultimate goal is to make generic `*` or `mul` (and eventually solves) handle most of the specific cases.

araujoms:
Ok, I resolved the merge conflicts, since they were my fault, now it's easier to see what this PR is doing.

dkarrasch:
Awesome. Then we need to check for performance changes for methods that are no longer handled by specific methods but by the generic one instead. Especially the diagonal-triangular (and vice versa) multiplication got hit significantly.

araujoms:
A technique that I use to test whether the dedicated method is being called or the generic fallback is to test whether the output is `==` the one produced by calling the dedicated method directly. Might be useful here.

dkarrasch:
The issue is somewhat different: I want to remove specialized methods and want to make sure the generic fallback essentially generates the same code (runs as fast!), with all the `postop_proc`s and `matop_dest`s inserted.

araujoms:
I see, so you're not worried about the performance of the matrix multiplication itself, but about the latency of calling it. I'm afraid that's beyond my skills then.

dkarrasch:
No, sorry, I'm being unclear. What I mean is that this PR removes some specialized `mul` methods, and hence the generic `mul` method is taking over. I want the performance to be the same as without this change. For instance, on my machine, I get the following timings:

```julia
julia> using LinearAlgebra, BenchmarkTools

julia> D = Diagonal(randn(1000));

julia> U = UpperTriangular(randn(1000,1000));

julia> @btime $D * $U;
  835.078 μs (3 allocations: 7.63 MiB) # this PR
  300 μs (3 allocations: 7.63 MiB) # on v1.12.6
```

The "multiplication kernel" (the function that does the actual work) runs equally fast on both versions (roughly 200 µs), but there must be some surrounding code that increases the runtime unacceptably.

dkarrasch:
So, I don't understand the locally observed performance regression: I checked out older commits of LinearAlgebra.jl in my VS Code environment and saw the same "regression" (measured in VS Code) compared to Julia v1.12 in an extra terminal.

I put back the diag-tri-mul methods, because I think they have value: in the non-strided case (like sparse), there may be optimized multiplication methods with diagonals.

araujoms:
Same here, I've tested 1.13-rc1:
```julia
julia> @btime D * U setup=(D = Diagonal(randn(1000)); U = UpperTriangular(randn(1000,1000)));
  108.275 μs (3 allocations: 7.63 MiB)
```
and nightly:
```julia
julia> @btime D * U setup=(D = Diagonal(randn(1000)); U = UpperTriangular(randn(1000,1000)));
  741.977 μs (3 allocations: 7.63 MiB)
```

dkarrasch:
It's good to know that our latest changes haven't introduced that, but it's sad to see such a significant regression. I checked with `@code_typed`, but it seems that in both cases code is inlined up until the same function (`__diagmul_nonzeroalpha!` or something like that), which hasn't changed in a while...

araujoms:
Ok, the regression is not in `LinearAlgebra` at all, I just ran this PR with the 1.13-rc1 binary and its fine. Conversely, I tried ancient versions of `LinearAlgebra` with the nightly binary and the regression is there.

dkarrasch:
When I look at the method list for the three operations with `(AbstractMatrix, AbstractMatrix)` signatures, it looks very clean. The ones left do something very different than "create an array" and "call `mul!` or `ldiv!`.

I believe this is in good shape now. Thanks @araujoms for your support!

src/matmul.jl
```diff
@@ -118,33 +118,43 @@ julia> [1 1; 0 1] * [1 0; 1 1]
...
 """
-matop_dest(::typeof(*), A, B) = similar(B, promote_op(matprod, eltype(A), eltype(B)), (size(A, 1), size(B, 2)))
-matop_dest(::typeof(*), A, x::AbstractVector) = similar(x, promote_op(matprod, eltype(A), eltype(x)), axes(A, 1))
-
 matop_dest(::typeof(\), A, B) = similar(B, promote_op(\, eltype(A), eltype(B)), size(B))
 matop_dest(::typeof(/), A, B) = similar(A, promote_op(/, eltype(A), eltype(B)), size(A))
+matop_dest(::typeof(*), A, B) = similar(B, promote_op(matprod, eltype(A), eltype(B)), (size(A, 1), size(B, 2)))
+matop_dest(::typeof(*), A, b::AbstractVector) = similar(b, promote_op(matprod, eltype(A), eltype(b)), axes(A, 1))
+
+const Ops = Union{typeof(*), typeof(\), typeof(/)}
```
araujoms: Perhaps a more distinctive name? If someone sees `Ops` on a different file they'll have no idea what's supposed to be. Maybe `MatrixBinaryOps`, or `MatOps`?

dkarrasch: Good point. Perhaps `MatrixMulOrDiv`, or `Mat[rix]LinalgOps`? Technically, `+` and `-` (and many others) are also binary operations.

araujoms: Or just `MulOrDiv`?

araujoms (approved):

dkarrasch:
Once tests pass, we should merge and bump the stdlib. Then we can continue with SparseArrays.jl. The PRs should be "ready".

IanButterworth:
Can I get some help with https://github.com/JuliaLang/LinearAlgebra.jl/pull/1598 which tries to fix performance regressions from this PR

PR #1591, opened 2026-05-07
[RFC] remove matprod_dest

araujoms:
This is technically breaking because it removes a public function. However, it is effectively non-breaking because nobody uses it:
https://juliahub.com/ui/Search?q=matprod_dest&type=code (except for `SparseArrays`, which we can fix ourselves). It has been public since 1.12.

Conversely, #1587 was technically non-breaking, because it didn't change the behaviour of any public function, but it was effectively breaking, because any packaged that relied on the result of `*` being defined via `matprod_dest` would stop working. Again, this was only `SparseArrays` as far as I can tell.

I think the sensible alternatives is to either remove the now-useless `matprod_dest` or undo #1587.

I'd specially like to hear @jishnub's opinion, since he's the one who added `matprod_dest` in the first place, in https://github.com/JuliaLang/julia/pull/55537


dkarrasch:
In fact, `SparseArrays.jl` will be the prototype for how to override output array determination without overloading `*` (or `LinearAlgebra.mul`). x-ref: https://github.com/JuliaSparse/SparseArrays.jl/pull/693

jishnub:
Seems fine to me. I've wanted this change for a while, and thanks for leading this.

araujoms:
Awesome. Since there's no hurry I'll leave this PR open for another week in case somebody else wants to comment.

PR #1594, opened 2026-05-11
fix: add missing math functions for UniformScaling

### LINKED ISSUE #695: Matrix functions for UniformScaling
I was surprised by this:
```julia
julia> sqrt(5.0I)
ERROR: MethodError: no method matching sqrt(::UniformScaling{Float64})
Closest candidates are:
  sqrt(::Float16) at math.jl:1085
  sqrt(::Complex{Float16}) at math.jl:1086
  sqrt(::Missing) at math.jl:1138
  ...
Stacktrace:
 [1] top-level scope at REPL[25]:1

julia> sqrt(Matrix(5I, 5, 5))
5×5 Array{Float64,2}:
 2.23607  0.0      0.0      0.0      0.0
 0.0      2.23607  0.0      0.0      0.0
 0.0      0.0      2.23607  0.0      0.0
 0.0      0.0      0.0      2.23607  0.0
 0.0      0.0      0.0      0.0      2.23607
```
Since we already have a few [matrix functions](https://en.wikipedia.org/wiki/Matrix_function) defined, such as
https://github.com/JuliaLang/julia/blob/e36fe95a639c6956aa39367bef8e727b174e52b2/stdlib/LinearAlgebra/src/dense.jl#L702-L759
it would seem reasonable to expect at least these elementary functions to be defined for `UniformScaling` as well:
- [x] `^(::UniformScaling, ::Number)` (already added in https://github.com/JuliaLang/julia/pull/34008)
- [x] `exp`, `log`
- [x] `expm1`, `logp1`
- [x] `sin`, `cos`, `tan`, `csc`, `sec`, `cot` + inverses and hyperbolic versions
- [x] `sqrt`, `cbrt`
- [x] `pinv`, `tr`
- [x] `real`, `imag`, `reim`

All of the above have been added in JuliaLang/julia#28872, the question is now which additional functions are of interest, i.e. which are used for matrices? Candidates are

- [ ] `abs` - one could argue that the matrix generalization is `det`
      (which yields a scalar), but you can compute `abs(A)` as
      `Q*abs.(λ)*Q'`, which is something different.
- [ ] `abs2`
- [ ] `log10`, `log2`
- [ ] `exp2`, `exp10`
- [ ] `sinc`, `cosc`

I am unsure if the functions below are sensible for matrices, possibly
with the exception of `sinpi` and `cospi`.

- [ ] `sinpi`, `cospi`
- [ ] `sind`, `cosd`, `tand`, `secd`, `cscd`, `cotd`
- [ ] `asind`, `acosd`, `atand`, `asecd`, `acscd`, `acotd`
- [ ] `deg2rad`, `rad2deg`
- [ ] `mod2pi`

From JuliaLang/julia#28872 remains also the idea of matrix decompositions.

Mxyank24:
Fixes #695

Found some missing math functions for UniformScaling while going 
through open issues.

```julia
abs(5.0I)     # errors
abs2(5.0I)    # errors
log2(4.0I)    # errors
log10(100.0I) # errors
sinpi(0.5I)   # errors
cospi(0.0I)   # errors
```

The fix was pretty straightforward — I just added them to the 
existing loop that handles the other math functions in 
uniformscaling.jl. After the fix they all work as expected.

### DIFF
--- a/src/uniformscaling.jl
+++ b/src/uniformscaling.jl
@@ -173,8 +173,8 @@ for f in (:exp, :log, :cis,
           :asinh, :acosh, :atanh,
           :csch, :sech, :coth,
           :acsch, :asech, :acoth,
-          :abs, :abs2, :log2, :log10,
-          :sinpi, :cospi )
+          :log2, :log10,
+          :sinpi, :cospi)
     @eval Base.$f(J::UniformScaling) = UniformScaling($f(J.λ))
 end
 for f in (:sincos, :sincosd)
--- a/src/uniformscaling.jl
+++ b/src/uniformscaling.jl
@@ -162,17 +162,19 @@ isposdef(J::UniformScaling) = isposdef(J.λ)
 (-)(A::AbstractMatrix, J::UniformScaling)   = A + (-J)
 
 # matrix functions
-for f in ( :exp,   :log, :cis,
-           :expm1, :log1p,
-           :sqrt,  :cbrt,
-           :sin,   :cos,   :tan,
-           :asin,  :acos,  :atan,
-           :csc,   :sec,   :cot,
-           :acsc,  :asec,  :acot,
-           :sinh,  :cosh,  :tanh,
-           :asinh, :acosh, :atanh,
-           :csch,  :sech,  :coth,
-           :acsch, :asech, :acoth )
+for f in (:exp, :log, :cis,
+          :expm1, :log1p,
+          :sqrt, :cbrt,
+          :sin, :cos, :tan,
+          :asin, :acos, :atan,
+          :csc, :sec, :cot,
+          :acsc, :asec, :acot,
+          :sinh, :cosh, :tanh,
+          :asinh, :acosh, :atanh,
+          :csch, :sech, :coth,
+          :acsch, :asech, :acoth,
+          :abs, :abs2, :log2, :log10,
+          :sinpi, :cospi )
     @eval Base.$f(J::UniformScaling) = UniformScaling($f(J.λ))
 end
 for f in (:sincos, :sincosd)
--- a/test/uniformscaling.jl
+++ b/test/uniformscaling.jl
@@ -597,15 +597,12 @@ end
     @test 0*A != 0*I
 end
 
-@testset "math functions" begin
-    J = UniformScaling(4.0)
-    @test abs(J) == UniformScaling(4.0)
-    @test abs(UniformScaling(-3.0)) == UniformScaling(3.0)
-    @test abs2(J) == UniformScaling(16.0)
-    @test log2(J) == UniformScaling(2.0)
+@testset "log2, log10, sinpi, cospi for UniformScaling" begin
+    @test log2(UniformScaling(4.0))   == UniformScaling(2.0)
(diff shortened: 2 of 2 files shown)


stevengj:
I think the lack of methods here is intentional?

`abs` and `abs2` aren't unambiguously defined for square matrices, I think?

The other functions *could* be defined for square matrices, but currently aren't defined for matrices in Julia.

Mxyank24:
> I think the lack of methods here is intentional?
> 
> `abs` and `abs2` aren't unambiguously defined for square matrices, I think?
> 
> The other functions _could_ be defined for square matrices, but currently aren't defined for matrices in Julia.

That's a fair point, thanks for the clarification! 

Should I remove abs and abs2 and just keep log2, log10, 
sinpi and cospi? Those seem less ambiguous since they 
follow the same scalar-applied-to-eigenvalues pattern 
as the other functions already in the loop.

stevengj:
I don't think it makes sense to add methods of functions like `log2` for `UniformScaling` unless we also define them for arbitrary square matrices.  Which we absolutely could, easily!  But there doesn't seem to have been much demand for them, and they aren't provided in e.g. scipy or matlab?

Mxyank24:
> I don't think it makes sense to add methods of functions like `log2` for `UniformScaling` unless we also define them for arbitrary square matrices. Which we absolutely could, easily! But there doesn't seem to have been much demand for them, and they aren't provided in e.g. scipy or matlab?

Ah that makes sense, I didn't think about it from that angle!
Is there a smaller subset of these that would make sense to add? 
Or would it be better to first open a discussion about adding 
them for general matrices?

araujoms:
What's the ambiguity in the definition of `abs`? As far as I know it is defined as `abs(M) = sqrt(M'M)`, even for non-square matrices. Are you thinking about the possibility of defining it instead as `sqrt(M*M')`? That's of course an arbitrary choice, but since both alternatives are equally valid we can pick either.

araujoms:
I think it only makes sense to add functions to `UniformScaling` that are defined for general matrices. I'd close this PR and the linked issue, and open another issue about whichever functions we might want to add to general matrices.

Mxyank24:
> I think it only makes sense to add functions to `UniformScaling` that are defined for general matrices. I'd close this PR and the linked issue, and open another issue about whichever functions we might want to add to general matrices.

That makes sense! Happy to close this one. Should I open a new 
issue about adding these functions for general matrices, or is 
that something the team would prefer to handle internally?

araujoms:
Great. Feel free to open it yourself.

stevengj:
> What's the ambiguity in the definition of abs? As far as I know it is defined as abs(M) = sqrt(M'M), even for non-square matrices.

Oh, right, that seems to be the convention.   Might be nice to have a function for this (which is subtle to make accurate, without squaring the condition number).

See also https://discourse.julialang.org/t/absolute-value-of-a-matrix/98301

It's related to computing the polar decomposition; see also Highham et al, ["Computing the Polar Decomposition and the Matrix Sign Decomposition in Matrix Groups"](https://epubs.siam.org/doi/10.1137/S0895479803426644) (2004).  See also this discussion of polar-decomposition algorithms in Julia: https://discourse.julialang.org/t/polar-decomposition-of-a-matrix/48830/14

PR #1493, opened 2025-11-26
use syevr meta-algorithm for SymTridiagonal

stevengj:
Implement the same meta-algorithm for `SymTridiagonal` as what `syevr` (the `RobustRepresentations` algorithm) [does internally](https://netlib.org/lapack/explore-html/d1/d56/group__heevr_gaa334ac0c11113576db0fc37b7565e8b5.html) for its tridiagonal eigensolve.  Fixes #1491.

This involves two changes to our current algorithm:
1. For `eigen` or `eigvals`, if `stegr` or `stev` (respectively) fails it now falls back to `stebz` and `stein`.
2. For `eigen` and `eigvals`, if a subset of the eigenvalues is requested, it now calls `stebz` (and `stein`, for eigvecs) instead of `stegr`.

I'm not so sure about change (2) — it seems to be slightly slower on my machine for a random 1000x1000 matrix.  Any idea why `syevr` does this?

In order to implement change (1), I had to make an additional copy of the entries of the `SymTridiagonal` matrix, to preserve them for the fallback routine.  I compensated for this by calling the new meta-algorithm directly from `eigen` and `eigvals`.  However, it means that if you directly call `eigen!` or `eigvals!` on `SymTridiagonal` it will have an extra allocation.  Ultimately, I would like us to allow multiple algorithm choices similar to `eigen(::Symmetric)`.

PR #1595, opened 2026-05-11
make eigvals(::SymTridiagonal) accept sortby

araujoms:
Follow-up to #1477. That closes the last exception, and now all matrix types accept the `sortby` argument.

Also fixes a (very unlikely) bug introduced in #1493, as eigenvalues computed from `stebz!` with ordering `'B'` are in general not ordered.

### DIFF
--- a/src/tridiag.jl
+++ b/src/tridiag.jl
@@ -282,11 +282,12 @@ rdiv!(B::AbstractVecOrMat, A::SymTridiagonal; shift::Number=false) = rdiv!(B, ld
 # tridiagonal eigensolver meta-algorithm from LAPACK.syevr! for alg==RobustRepresentations()
 #   - if all eigenvalues are desired, call stev == sterf (eigvals) or stegr == stemr (eigen)
 #   - otherwise, and also if an error occurs, fall back to stebz and (if eigvecs wanted) stein
-function syevr_tri_eigen(range::AbstractChar, dv::AbstractVector{T}, ev::AbstractVector{T}, vl::Real, vu::Real, il::Integer, iu::Integer) where {T<:BlasReal}
+function syevr_tri_eigen(range::AbstractChar, dv::AbstractVector{T}, ev::AbstractVector{T}, vl::Real, vu::Real, il::Integer, iu::Integer; sortby = eigsortby) where {T<:BlasReal}
     if range == 'A' || (range == 'I' && il == 1 && iu == length(dv))
         try
             # need to copy dv, ev so that they are available for fallbacks, below
-            return Eigen(LAPACK.stegr!('V', range, copymutable(dv), copymutable(ev), vl, vu, il, iu)...)
+            values, vectors = LAPACK.stegr!('V', range, copymutable(dv), copymutable(ev), vl, vu, il, iu)
+            return Eigen(sorteig!(values, vectors, sortby == eigsortby ? nothing : sortby)...)
         catch ex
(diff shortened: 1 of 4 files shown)


src/tridiag.jl
```diff
@@ -282,19 +282,20 @@ rdiv!(B::AbstractVecOrMat, A::SymTridiagonal; shift::Number=false) = rdiv!(B, ld
...
 #   - if all eigenvalues are desired, call stev == sterf (eigvals) or stegr == stemr (eigen)
 #   - otherwise, and also if an error occurs, fall back to stebz and (if eigvecs wanted) stein
-function syevr_tri_eigen(range::AbstractChar, dv::AbstractVector{T}, ev::AbstractVector{T}, vl::Real, vu::Real, il::Integer, iu::Integer) where {T<:BlasReal}
+function syevr_tri_eigen(range::AbstractChar, dv::AbstractVector{T}, ev::AbstractVector{T}, vl::Real, vu::Real, il::Integer, iu::Integer; sortby = eigsortby) where {T<:BlasReal}
     if range == 'A' || (range == 'I' && il == 1 && iu == length(dv))
         try
             # need to copy dv, ev so that they are available for fallbacks, below
-            return Eigen(LAPACK.stegr!('V', range, copymutable(dv), copymutable(ev), vl, vu, il, iu)...)
+            values, vectors = LAPACK.stegr!('V', range, copymutable(dv), copymutable(ev), vl, vu, il, iu)
+            return Eigen(sorteig!(values, vectors, sortby == eigsortby ? nothing : sortby)...)
```
stevengj, line 290: Does this conditional introduce a dynamic dispatch, or is the `sortby == eigsortby` determined statically?

Since `sorteig!` calls `issorted` before doing anything, in practice the cost of just passing `sortby` unconditionally is probably negligible?

src/tridiag.jl
```diff
@@ -306,53 +307,54 @@ function syevr_tri_eigvals(range::AbstractChar, dv::AbstractVector{T}, ev::Abstr
...
+eigen!(A::SymTridiagonal{<:BlasReal,<:StridedVector}, vl::Real, vu::Real; kws...) =
+    syevr_tri_eigen('V', A.dv, A.ev, vl, vu, 0, 0; kws...)
+eigen(A::SymTridiagonal{<:BlasReal,<:StridedVector}, vl::Real, vu::Real; kws...) =
+    eigen!(A, vl, vu; kws...)
+eigen(A::SymTridiagonal{T}, vl::Real, vu::Real; kws...) where T =
+    eigen!(copymutable_oftype(A, eigtype(T)), vl, vu; kws...)
+
+function eigvals!(A::SymTridiagonal{<:BlasReal,<:StridedVector}; sortby = eigsortby)
+    vals = syevr_tri_eigvals('A', A.dv, A.ev, 0.0, 0.0, 0, 0)
+    return sorteig!(vals, sortby == eigsortby ? nothing : sortby)
```
stevengj, line 334: I thought you said that `stebz!` (which might be called by `syevr_tri_eigvals`) doesn't necessarily sort?

stevengj: Oh, nevermind, `stebz!` with the `'E'` argument should return sorted eigvals, right?

src/tridiag.jl
araujoms: It's static, because `eigsortby` is a function, just checked with Cthulhu.

stevengj (approved):

PR #1598, opened 2026-05-12
matmul: Fix sparse perf regression - route default `matop_dest(*, …)` through `matprod_dest`

IanButterworth:
Developed with Claude:

---

After #1588, `*` dispatches through `matop_dest(*, A, B)`, whose default allocates the destination directly via `similar(B, …)`. This bypassed out-of-tree `matprod_dest` overloads (notably the SparseArrays ones for dense × sparse, sparse × dense_triangular, etc.), causing major performance regressions where the destination is now an empty `SparseMatrixCSC` instead of a dense matrix:
https://github.com/JuliaCI/NanosoldierReports/blob/master/benchmark/by_date/2026-05/11/report.md

image

Restore the previous extension point by making the matrix-matrix `matop_dest(*, …)` fallback delegate to `matprod_dest(A, B, T)`, and keep `matprod_dest` as the documented hook (no longer a backwards-compat-only shim).

### DIFF
--- a/src/matmul.jl
+++ b/src/matmul.jl
@@ -125,12 +125,11 @@ end
     matprod_dest(A, B, T)
 
 Return an appropriate `AbstractArray` with element type `T` that may be used to store the result of `A * B`.
-This function is only kept for backwards compatibility, use `matop_dest` instead.
 
 !!! compat "Julia 1.11"
     This function requires at least Julia 1.11.
 """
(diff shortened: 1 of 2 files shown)


araujoms:
The idea is to fix it in SparseArrays via https://github.com/JuliaSparse/SparseArrays.jl/pull/693

Issue #1597, opened 2026-05-12
eigvecs(::AbstractTriangular) are not normalized to unit length


stevengj:
`eigvecs` and `eigen` almost always normalizes eigenvectors to unit length, but I noticed that they do not do so for triangular matrices.  For example:
```jl
julia> eigvecs(UpperTriangular([1 2 3; 0 4 5; 0 0 7]))
3×3 Matrix{Float32}:
 1.0  0.666667  0.633333
 0.0  1.0       1.0
 0.0  0.0       0.6
```
The `trevc` function that we call for eigenvectors (see also #1596) instead normalizes them so that the largest component is ±1, apparently.   When `geevx!` calls the (quasi-)triangular eigensolver, however, I noticed [in the source code of `zgeevx`](https://www.netlib.org/lapack/explore-html/d1/d2a/group__geevx_gae1b71c9b357fa3b500dc74ceb4ce0696.html) that it normalizes the results according to the following algorithm:
```jl
# similar to geevx!, normalize eigenvectors to unit length and make largest component real and positive
function eigvec_normalize!(v::AbstractVector)
    normalize!(v)
    maxabs2, k = findmax(abs2, v) # largest component
    if eltype(v) <: Real # just a sign flip
        v[k] < 0 && (v .= .- v)
    elseif maxabs2 > 0
        v .*= conj(v[k]) / sqrt(maxabs2) # change phase to make v[k] real > 0
        v[k] = real(v[k]) # imaginary part is just roundoff error
    end
    return v
end
```
(This could be optimized further to save a pass over the array, but I doubt it's worth it compared to the cost of computing the eigenvectors in the first place.  The cost of normalization is $O(n^2)$ for an $n \times n$ matrix, but the cost of computing eigenvectors is $O(n^3)$ even for a triangular matrix.)

(Unfortunately, [`dgeevx`](https://www.netlib.org/lapack/explore-html/d1/d2a/group__geevx_ga11536a206dc641c6a59cab916bb1c721.html) does not normalize the sign in the same way, but I think `zgeevx`'s choice is reasonable here whereas the sign in `dgeevx` can be more arbitary.)

Of course, the normalization is arbitrary, and I think we don't document the normalization, but it is a little bit annoying to have the normalization be different just for this matrix type.

[...]

Issue #1596, opened 2026-05-12
use newer LAPACK (3.6.1) trevc3 function for quasi-triangular eigenvectors


stevengj:
Currently, our eigenvectors for triangular matrices are computed by the `LAPACK.trevc!` routine, ever since they were implemented in https://github.com/JuliaLang/julia/pull/5255 by @jiahao in 2013.  However, in 2016, [LAPACK 3.6.1 released a new routine ](https://github.com/Reference-LAPACK/lapack-www/blob/master/lapack-3.6.1.txt)`*trevc3` ([`ztrevc3`](https://netlib.org/lapack/explore-html/d1/d3d/ztrevc3_8f_source.html) etc.) which ~~implements better blocking and hence (reportedly) better performance~~ lets us allocate less memory.

It should be straightforward to provide a `trevc3!` wrapper similar to our existing `trevc!` wrapper, and replace calls to `trevc!` with calls to `trevc3!`.

(I noticed this because, for #858, I was looking through the source of [the `geevx` routine](https://www.netlib.org/lapack/explore-html/d1/d2a/group__geevx_gae1b71c9b357fa3b500dc74ceb4ce0696.html) which is called by our general `eigen!` function.  Nowadays, `geevx` calls `trevc3`.)

stevengj:
Closing as not needed — see https://github.com/JuliaLang/LinearAlgebra.jl/pull/1609#issuecomment-4476176159

stevengj:
Re-opening because `trevc3` allocates less memory than `trevc` (because `trevc3` can tell you the correct buffer sizes, while `trevc` requires you to over-estimate).

PR #1565, opened 2026-03-19
fix dispatch for schur(::UpperHessenberg)

stevengj:
Fixes an apparent bug in https://github.com/JuliaLang/julia/pull/47872 — `schur(A::UpperHessenberg{T})` called `copy_similar(A, eigtype(T)))` but the `copy_similar` function strips away the `UpperHessenberg` wrapper.  As a result, the optimized `schur!` function was not being called.

Following #1557, this changes it to call `eigencopy_oftype`, which preserves the `UpperHessenberg` wrapper (and also `Hermitian` wrappers, but `schur` already has a specialized dispatch for Hermitian cases).   It also adds a fallback `schur!` for non-BLAS types, to avoid breaking anything.

cc @albertomercurio, @Jutho, and @dkarrasch, who were involved in the original PR.

PR #1599, opened 2026-05-12
LAPACK eigen(::UpperHessenberg)

### LINKED ISSUE #858: eigensolvers for Hessenberg matrices
As mentioned in JuliaLang/LinearAlgebra.jl#625, it would be good to have specialized eigensolvers for upper-Hessenberg matrices (i.e. our `UpperHessenberg`) type.

In particular, Hessenberg factorization is the first step of LAPACK's `Matrix` eigensolver and we can skip it if the matrix is already upper Hessenberg.    Indeed, LAPACK provides functions [CHESQR](http://www.netlib.org/lapack/explore-html/d3/db9/group__complex_o_t_h_e_rcomputational_gac942d4e00d833830e58c722d270621c8.html) and [CHGEQZ](https://www.hpc.nec/documents/sdk/SDK_NLC/UsersGuide/man/chgeqz.html) etc for solving eigenproblems with Hessenberg matrices.

stevengj:
Closes #858, providing optimized LAPACK-based `eigen!` for `UpperHessenberg` matrices.  About 50% faster (single-threaded) than the `Matrix` eigensolver for 1000x1000 matrices on my machine.

Includes an `eigvec_normalize!` routine that can also be used for #1597.  Once #1596 is fixed, this code can also be updated to use the improved `trevc3!` for eigenvectors.

### DIFF
--- a/test/hessenberg.jl
+++ b/test/hessenberg.jl
@@ -331,10 +331,21 @@ end
         F = eigen(H)
         @test λ ≈ eigvals(Matrix(H)) ≈ F.values
         @test H * F.vectors ≈ F.vectors * Diagonal(λ)
+        @test F.vectors ≈ eigvecs(Matrix(H)) # normalization should be same
         if T <: LinearAlgebra.BlasFloat
             λ2 = @invoke eigvals!(copy(H)::UpperHessenberg) # test fallback
             @test λ ≈ λ2
         end
+    end
+
+    # be sure to test real-H cases with both purely real and complex eigvals
+    for H in (UpperHessenberg([-1.1 -0.3 -0.0 -0.6 0.5; -0.6 -1.8 -0.3 1.5 1.1; 0.0 1.8 -0.6 1.2 0.9; 0.0 0.0 -0.1 -0.5 -0.2; 0.0 0.0 0.0 -0.5 -1.7]),
+              UpperHessenberg([-0.6 1.0 0.4 0.4 0.3; 1.2 -0.2 -1.5 0.7 0.0; 0.0 -1.3 -1.0 0.5 0.2; 0.0 0.0 0.5 -0.6 0.0; 0.0 0.0 0.0 0.7 -1.5]))
+        λ = eigvals(H)
+        F = eigen(H)
+        @test λ ≈ eigvals(Matrix(H)) ≈ F.values
+        @test H * F.vectors ≈ F.vectors * Diagonal(λ)
+        @test F.vectors ≈ eigvecs(Matrix(H)) # normalization should be same
     end
 end
 
--- a/test/hessenberg.jl
+++ b/test/hessenberg.jl
@@ -331,7 +331,7 @@ end
         F = eigen(H)
         @test λ ≈ eigvals(Matrix(H)) ≈ F.values
         @test H * F.vectors ≈ F.vectors * Diagonal(λ)
-        @test F.vectors ≈ eigvecs(Matrix(H)) # normalization should be same
+        @test Diagonal(F.vectors' * F.vectors) ≈ I
         if T <: LinearAlgebra.BlasFloat
             λ2 = @invoke eigvals!(copy(H)::UpperHessenberg) # test fallback
             @test λ ≈ λ2
@@ -345,7 +345,7 @@ end
         F = eigen(H)
         @test λ ≈ eigvals(Matrix(H)) ≈ F.values
         @test H * F.vectors ≈ F.vectors * Diagonal(λ)
-        @test F.vectors ≈ eigvecs(Matrix(H)) # normalization should be same
+        @test Diagonal(F.vectors' * F.vectors) ≈ I
     end
 end
 
--- a/test/hessenberg.jl
+++ b/test/hessenberg.jl
@@ -334,7 +334,7 @@ end
         @test Diagonal(F.vectors' * F.vectors) ≈ I
         if T <: LinearAlgebra.BlasFloat
             λ2 = @invoke eigvals!(copy(H)::UpperHessenberg) # test fallback
-            F2 = @invoke eigen!(copy(H)::UpperHessenberg{T, <:StridedMatrix{T}}) # fallback
+            F2 = @invoke eigen!(copy(H)::UpperHessenberg) # fallback
             @test λ ≈ λ2 ≈ F2.values
             @test H * F2.vectors ≈ F2.vectors * Diagonal(λ)
             @test Diagonal(F2.vectors' * F2.vectors) ≈ I
--- a/test/hessenberg.jl
+++ b/test/hessenberg.jl
@@ -337,6 +337,7 @@ end
             F2 = @invoke eigen!(copy(H)::UpperHessenberg{T, <:StridedMatrix{T}}) # fallback
             @test λ ≈ λ2 ≈ F2.values
             @test H * F2.vectors ≈ F2.vectors * Diagonal(λ)
+            @test Diagonal(F2.vectors' * F2.vectors) ≈ I
         end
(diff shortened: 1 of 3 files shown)


test/hessenberg.jl
```diff
@@ -331,11 +331,22 @@ end
         F = eigen(H)
         @test λ ≈ eigvals(Matrix(H)) ≈ F.values
         @test H * F.vectors ≈ F.vectors * Diagonal(λ)
+        @test Diagonal(F.vectors' * F.vectors) ≈ I
         if T <: LinearAlgebra.BlasFloat
             λ2 = @invoke eigvals!(copy(H)::UpperHessenberg) # test fallback
             @test λ ≈ λ2
         end
```
araujoms, line 341:
```suggestion
        if T <: LinearAlgebra.BlasFloat #test whether we're calling the specialized method
            λ2 = @invoke eigvals!(copy(H)::UpperHessenberg{T, <:StridedMatrix{T}})
            @test λ == λ2
            F2 = @invoke eigen!(copy(H)::UpperHessenberg{T, <:StridedMatrix{T}})
            @test F == F2
```
There's only a point in making this test if we use exact equality; the `@invoke` macro was selecting the fallback instead of the specialized method, which is why it was failing.

stevengj: I'm confused, why should you expect exact equality between the specialized method and the non-specialized method?

araujoms: The point is to test whether the general method correctly selects the specialized one for `BlasFloat`. It's to prevent silent performance regressions in case the plumbing is redone and the fallback gets called.

stevengj: My point is that `@test F == F2` will generally fail, since they are computed by different algorithms

The original reason for this test was to make sure that the fallback is exercised by the tests.  It need not produce exact equality.

test/hessenberg.jl
```diff
@@ -331,11 +331,22 @@ end
         F = eigen(H)
         @test λ ≈ eigvals(Matrix(H)) ≈ F.values
         @test H * F.vectors ≈ F.vectors * Diagonal(λ)
+        @test Diagonal(F.vectors' * F.vectors) ≈ I
         if T <: LinearAlgebra.BlasFloat
             λ2 = @invoke eigvals!(copy(H)::UpperHessenberg) # test fallback
             @test λ ≈ λ2
```
stevengj:
```suggestion
            F2 = @invoke eigen!(copy(H)::UpperHessenberg{T, <:StridedMatrix{T}}) # fallback
            @test λ ≈ λ2 ≈ F2.values
            @test H * F2.vectors ≈ F2.vectors * Diagonal(λ)
```

test/hessenberg.jl
stevengj: I've updated it to include a test for the fallback `eigen!` method.

test/hessenberg.jl
```diff
@@ -331,11 +331,24 @@ end
         F = eigen(H)
         @test λ ≈ eigvals(Matrix(H)) ≈ F.values
         @test H * F.vectors ≈ F.vectors * Diagonal(λ)
+        @test Diagonal(F.vectors' * F.vectors) ≈ I
         if T <: LinearAlgebra.BlasFloat
             λ2 = @invoke eigvals!(copy(H)::UpperHessenberg) # test fallback
-            @test λ ≈ λ2
+            F2 = @invoke eigen!(copy(H)::UpperHessenberg{T, <:StridedMatrix{T}}) # fallback
+            @test λ ≈ λ2 ≈ F2.values
+            @test H * F2.vectors ≈ F2.vectors * Diagonal(λ)
```
stevengj, line 339:
```suggestion
            @test H * F2.vectors ≈ F2.vectors * Diagonal(λ)
            @test Diagonal(F2.vectors' * F2.vectors) ≈ I
```

test/hessenberg.jl
araujoms: No, it won't fail, they are computed by the same algorithm. I tested it.

test/hessenberg.jl
```diff
@@ -331,11 +331,25 @@ end
         F = eigen(H)
         @test λ ≈ eigvals(Matrix(H)) ≈ F.values
         @test H * F.vectors ≈ F.vectors * Diagonal(λ)
+        @test Diagonal(F.vectors' * F.vectors) ≈ I
         if T <: LinearAlgebra.BlasFloat
             λ2 = @invoke eigvals!(copy(H)::UpperHessenberg) # test fallback
-            @test λ ≈ λ2
+            F2 = @invoke eigen!(copy(H)::UpperHessenberg{T, <:StridedMatrix{T}}) # fallback
```
stevengj: Whoops, should be calling the fallback:
```suggestion
            F2 = @invoke eigen!(copy(H)::UpperHessenberg) # fallback
```

test/hessenberg.jl
stevengj: Oh, that's because you're calling `@invoke eigen!(copy(H)::UpperHessenberg{T, <:StridedMatrix{T}})`, which is literally the same method. 

If you want to test the fallback, as I'm doing in the current code, you want `F2 = @invoke eigen!(copy(H)::UpperHessenberg)`, which generally gives `false` for `F == F2` (since it is a different algorithm).

araujoms: I don't want to test the fallback; that's already tested by running the code on `Float16` and `ComplexF16`. Here I want to test whether the specialized method is used when you call `eigvals(H)` and `eigen(H)`.

stevengj: > I don't want to test the fallback; that's already tested by running the code on Float16 and ComplexF16.

No it's not — those convert to `Float32` via `eigencopy`, because `LinearAlgebra.eigtype(Float16) == Float32`.

araujoms: Argh, you're right, I hadn't realized it. Then there's no way to test the fallback except by directly invoking it. Sorry for the confusion, I completely misunderstood the point of this test. I still think it's valuable to test whether the specialized method gets called, though.

stevengj: codecov tests whether the specialized method gets called.

(It's not common in Julia to explicitly test whether specialized methods are called except via code coverage, as far as I can tell?)

araujoms: Indeed it's uncommon, but the specialized methods do get missed sometimes even with codecov. I've added such a test before in your PR #1565

araujoms: Shouldn't the fallback be removed, though? It's useless for `LinearAlgebra` itself, and prevents other packages from extending it to work with other types (specifically it makes `GenericLinearAlgebra` unhappy).

stevengj: Removing the `eigen!` and `eigvals!` fallbacks seems like a separate question from this PR, since it already had the fallbacks before this PR?  Should we discuss that in a separate issue?

(The fallback was added when `eigencopy_oftype` changed in #1557 to preserve the `UpperHessenberg` wrapper, in order to maintain backwards compatibility.  I see you started using it in https://github.com/JuliaLinearAlgebra/GenericLinearAlgebra.jl/pull/172 … because it relies on type piracy, GenericLinearAlgebra.jl seems like a tricky case to support properly.   Removing the fallback would technically be a breaking change since it would break a package that implemented `eigen!` for `Matrix{MyType}` but not for `UpperHessenberg{MyType}`, though probably you are unlikely to encounter a problem from that in practice.)

araujoms: Sorry, didn't mean to hold up this PR with an unrelated issue. Good for merging?

araujoms (approved):

PR #1601, opened 2026-05-13
bugfix: consistent ordering for composition of Givens rotations

stevengj:
Fixed a bug identified in [this post](https://discourse.julialang.org/t/julia-stability-vs-rust-for-scientific-computing/137094/11?u=stevengj): multiplying two `Givens` objects produces a `Rotation` object representing the composition of the Givens rotations, but adjoints and `rmul` multiplied by the individual Givens objects in an inconsistent order.

cc @dkarrasch, since the bug seems to date back to https://github.com/JuliaLang/julia/pull/46233.

### DIFF
--- a/src/givens.jl
+++ b/src/givens.jl
@@ -391,7 +391,7 @@ function lmul!(R::Rotation, A::AbstractVecOrMat)
     return A
 end
 function rmul!(A::AbstractMatrix, R::Rotation)
-    @inbounds for i in eachindex(R.rotations)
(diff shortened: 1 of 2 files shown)


stevengj:
(Codecov failure is obviously spurious, since this PR adds no new lines of code to the module, but adds more tests.)

Should be good to merge once someone else takes a look?

dkarrasch:
Thanks for fixing this, @stevengj!

PR #1602, opened 2026-05-15
remove Hessenberg fallbacks

araujoms:
Remove the generic fallbacks added in #1557 and #1565, because they break `GenericLinearAlgebra` (and presumably any other package that wants to extend these methods to generic types).

Removing them will break any hypothetical package that defines `eigvals!`, `eigen!`, and `schur!` for a general matrix but not a `UpperHessenberg` one. But I believe it's better to break hypothetical packages than existing ones. Moreover, the hypothetical fix for the hypothetical package is straightforward, but I don't know how `GenericLinearAlgebra` could be fixed without the removal.

### DIFF
--- a/src/hessenberg.jl
+++ b/src/hessenberg.jl
@@ -463,15 +463,7 @@ end
 # preserve the wrapper for eigensolves with UpperHessenberg
 eigencopy_oftype(H::UpperHessenberg, S) = UpperHessenberg(eigencopy_oftype(H.data, S))
 
-# fallback to dense algorithms
-eigvals!(H::UpperHessenberg; permute::Bool=false, scale::Bool=true, sortby::Union{Function,Nothing}=eigsortby) =
-    eigvals!(triu!(H.data,-1); permute, scale, sortby)
-eigen!(H::UpperHessenberg; permute::Bool=false, scale::Bool=true, sortby::Union{Function,Nothing}=eigsortby) =
-    eigen!(triu!(H.data,-1); permute, scale, sortby)
-
-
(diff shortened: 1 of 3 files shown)


stevengj:
So, basically removing these is good for packages that commit type piracy, but potentially breaking for packages that don't? 😕 

But I guess there isn't much choice here, and GenericLinearAlgebra is pretty useful (and almost inevitable) given the lack of generic fallbacks for eigensolvers etc in LinearAlgebra.

araujoms:
Indeed.

I think the only sensible solution is to merge the required functionality from `GenericLinearAlgebra` and `GenericSchur` here. After all it's impossible to make generic eigensolvers without piracy.

I guess the current situation is a relic of the time `LinearAlgebra` was part of the basic image. But now that it's a separate package I think there's no problem growing it a little bit.

jishnub:
Perhaps we may forward the generic `eigvals!` method to an internal `_eigvals!` that we specialize here? This way, it'll be possible to specialize `eigvals!` in packages, and the fallback here would be specialized for matrix structure.

araujoms:
I think that would be a lot of work for little benefit: it doesn't solve the fundamental piracy problem, and would only be useful for a package that probably does not exist.

araujoms:
@nanosoldier `runtests()`

araujoms:
That's embarrassing, Nanosoldier gave me the cold shoulder.

jishnub:
I think I had misunderstood the objective of this PR, so please disregard my comment above.

araujoms:
Since #1557 was included in 1.13 this PR needs to be backported, otherwise GenericLinearAlgebra fails, see https://github.com/JuliaLinearAlgebra/GenericLinearAlgebra.jl/issues/180

PR #1603, opened 2026-05-15
Normalize eigvecs(::AbstractTriangular) to unit length

### LINKED ISSUE #1597: eigvecs(::AbstractTriangular) are not normalized to unit length

Mxyank24:
Previously, `eigvecs` for triangular matrices used LAPACK's `trevc`, which normalizes eigenvectors so that the largest component is ±1. This behavior is inconsistent with all other matrix types in Julia, which return unit-norm eigenvectors.

This PR fixes the issue by applying `eigvec_normalize!` to each eigenvector column in the `AbstractTriangular` dispatcher, bringing it into alignment with the convention used by `zgeevx` for general matrices.

Fixes #1597

### DIFF
--- a/src/eigen.jl
+++ b/src/eigen.jl
@@ -157,6 +157,7 @@ function eigvec_normalize!(v::AbstractVector)
     end
     return v
 end
+eigvec_normalize!(X::AbstractMatrix) = (foreach(eigvec_normalize!, eachcol(X)); X)
 
 """
     eigen!(A; permute, scale, sortby)
--- a/src/hessenberg.jl
+++ b/src/hessenberg.jl
@@ -438,8 +438,7 @@ function eigen!(H::UpperHessenberg{T, <:StridedMatrix{T}}; permute::Bool=false, 
     LAPACK.trevc!('R', 'B', BlasInt[], H.data, Z, Z) # set Z to right eigenvecs (for complex, see below)
     LAPACK.gebak!(scale ? 'S' : 'N', 'R', ilo, ihi, s, Z) # undo balancing
     if isreal(vals)
-        eigvec_normalize!(Z)
-        return Eigen(sorteig!(real(vals), Z, sortby)...)
+return Eigen(sorteig!(real(vals), eigvec_normalize!(Z), sortby)...)
     else # complex eigenvalues: real/imag eigenvec parts stored in consecutive cols of Z
         V = complex(Z)
         k = 1
--- a/src/hessenberg.jl
+++ b/src/hessenberg.jl
@@ -429,8 +429,7 @@ function eigen!(H::UpperHessenberg{T, <:StridedMatrix{T}}; permute::Bool=false, 
     _, Z, vals = LAPACK.hseqr!(H.data)
     LAPACK.trevc!('R', 'B', BlasInt[], H.data, Z, Z) # set Z to right eigenvecs
     LAPACK.gebak!(scale ? 'S' : 'N', 'R', ilo, ihi, s, Z) # undo balancing
-    eigvec_normalize!(Z)
-    return Eigen(sorteig!(vals, Z, sortby)...)
+return Eigen(sorteig!(vals, eigvec_normalize!(Z), sortby)...)
 end
 
 function eigen!(H::UpperHessenberg{T, <:StridedMatrix{T}}; permute::Bool=false, scale::Bool=true, sortby::Union{Function,Nothing}=eigsortby) where {T<:BlasReal}
--- a/src/hessenberg.jl
+++ b/src/hessenberg.jl
@@ -454,8 +454,7 @@ function eigen!(H::UpperHessenberg{T, <:StridedMatrix{T}}; permute::Bool=false, 
                 k += 2
             end
         end
-        eigvec_normalize!(V)
-        return Eigen(sorteig!(vals, V, sortby)...)
+return Eigen(sorteig!(vals, eigvec_normalize!(V), sortby)...)
     end
 end
 
--- a/src/hessenberg.jl
+++ b/src/hessenberg.jl
@@ -429,7 +429,7 @@ function eigen!(H::UpperHessenberg{T, <:StridedMatrix{T}}; permute::Bool=false, 
     _, Z, vals = LAPACK.hseqr!(H.data)
     LAPACK.trevc!('R', 'B', BlasInt[], H.data, Z, Z) # set Z to right eigenvecs
     LAPACK.gebak!(scale ? 'S' : 'N', 'R', ilo, ihi, s, Z) # undo balancing
-    foreach(eigvec_normalize!, eachcol(Z)) # normalize eigenvecs
+    eigvec_normalize!(Z)
     return Eigen(sorteig!(vals, Z, sortby)...)
 end
 
@@ -439,7 +439,7 @@ function eigen!(H::UpperHessenberg{T, <:StridedMatrix{T}}; permute::Bool=false, 
     LAPACK.trevc!('R', 'B', BlasInt[], H.data, Z, Z) # set Z to right eigenvecs (for complex, see below)
     LAPACK.gebak!(scale ? 'S' : 'N', 'R', ilo, ihi, s, Z) # undo balancing
     if isreal(vals)
-        foreach(eigvec_normalize!, eachcol(Z)) # normalize eigenvecs
+        eigvec_normalize!(Z)
         return Eigen(sorteig!(real(vals), Z, sortby)...)
     else # complex eigenvalues: real/imag eigenvec parts stored in consecutive cols of Z
         V = complex(Z)
@@ -455,7 +455,7 @@ function eigen!(H::UpperHessenberg{T, <:StridedMatrix{T}}; permute::Bool=false, 
                 k += 2
             end
         end
-        foreach(eigvec_normalize!, eachcol(V)) # normalize eigenvecs
+        eigvec_normalize!(V)
         return Eigen(sorteig!(vals, V, sortby)...)
     end
 end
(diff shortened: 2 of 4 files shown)


src/triangular.jl
```diff
@@ -1377,6 +1377,20 @@ end
...
+    normalize!(v)
+    maxabs2, k = findmax(abs2, v)
+    if eltype(v) <: Real
+        v[k] < 0 && (v .= .- v)
+    elseif maxabs2 > 0
+        v .*= conj(v[k]) / sqrt(maxabs2)
+        v[k] = real(v[k])
+    end
+    return v
+end
```
stevengj: This function is already defined in eigen.jl, so you don't need to redefine it here.

src/triangular.jl
```diff
@@ -2981,11 +2995,13 @@ end
 eigvals(A::AbstractTriangular) = diag(A)
 # fallback for unknown types
 function eigvecs(A::AbstractTriangular{<:BlasFloat})
-    if istriu(A)
+    vecs = if istriu(A)
         eigvecs(UpperTriangular(Matrix(A)))
     else # istril(A)
         eigvecs(LowerTriangular(Matrix(A)))
     end
+    foreach(eigvec_normalize!, eachcol(vecs))
```
stevengj: This call shouldn't be in this function, it should be in the lower-level methods — immediately after the calls to `trevc!`.

It would also be good to define a method
```jl
eigvec_normalize!(X::AbstractMatrix) = (foreach(eigvec_normalize!, eachcol(X)); X)
```
to encapsulate this `foreach` pattern.    Then you can just wrap `eigvec_normalize!` around the call to `trevc!`.

src/triangular.jl
Mxyank24: Updated

src/triangular.jl
Mxyank24: Also added eigvec_normalize!(X::AbstractMatrix) to eigen.jl to encapsulate the foreach pattern as suggested.

src/triangular.jl
araujoms: This is clearly an LLM-written comment.

Mxyank24: coz it is written in well polished manner ?sorry about that. I'll keep comments short going forward.

src/eigen.jl
```diff
@@ -157,6 +157,7 @@ function eigvec_normalize!(v::AbstractVector)
     end
     return v
 end
+eigvec_normalize!(X::AbstractMatrix) = (foreach(eigvec_normalize!, eachcol(X)); X)
```
stevengj, line 160: You should also call this new method from `hessenberg.jl` (replacing the explicit `foreach` call).

stevengj:
Needs tests (just check that `Diagonal(X'X) ≈ I` for the returned eigenvectors `X` in `tests/triangular.jl`).

src/eigen.jl
Mxyank24: Done.

Mxyank24:
> Needs tests (just check that `X'X ≈ I` for the returned eigenvectors `X` in `tests/triangular.jl`).

added.

src/hessenberg.jl
```diff
@@ -429,7 +429,7 @@ function eigen!(H::UpperHessenberg{T, <:StridedMatrix{T}}; permute::Bool=false,
     _, Z, vals = LAPACK.hseqr!(H.data)
     LAPACK.trevc!('R', 'B', BlasInt[], H.data, Z, Z) # set Z to right eigenvecs
     LAPACK.gebak!(scale ? 'S' : 'N', 'R', ilo, ihi, s, Z) # undo balancing
-    foreach(eigvec_normalize!, eachcol(Z)) # normalize eigenvecs
+    eigvec_normalize!(Z)
     return Eigen(sorteig!(vals, Z, sortby)...)
```
stevengj:
```suggestion
return Eigen(sorteig!(vals, eigvec_normalize!(Z), sortby)...)
```

src/hessenberg.jl
```diff
@@ -455,7 +455,7 @@ function eigen!(H::UpperHessenberg{T, <:StridedMatrix{T}}; permute::Bool=false,
                 k += 2
             end
         end
-        foreach(eigvec_normalize!, eachcol(V)) # normalize eigenvecs
+        eigvec_normalize!(V)
         return Eigen(sorteig!(vals, V, sortby)...)
```
stevengj:
```suggestion
return Eigen(sorteig!(vals, eigvec_normalize!(V), sortby)...)
```

src/hessenberg.jl
```diff
@@ -439,7 +439,7 @@ function eigen!(H::UpperHessenberg{T, <:StridedMatrix{T}}; permute::Bool=false,
     LAPACK.trevc!('R', 'B', BlasInt[], H.data, Z, Z) # set Z to right eigenvecs (for complex, see below)
     LAPACK.gebak!(scale ? 'S' : 'N', 'R', ilo, ihi, s, Z) # undo balancing
     if isreal(vals)
-        foreach(eigvec_normalize!, eachcol(Z)) # normalize eigenvecs
+        eigvec_normalize!(Z)
         return Eigen(sorteig!(real(vals), Z, sortby)...)
```
stevengj:
```suggestion
return Eigen(sorteig!(real(vals), eigvec_normalize!(Z), sortby)...)
```

src/triangular.jl
```diff
@@ -1377,23 +1377,24 @@ end
 
 # Eigensystems
 ## Notice that trecv works for quasi-triangular matrices and therefore the lower sub diagonal must be zeroed before calling the subroutine
+
```
dkarrasch:
```suggestion
```

src/hessenberg.jl
```diff
@@ -429,8 +429,7 @@ function eigen!(H::UpperHessenberg{T, <:StridedMatrix{T}}; permute::Bool=false,
     _, Z, vals = LAPACK.hseqr!(H.data)
     LAPACK.trevc!('R', 'B', BlasInt[], H.data, Z, Z) # set Z to right eigenvecs
     LAPACK.gebak!(scale ? 'S' : 'N', 'R', ilo, ihi, s, Z) # undo balancing
-    foreach(eigvec_normalize!, eachcol(Z)) # normalize eigenvecs
-    return Eigen(sorteig!(vals, Z, sortby)...)
+return Eigen(sorteig!(vals, eigvec_normalize!(Z), sortby)...)
```
dkarrasch:
```suggestion
    return Eigen(sorteig!(vals, eigvec_normalize!(Z), sortby)...)
```

src/hessenberg.jl
```diff
@@ -439,8 +438,7 @@ function eigen!(H::UpperHessenberg{T, <:StridedMatrix{T}}; permute::Bool=false,
     LAPACK.trevc!('R', 'B', BlasInt[], H.data, Z, Z) # set Z to right eigenvecs (for complex, see below)
     LAPACK.gebak!(scale ? 'S' : 'N', 'R', ilo, ihi, s, Z) # undo balancing
     if isreal(vals)
-        foreach(eigvec_normalize!, eachcol(Z)) # normalize eigenvecs
-        return Eigen(sorteig!(real(vals), Z, sortby)...)
+return Eigen(sorteig!(real(vals), eigvec_normalize!(Z), sortby)...)
```
dkarrasch:
```suggestion
        return Eigen(sorteig!(real(vals), eigvec_normalize!(Z), sortby)...)
```

src/hessenberg.jl
```diff
@@ -455,8 +453,7 @@ function eigen!(H::UpperHessenberg{T, <:StridedMatrix{T}}; permute::Bool=false,
                 k += 2
             end
         end
-        foreach(eigvec_normalize!, eachcol(V)) # normalize eigenvecs
-        return Eigen(sorteig!(vals, V, sortby)...)
+return Eigen(sorteig!(vals, eigvec_normalize!(V), sortby)...)
```
dkarrasch:
```suggestion
        return Eigen(sorteig!(vals, eigvec_normalize!(V), sortby)...)
```

stevengj (approved):

PR #1606, opened 2026-05-17
Extend `matop_dest` mechanism to `UpperHessenberg` and `Factorization`

dkarrasch:
I'm not sure if extending this to `Factorization` goes against the `[MAT]op_dest` naming, but this came up in the SparseArrays.jl PR: `\(S::SymTridiagional, b::SparseVector)` calls `ldlt(S) \ b`, so we want to allocate a dense output vector by the same mechanism as with matrices.

### DIFF
--- a/src/factorization.jl
+++ b/src/factorization.jl
@@ -135,8 +135,7 @@ end
 
 function (\)(F::Factorization, B::AbstractVecOrMat)
     require_one_based_indexing(B)
-    TFB = typeof(oneunit(eltype(F)) \ oneunit(eltype(B)))
-    ldiv!(F, copy_similar(B, TFB))
+    ldiv!(F, copyto!(matop_dest(\, F, B), B))
 end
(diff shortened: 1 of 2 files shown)


araujoms (approved):
I firmly believe vectors are n x 1 matrices.

dkarrasch:
> I firmly believe vectors are n x 1 matrices.

This is not so much an extension from MATrices to vectors, but rather to `Factorization`s. But I'll happily accept the approval. ;-)

After second thought, when we have `matop_dest` for operations including solves, including `Factorization`s in this machinery is nothing but consistent, at least to me. The fact that a call to `\` or `/` may pass through corresponding calls for factorizations is the nature of factorizations and an "implementation detail". The destination array determination should equally apply.

araujoms:
Ops, indeed, we were already using `matop_dest` with vectors. But yeah, I think `Factorization`s also make perfect sense.

PR #1608, opened 2026-05-17
make eigenvalues always sorted by default

araujoms:
Followup to #1477, #1595, and #1626: this PR changes the default sorting from `nothing` to `eigsortby` for `Diagonal` and *`Triangular` matrices. With this there are no exceptions remaining, and we can change the docstring to state that eigenvalues are always sorted by default.

In the `Diagonal` case I think it's a no-brainer: nobody calls `eigvals` intentionally on a diagonal matrix. This only happens in generic code, and in this case it can cause bugs.

In the *`Triangular` case it's not so clear-cut: people might very well call `eigvecs` on a triangular matrix, and the sorted output will be surprising and unhelpful. I think it's still worth it for consistency.

### DIFF
--- a/src/eigen.jl
+++ b/src/eigen.jl
@@ -217,8 +217,7 @@ make rows and columns more equal in norm. The default is `true` for both options
 
 By default, the eigenvalues and vectors are sorted lexicographically by `(real(λ),imag(λ))`.
 A different comparison function `by(λ)` can be passed to `sortby`, or you can pass
-`sortby=nothing` to leave the eigenvalues in an arbitrary order. Some special matrix types
-(e.g. [`Diagonal`](@ref)) may have a different default.
+`sortby=nothing` to leave the eigenvalues in an arbitrary order.
 
 # Examples
 ```jldoctest
--- a/test/diagonal.jl
+++ b/test/diagonal.jl
@@ -1068,6 +1068,7 @@ end
 
 @testset "eigenvalue sorting" begin
     D = Diagonal([0.4, 0.2, -1.3])
+    @test eigvals(D) == eigen(D).values == [-1.3, 0.2, 0.4] #sorted by default
     @test eigvals(D; sortby=nothing) == eigen(D; sortby=nothing).values == [0.4, 0.2, -1.3]
     @test eigvals(D; sortby=LinearAlgebra.eigsortby) == [-1.3, 0.2, 0.4] # sortby keyword supported for eigvals(::Diagonal)
     @test eigvals(Matrix(D)) == eigen(Matrix(D)).values == [-1.3, 0.2, 0.4] # sorted even if diagonal special case is detected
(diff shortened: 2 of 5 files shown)


test/diagonal.jl
```diff
@@ -1068,7 +1068,7 @@ end
 
 @testset "eigenvalue sorting" begin
     D = Diagonal([0.4, 0.2, -1.3])
-    @test eigvals(D) == eigen(D).values == [0.4, 0.2, -1.3] # not sorted by default
+    @test eigvals(D; sortby=nothing) == eigen(D; sortby=nothing).values == [0.4, 0.2, -1.3]
```
stevengj, line 1072: Add a test that `eigvals(D) == [-1.3, 0.2, 0.4]` ?

araujoms: done

src/eigen.jl
```diff
@@ -217,8 +217,7 @@ make rows and columns more equal in norm. The default is `true` for both options
 
 By default, the eigenvalues and vectors are sorted lexicographically by `(real(λ),imag(λ))`.
 A different comparison function `by(λ)` can be passed to `sortby`, or you can pass
-`sortby=nothing` to leave the eigenvalues in an arbitrary order. Some special matrix types
-(e.g. [`Diagonal`](@ref)) may have a different default.
+`sortby=nothing` to leave the eigenvalues in an arbitrary order.
```
jishnub, line 221: Is there any other matrix type that also needs to be checked?

araujoms: Argh, there is, `UpperTriangular` and `LowerTriangular` do not sort right now.

araujoms: Ok, fixed and rebased, now all matrix types defined in `LinearAlgebra` are checked.

stevengj:
> In the `*Triangular` case it's not so clear-cut: people might very well call eigvecs on a triangular matrix, and the sorted output will be surprising and unhelpful. I think it's still worth it for consistency.

The downside here is that the unsorted output $X$ of `eigvecs` is triangular, which allows you to quickly multiply by $X^{-1}$ (to put a vector in the diagonalized basis).   I wonder if anyone in the Julia ecosystem is using this?  I can't find any examples.

Hopefully it doesn't matter in practice, because diagonalizing triangular matrices is numerically unstable: the matrix can be defective or arbitrarily close to defective (= ill-conditioned eigenvector basis).  So it shouldn't be used much in serious applications, and casual applications probably don't worry about exploiting the triangular eigenvector matrix.

(In any case, I don't think we documented or guaranteed anything about the returned eigenvector ordering, so this does not break any semver guarantees.)

stevengj (approved):

araujoms:
I find it unlikely that anybody would be using it for serious applications; I had in mind students calling `eigvecs` and expecting to get a triangular matrix.

If anyone actually needs it it's easy enough to pass `sortby=nothing`.

stevengj:
Maybe a NEWS item for these `sortby` changes?

araujoms:
You mean a PR to the NEWS.md file in the julia repository, as opposed to a PR to the HISTORY.md file in this repository?

PR #1610, opened 2026-05-18
Pass through `nothing` strides

nhz2:
Companion PR to https://github.com/JuliaLang/julia/pull/61807

That PR changes `strides` to return `nothing` on non-strided arrays instead of throwing an error.

This PR changes `strides` of `Adjoint` and `Transpose` to pass through `nothing` if `strides(parent)` is `nothing`.

This PR should be merged before https://github.com/JuliaLang/julia/pull/61807 is.

### DIFF
--- a/src/adjtrans.jl
+++ b/src/adjtrans.jl
@@ -368,11 +368,27 @@ convert(::Type{Adjoint{T,S}}, A::Adjoint) where {T,S} = Adjoint{T,S}(convert(S, 
 convert(::Type{Transpose{T,S}}, A::Transpose) where {T,S} = Transpose{T,S}(convert(S, A.parent))::Transpose{T,S}
 
 # Strides and pointer for transposed strided arrays — but only if the elements are actually stored in memory
-Base.strides(A::Adjoint{<:Real, <:AbstractVector}) = (stride(A.parent, 2), stride(A.parent, 1))
-Base.strides(A::Transpose{<:Any, <:AbstractVector}) = (stride(A.parent, 2), stride(A.parent, 1))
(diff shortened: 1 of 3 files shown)


src/adjtrans.jl
```diff
@@ -368,11 +368,27 @@ convert(::Type{Adjoint{T,S}}, A::Adjoint) where {T,S} = Adjoint{T,S}(convert(S,
 convert(::Type{Transpose{T,S}}, A::Transpose) where {T,S} = Transpose{T,S}(convert(S, A.parent))::Transpose{T,S}
 
 # Strides and pointer for transposed strided arrays — but only if the elements are actually stored in memory
-Base.strides(A::Adjoint{<:Real, <:AbstractVector}) = (stride(A.parent, 2), stride(A.parent, 1))
-Base.strides(A::Transpose{<:Any, <:AbstractVector}) = (stride(A.parent, 2), stride(A.parent, 1))
+function Base.strides(A::Adjoint{<:Real, <:AbstractVector})
+    st = strides(A.parent)
+    isnothing(st) && return nothing
+    (st[1]*Int(length(A.parent)), st[1])
```
nhz2, line 374: `st[1]*Int(length(A.parent))` should match `stride(A.parent, 2)` since the parent is an abstract vector: https://github.com/JuliaLang/julia/blob/201e056e9ee2398f01ecd2f104f9d8000b04d195/base/abstractarray.jl#L602

https://github.com/JuliaLang/julia/pull/30432 has this return `(1, st[1])`. The first stride should not get used, so in theory its value shouldn't matter.

stevengj:
Needs to wait on whether https://github.com/JuliaLang/julia/pull/61807 is accepted.

PR #1615, opened 2026-05-18
rm trailing whitespace

stevengj:
I noticed that some files still have trailing whitespace, ~~which I believe we no longer allow in new commits~~ *which is not currently checked for in PRs (see below)*, and as a result I was seeing extraneous diffs in my PRs due to my editor deleting trailing whitespace when saving a file.

This PR just deletes all trailing whitespace from `.jl` and `.md` files.


dkarrasch:
I don't think we have a specific CI step to check for that. So files with trailing whitespace can be merged accidentally.

DilumAluthge:
Could be a nice CI check to add. `good first issue` maybe? Just copy the check over from JuliaLang/julia?

stevengj:
For reference, the whitespace check is in the julia `Makefile` here: https://github.com/JuliaLang/julia/blob/b616c52ecdd2c4b7dc74ab809c4ed42a789d1651/Makefile#L162-L176 running [`contrib/check-whitespace.jl`](https://github.com/JuliaLang/julia/blob/b616c52ecdd2c4b7dc74ab809c4ed42a789d1651/contrib/check-whitespace.jl), called from https://github.com/JuliaLang/julia/blob/b616c52ecdd2c4b7dc74ab809c4ed42a789d1651/.github/workflows/Whitespace.yml#L23

PR #1620, opened 2026-05-19
fix: zero(::UnitUpperTriangular/UnitLowerTriangular) 

### LINKED ISSUE #1613: zero(::UnitUpperTriangular) throws
```jl
julia> using LinearAlgebra

julia> zero(UnitUpperTriangular(rand(5,5)))
ERROR: ArgumentError: cannot set index (1, 1) on the diagonal of a UnitUpperTriangular matrix to a non-unit value (0.0)
```
Seems like this should return an `UpperTriangular` matrix of zeros?

Similarly for `UnitLowerTriangular`.

Usually, we try to have `zero` return the same type, but that's not always the case (e.g. `zero(1:2)` returns a `Vector`), and here it is impossible.  (See also #1050.)

Mxyank24:
Fixes #1613, part of #1050.

Calling `zero` on `UnitUpperTriangular` or `UnitLowerTriangular` was throwing:
ERROR: ArgumentError: cannot set index (1, 1) on the diagonal of a UnitUpperTriangular matrix to a non-unit value (0.0)
The diagonal is locked to 1 in these types, so creating a zero matrix while preserving the type is impossible. Instead I return `UpperTriangular`/`LowerTriangular` — same idea as `zero(1:2)` returning a `Vector` instead of a `Range`.
Also handled `UpperTriangular` and `LowerTriangular` to forward to `zero(parent(A))`.

### DIFF
--- a/src/triangular.jl
+++ b/src/triangular.jl
@@ -1044,8 +1044,8 @@ end
 +(A::UpperOrLowerTriangular, B::UpperOrLowerTriangular) = full(A) + full(B)
 +(A::AbstractTriangular, B::AbstractTriangular) = copyto!(similar(parent(A), size(A)), A) + copyto!(similar(parent(B), size(B)), B)
 
-zero(A::UpperTriangular{T,<:StridedMatrix}) where {T} = fill!(similar(A, T), zero(T)))
-zero(A::LowerTriangular{T,<:StridedMatrix}) where {T} = fill!(similar(A, T), zero(T)))
+zero(A::UpperTriangular{T,<:StridedMatrix}) where {T} = fill!(similar(A, T), zero(T))
+zero(A::LowerTriangular{T,<:StridedMatrix}) where {T} = fill!(similar(A, T), zero(T))
 zero(A::UnitUpperTriangular) = zero(UpperTriangular(parent(A)))
 zero(A::UnitLowerTriangular) = zero(LowerTriangular(parent(A)))
 
--- a/src/triangular.jl
+++ b/src/triangular.jl
@@ -1044,10 +1044,10 @@ end
 +(A::UpperOrLowerTriangular, B::UpperOrLowerTriangular) = full(A) + full(B)
 +(A::AbstractTriangular, B::AbstractTriangular) = copyto!(similar(parent(A), size(A)), A) + copyto!(similar(parent(B), size(B)), B)
 
-zero(A::UpperTriangular) = UpperTriangular(zero(parent(A)))
-zero(A::LowerTriangular) = LowerTriangular(zero(parent(A)))
-zero(A::UnitUpperTriangular) = UpperTriangular(zero(parent(A)))
-zero(A::UnitLowerTriangular) = LowerTriangular(zero(parent(A)))
+zero(A::UpperTriangular{T,<:StridedMatrix}) where {T} = UpperTriangular(fill!(similar(parent(A), T), zero(T)))
+zero(A::LowerTriangular{T,<:StridedMatrix}) where {T} = LowerTriangular(fill!(similar(parent(A), T), zero(T)))
+zero(A::UnitUpperTriangular) = zero(UpperTriangular(parent(A)))
+zero(A::UnitLowerTriangular) = zero(LowerTriangular(parent(A)))
 
 function -(A::UpperTriangular, B::UpperTriangular)
     (parent(A) isa StridedMatrix || parent(B) isa StridedMatrix) && return A .- B
--- a/src/triangular.jl
+++ b/src/triangular.jl
@@ -1046,8 +1046,8 @@ end
 
 zero(A::UpperTriangular) = UpperTriangular(zero(parent(A)))
 zero(A::LowerTriangular) = LowerTriangular(zero(parent(A)))
-zero(A::UpperTriangular{T,<:StridedMatrix}) where {T<:Number} = fill!(similar(A, T), zero(T))
-zero(A::LowerTriangular{T,<:StridedMatrix}) where {T<:Number} = fill!(similar(A, T), zero(T))
+zero(A::UpperTriangular{T,<:StridedMatrix}) where {T<:Number} = fill!(similar(A, typeof(zero(T))), zero(T))
+zero(A::LowerTriangular{T,<:StridedMatrix}) where {T<:Number} = fill!(similar(A, typeof(zero(T)))), zero(T))
 zero(A::UnitUpperTriangular) = zero(UpperTriangular(parent(A)))
 zero(A::UnitLowerTriangular) = zero(LowerTriangular(parent(A)))
 
--- a/src/triangular.jl
+++ b/src/triangular.jl
@@ -1044,6 +1044,11 @@ end
 +(A::UpperOrLowerTriangular, B::UpperOrLowerTriangular) = full(A) + full(B)
 +(A::AbstractTriangular, B::AbstractTriangular) = copyto!(similar(parent(A), size(A)), A) + copyto!(similar(parent(B), size(B)), B)
 
+zero(A::UpperTriangular) = UpperTriangular(zero(parent(A)))
+zero(A::LowerTriangular) = LowerTriangular(zero(parent(A)))
+zero(A::UnitUpperTriangular) = UpperTriangular(zero(parent(A)))
+zero(A::UnitLowerTriangular) = LowerTriangular(zero(parent(A)))
+
 function -(A::UpperTriangular, B::UpperTriangular)
     (parent(A) isa StridedMatrix || parent(B) isa StridedMatrix) && return A .- B
     UpperTriangular(A.data - B.data)
--- a/src/triangular.jl
+++ b/src/triangular.jl
@@ -1044,8 +1044,6 @@ end
 +(A::UpperOrLowerTriangular, B::UpperOrLowerTriangular) = full(A) + full(B)
 +(A::AbstractTriangular, B::AbstractTriangular) = copyto!(similar(parent(A), size(A)), A) + copyto!(similar(parent(B), size(B)), B)
 
-zero(A::UpperTriangular{T,<:StridedMatrix}) where {T<:Number} = fill!(similar(A, typeof(zero(T))), zero(T))
-zero(A::LowerTriangular{T,<:StridedMatrix}) where {T<:Number} = fill!(similar(A, typeof(zero(T))), zero(T))
 zero(A::UpperTriangular) = UpperTriangular(zero(parent(A)))
 zero(A::LowerTriangular) = LowerTriangular(zero(parent(A)))
 zero(A::UnitUpperTriangular) = zero(UpperTriangular(parent(A)))
--- a/src/triangular.jl
+++ b/src/triangular.jl
@@ -1044,10 +1044,10 @@ end
 +(A::UpperOrLowerTriangular, B::UpperOrLowerTriangular) = full(A) + full(B)
 +(A::AbstractTriangular, B::AbstractTriangular) = copyto!(similar(parent(A), size(A)), A) + copyto!(similar(parent(B), size(B)), B)
 
+zero(A::UpperTriangular{T,<:StridedMatrix}) where {T<:Number} = fill!(similar(A, typeof(zero(T))), zero(T))
+zero(A::LowerTriangular{T,<:StridedMatrix}) where {T<:Number} = fill!(similar(A, typeof(zero(T))), zero(T))
 zero(A::UpperTriangular) = UpperTriangular(zero(parent(A)))
 zero(A::LowerTriangular) = LowerTriangular(zero(parent(A)))
-zero(A::UpperTriangular{T,<:StridedMatrix}) where {T<:Number} = fill!(similar(A, typeof(zero(T))), zero(T))
-zero(A::LowerTriangular{T,<:StridedMatrix}) where {T<:Number} = fill!(similar(A, typeof(zero(T)))), zero(T))
 zero(A::UnitUpperTriangular) = zero(UpperTriangular(parent(A)))
 zero(A::UnitLowerTriangular) = zero(LowerTriangular(parent(A)))
 
--- a/src/triangular.jl
+++ b/src/triangular.jl
@@ -1044,8 +1044,10 @@ end
 +(A::UpperOrLowerTriangular, B::UpperOrLowerTriangular) = full(A) + full(B)
 +(A::AbstractTriangular, B::AbstractTriangular) = copyto!(similar(parent(A), size(A)), A) + copyto!(similar(parent(B), size(B)), B)
 
-zero(A::UpperTriangular{T,<:StridedMatrix}) where {T} = fill!(similar(A, T), zero(T))
-zero(A::LowerTriangular{T,<:StridedMatrix}) where {T} = fill!(similar(A, T), zero(T))
+zero(A::UpperTriangular) = UpperTriangular(zero(parent(A)))
+zero(A::LowerTriangular) = LowerTriangular(zero(parent(A)))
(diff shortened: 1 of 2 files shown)


src/triangular.jl
```diff
@@ -1044,6 +1044,11 @@ end
 +(A::UpperOrLowerTriangular, B::UpperOrLowerTriangular) = full(A) + full(B)
 +(A::AbstractTriangular, B::AbstractTriangular) = copyto!(similar(parent(A), size(A)), A) + copyto!(similar(parent(B), size(B)), B)
 
+zero(A::UpperTriangular) = UpperTriangular(zero(parent(A)))
+zero(A::LowerTriangular) = LowerTriangular(zero(parent(A)))
+zero(A::UnitUpperTriangular) = UpperTriangular(zero(parent(A)))
+zero(A::UnitLowerTriangular) = LowerTriangular(zero(parent(A)))
```
dkarrasch: With wrappers which use effectively only one half of the underlying matrix, we should make sure that only that half is filled with zeros, at least in the strided matrix case. So, I think we would also want to have something like

```julia
zero(A::UpperTriangular{T,<:StridedMatrix}) where {T} = fill!(similar(parent(A), T), zero(T))
```

As for the `Unit*Triangular` cases, we could actually just rewrap and use the non-unit analogues:

```julia
zero(A::UnitUpperTriangular) = zero(UpperTriangular(parent(A)))
```

Then the other methods would distinguish the strided from the non-strided cases.

dkarrasch (commented):
Thanks for picking this up! While developing, could you please keep an eye on runtime as well? Like measure the time for `zero(U)` for a 2000x2000 triangular matrix `U`. We should make sure that we don't regress.

As for the strided case, I think we need a corresponding PR for the symmetric case that you dealt with recently.

src/triangular.jl
Mxyank24: got 378 μs (3 allocations: 30.53 MiB) for 2000x2000, seems fine.

Mxyank24:
> Thanks for picking this up! While developing, could you please keep an eye on runtime as well? Like measure the time for `zero(U)` for a 2000x2000 triangular matrix `U`. We should make sure that we don't regress.
> 
> As for the strided case, I think we need a corresponding PR for the symmetric case that you dealt with recently.

will open a PR for the symmetric case as well!

src/triangular.jl
dkarrasch: And how much does it take on v1.12?

src/triangular.jl
```diff
@@ -1044,6 +1044,11 @@ end
 +(A::UpperOrLowerTriangular, B::UpperOrLowerTriangular) = full(A) + full(B)
 +(A::AbstractTriangular, B::AbstractTriangular) = copyto!(similar(parent(A), size(A)), A) + copyto!(similar(parent(B), size(B)), B)
 
+zero(A::UpperTriangular{T,<:StridedMatrix}) where {T} = UpperTriangular(fill!(similar(parent(A), T), zero(T)))
+zero(A::LowerTriangular{T,<:StridedMatrix}) where {T} = LowerTriangular(fill!(similar(parent(A), T), zero(T)))
```
dkarrasch: If we call 2-arg `similar`, then an `*Triangular` is created on top of the similar parent. Then `fill!` understands that is needs to fill only one half.

```suggestion
zero(A::UpperTriangular{T,<:StridedMatrix}) where {T} = fill!(similar(A, T), zero(T))
zero(A::LowerTriangular{T,<:StridedMatrix}) where {T} = fill!(similar(A, T), zero(T))
```

dkarrasch: Otherwise you'll again write the whole parent array, which we want to avoid in the strided case.

src/triangular.jl
```diff
@@ -1044,6 +1044,11 @@ end
 +(A::UpperOrLowerTriangular, B::UpperOrLowerTriangular) = full(A) + full(B)
 +(A::AbstractTriangular, B::AbstractTriangular) = copyto!(similar(parent(A), size(A)), A) + copyto!(similar(parent(B), size(B)), B)
 
+zero(A::UpperTriangular{T,<:StridedMatrix}) where {T} = fill!(similar(A, T), zero(T))
+zero(A::LowerTriangular{T,<:StridedMatrix}) where {T} = fill!(similar(A, T), zero(T))
```
dkarrasch: You will want to keep the general ones and restrict the `fill!` version to number eltypes (sorry, my bad):

```suggestion
zero(A::UpperTriangular) = UpperTriangular(zero(parent(A)))
zero(A::LowerTriangular) = LowerTriangular(zero(parent(A)))
zero(A::UpperTriangular{T,<:StridedMatrix}) where {T<:Number} = fill!(similar(A, T), zero(T))
zero(A::LowerTriangular{T,<:StridedMatrix}) where {T<:Number} = fill!(similar(A, T), zero(T))
```

src/triangular.jl
Mxyank24: 298 μs

Mxyank24:
> Thanks for picking this up! While developing, could you please keep an eye on runtime as well? Like measure the time for `zero(U)` for a 2000x2000 triangular matrix `U`. We should make sure that we don't regress.
> 
> As for the strided case, I think we need a corresponding PR for the symmetric case that you dealt with recently.

#1621

src/triangular.jl
```diff
@@ -1044,6 +1044,13 @@ end
 +(A::UpperOrLowerTriangular, B::UpperOrLowerTriangular) = full(A) + full(B)
 +(A::AbstractTriangular, B::AbstractTriangular) = copyto!(similar(parent(A), size(A)), A) + copyto!(similar(parent(B), size(B)), B)
 
+zero(A::UpperTriangular) = UpperTriangular(zero(parent(A)))
+zero(A::LowerTriangular) = LowerTriangular(zero(parent(A)))
+zero(A::UpperTriangular{T,<:StridedMatrix}) where {T<:Number} = fill!(similar(A, T), zero(T))
+zero(A::LowerTriangular{T,<:StridedMatrix}) where {T<:Number} = fill!(similar(A, T), zero(T))
```
dkarrasch: I realized another thing (sorry for the back and forth): in the current `abstractarray.jl` method, we use `typeof(zero(T))` for the eltype of the array. In some cases, this may be different from `T`, and we should take that into account here as well.

```suggestion
zero(A::UpperTriangular{T,<:StridedMatrix}) where {T<:Number} = fill!(similar(A, typeof(zero(T))), zero(T))
zero(A::LowerTriangular{T,<:StridedMatrix}) where {T<:Number} = fill!(similar(A, typeof(zero(T)))), zero(T))
```

dkarrasch: This applies equally well to the symmetric case you're working on.

Mxyank24: ok sir

dkarrasch:
Could you please also add tests in the style of the `ZeroTestMatrix` you had in the other PR. We need tests for the non-strided case and the fact that `zero` is forwarded to the parent, a la StaticArray.

src/triangular.jl
```diff
@@ -1044,6 +1044,13 @@ end
 +(A::UpperOrLowerTriangular, B::UpperOrLowerTriangular) = full(A) + full(B)
 +(A::AbstractTriangular, B::AbstractTriangular) = copyto!(similar(parent(A), size(A)), A) + copyto!(similar(parent(B), size(B)), B)
 
+zero(A::UpperTriangular{T,<:StridedMatrix}) where {T<:Number} = fill!(similar(A, typeof(zero(T))), zero(T))
+zero(A::LowerTriangular{T,<:StridedMatrix}) where {T<:Number} = fill!(similar(A, typeof(zero(T))), zero(T))
```
stevengj: These `StridedMatrix` methods seem redundant?

dkarrasch: I don't think so. They are meant to overload the next two lines, which forward the `zero` call to the parent. In the strided case, we don't want to zero out the entire underlying matrix.

Mxyank24:
> Could you please also add tests in the style of the `ZeroTestMatrix` you had in the other PR. We need tests for the non-strided case and the fact that `zero` is forwarded to the parent, a la StaticArray.

added sir

src/triangular.jl
stevengj: > In the strided case, we don't want to zero out the entire underlying matrix.

Why?  Zero-ing the entire matrix is probably faster.  e.g.
```jl
julia> using BenchmarkTools, LinearAlgebra

julia> A = rand(1000,1000); T = UpperTriangular(A);

julia> @btime zero($A);
  70.541 μs (3 allocations: 7.64 MiB)

julia> @btime fill!(similar($A), 0.0);
  70.541 μs (3 allocations: 7.64 MiB)

julia> @btime fill!(similar($T), 0.0);
  115.500 μs (3 allocations: 7.64 MiB)
```
(and if anything it seems a little nicer not to leave uninitialized memory in `parent(zero(T))`.)

dkarrasch: I noticed there is not much of a difference in runtime. The last benchmark call should internally dispatch to `fillbanded!(...)`, which operates on the parent and fills only half of the storage matrix. But I'm surprised this is even slower than filling the entire matrix. If we're not overlooking anything, then this would mean that simple forwarding to the parent is in any case the way to go.

@jishnub Thoughts?

dkarrasch: I can reproduce your timings on v1.12, but on v1.13-rc1 I get:

```julia
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.13.0-rc1 (2026-04-27)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org release
|__/                   |

julia> using BenchmarkTools, LinearAlgebra
[ Info: Precompiling BenchmarkTools [6e4b80f9-dd63-53aa-95a3-0cdb28fa8baf] (caches not reused: 4 for different Julia build configuration)
Precompiling BenchmarkTools finished.
  8 dependencies successfully precompiled in 12 seconds. 9 already precompiled.

julia> using BenchmarkTools, LinearAlgebra

julia> A = rand(1000,1000); T = UpperTriangular(A);

julia> @btime zero($A);
  426.245 μs (3 allocations: 7.63 MiB)

julia> @btime fill!(similar($A), 0.0);
  411.595 μs (3 allocations: 7.63 MiB)

julia> @btime fill!(similar($T), 0.0);
  239.872 μs (3 allocations: 7.63 MiB)

julia> import Base: zero

julia> @btime zero($T);
  199.901 μs (3 allocations: 7.63 MiB)

julia> zero(U::UpperTriangular) = UpperTriangular(zero(parent(U)))
zero (generic function with 25 methods)

julia> @btime zero($T);
  441.636 μs (3 allocations: 7.63 MiB)
```

BTW, the `zero(A)` call has a major regression on my machine, it's btime is 174.486 μs on v1.12.

The uninitialized memory is already the status quo, at least since v1.12.

stevengj: Ideally the compiler should be able to optimize the `fill(A, 0.0)` to something equivalent to a `memset`.  In particular, if I define:
```jl
unsafe_zero(A) = (X = similar(A); GC.@preserve X Libc.memset(pointer(X), Cint(0), sizeof(X)); X)
```
then `@btime unsafe_zero($A)` has the same performace as `zero(A)` and `fill!(similar(A), 0.0)` in the benchmark above with Julia 1.12.

Whereas calling `fill!` on an `UpperTriangular` matrix has a much less regular memory-access pattern, even though it does half as many writes.

jishnub: Some of these relative benchmarks are dependent on the matrix sizes. At small sizes, the simd impact from contiguous memory access dominates, whereas at large sizes memory access becomes more of a bottleneck (I think). For example, for a 10x larger matrix, zeroing the triangular matrix becomes faster.
```julia
julia> A = rand(10000,10000); T = UpperTriangular(A);

julia> @btime zero($A);
  254.802 ms (252651 allocations: 776.50 MiB)

julia> @btime zero($T);
  151.475 ms (81152 allocations: 766.95 MiB)

julia> @btime fill!(similar($T), 0.0);
  136.981 ms (79338 allocations: 766.88 MiB)

julia> VERSION
v"1.12.4"
```

stevengj: So, here we have a situation where the relative cost of `zero(T)` vs. `zero(A)` depends on the size of the matrix and the type of the elements (and possibly the Julia version, though the performance regression mentioned above should really get fixed).

One option is a poly-algorithm, or further optimization.  But `zero(::UpperTriangular)` hardly seems worth the effort — it's hard to imagine a real application in which even a factor of 2 in performance for this function will be important.

Another option is to err on the side of less code, in which case we could simply delete the `StridedMatrix` methods and accept that it may occasionally be suboptimal by a modest factor (a factor of 2 at worst).

stevengj: @dkarrasch, can you file an issue for the performance regression in Julia 1.13?

dkarrasch: > @dkarrasch, can you file an issue for the performance regression in Julia 1.13?

I already did: https://github.com/JuliaLang/julia/issues/61870.

Given the uncertainty in the many circumstances, I think I agree with your proposal of "less code".

dkarrasch: The performance regression has been fixed.

dkarrasch:
@Mxyank24 Once again, sorry for the back and forth, but as you may have noticed, in a collective review, we came to the conclusion to not specialize for strided arrays. Instead, we'll just go with the "forward `zero` to the parent" policy. So, please remove those methods here, and if you like, remove the corresponding symmetric methods in another PR.

And please, don't respond so formally ("sir"). I am no authority here, just one member of the group of contributors.

src/triangular.jl
```suggestion
```
i.e. this is the suggested change: delete these specialized methods, which complicate the code for dubious practical benefit.

Mxyank24: > i.e. this is the suggested change: delete these specialized methods, which complicate the code for dubious practical benefit.

yes you're right

stevengj (approved):

test/triangular.jl
```diff
@@ -1113,6 +1113,34 @@ end
     end
 end
 
+<<<<<<< fix/zero-unit-triangular
```
stevengj: broken merge

test/triangular.jl
```diff
@@ -1113,6 +1113,34 @@ end
...
+    Base.size(A::ZeroTestWrapTri) = size(A.parent)
+    Base.getindex(A::ZeroTestWrapTri, i, j) = A.parent[i,j]
+    Base.zero(A::ZeroTestWrapTri) = ZeroTestWrapTri(zero(A.parent))
+
+    for T in (UpperTriangular, LowerTriangular)
+        Z = zero(T(ZeroTestWrapTri([1.0 2.0; 3.0 4.0])))
+        @test Z isa T
+        @test parent(Z) isa ZeroTestWrapTri
+        @test iszero(Z)
+=======
```
dkarrasch:
```suggestion
    end
end

```

test/triangular.jl
```diff
@@ -1125,6 +1153,7 @@ end
             @test F.values ≈ eigvals(B; sortby = LinearAlgebra.eigsortby)
             @test F.vectors ≈ eigvecs(B; sortby = LinearAlgebra.eigsortby)
         end
+>>>>>>> master
```
dkarrasch:
```suggestion
```

test/triangular.jl
stevengj:
```suggestion
```

stevengj (approved):

Mxyank24:
thanks a lot @dkarrasch @stevengj i learned a lot from the review process!.

PR #1621, opened 2026-05-19
fix: zero(::Symmetric/Hermitian) use fill! for strided case

Mxyank24:
Part of #1592
For strided parents, the previous implementation wrote to the whole parent array. This uses fill! on similar(A, T) instead, so only the relevant half gets filled with zeros.

The non-strided fallback still forwards to zero(parent(A)).

### DIFF
--- a/src/symmetric.jl
+++ b/src/symmetric.jl
@@ -222,7 +222,7 @@ for (S, H) in ((:Symmetric, :Hermitian), (:Hermitian, :Symmetric))
     end
 end
 # zero should forward to the parent array type
-zero(A::Union{Symmetric{T,<:StridedMatrix},Hermitian{T,<:StridedMatrix}}) where {T<:Number} = fill!(similar(A, T), zero(T))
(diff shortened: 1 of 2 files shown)


src/symmetric.jl
```diff
@@ -222,6 +222,7 @@ for (S, H) in ((:Symmetric, :Hermitian), (:Hermitian, :Symmetric))
     end
 end
 # zero should forward to the parent array type
+zero(A::Union{Symmetric{T,<:StridedMatrix},Hermitian{T,<:StridedMatrix}}) where {T} = fill!(similar(A, T), zero(T))
```
dkarrasch: Let's make sure this applies only to eltypes for which `zero` is certainly defined. This covers the standard cases.

```suggestion
zero(A::Union{Symmetric{T,<:StridedMatrix},Hermitian{T,<:StridedMatrix}}) where {T<:Number} = fill!(similar(A, T), zero(T))
```

PR #1625, opened 2026-05-21
Array shape check in `isapprox`

### LINKED ISSUE #1624: Should `isapprox` check the shape of input arrays?
Currently, when calling `isapprox` on two arrays of different shapes, instead of directly returning `false`, a `DimensionMismatch` error is raised:

```julia
julia> isapprox(rand(3, 4), rand(5, 6))
ERROR: DimensionMismatch: a has size (3, 4), b has size (5, 6), mismatch at dim 1
```

This is because:

https://github.com/JuliaLang/LinearAlgebra.jl/blob/035c8b4874c5847c546f6821977a14980e0fd25d/src/generic.jl#L2003-L2007

before calculating the norm of `x - y`, there is no check on the shape of input arrays. On the contrary, in `Base.:(==)`, such a check is performed at the beginning:

```julia
function (==)(A::AbstractArray, B::AbstractArray)
    if axes(A) != axes(B)
        return false
    end
```

https://github.com/JuliaLang/julia/blob/062a90bc8c2e393cddc52de58d1b373645ea88ee/base/abstractarray.jl#L3126-L3129

May I ask if the omission in `isapprox` is intentional, and should it be added?

Yue-Zhengyuan:
This PR adds an array shape check in `isapprox`, so that it directly returns `false` when the two arrays to be compared have different shapes.

Closes #1624.

### DIFF
--- a/src/generic.jl
+++ b/src/generic.jl
@@ -2004,8 +2004,16 @@ function isapprox(x::AbstractArray, y::AbstractArray;
     atol::Real=0,
     rtol::Real=Base.rtoldefault(promote_leaf_eltypes(x),promote_leaf_eltypes(y),atol),
     nans::Bool=false, norm::Function=norm)
-    _nested_axes(x) == _nested_axes(y) || return false
-    d = norm_x_minus_y(x, y)
+    local d
+    try
+        d = norm_x_minus_y(x, y)
+    catch e
+        if isa(e, DimensionMismatch)
+            return false
+        else
+            rethrow()
+        end
+    end
     if isfinite(d)
         return iszero(rtol) ? d <= atol : d <= max(atol, rtol*max(norm(x), norm(y)))
     else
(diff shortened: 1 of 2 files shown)


Yue-Zhengyuan:
There are some limitations on nested arrays: a simple check of the overall `axes` cannot catch all cases. For example:

```julia
a = [[1,2, [3,4]], 5.0, [6im, [7.0, 8.0]]]
b = [[1, [3,4]], 5.0, [6im, [7.0, 8.0]]]
```

In this case both `axes(a)` and `axes(b)` evaluate to `(Base.OneTo(3),)`, but the sub-arrays still have different shapes.

Yue-Zhengyuan:
To handle arbitrary nested arrays, it seems that I need to introduce an auxiliary function:

```julia
function _nested_axes(arr::AbstractArray)
    # Base case: If the array contains no sub-arrays, return its shape
    any(x -> x isa AbstractArray, arr) || return axes(arr)
    # Recursive step: Traverse the elements, recursing if they are arrays
    return map(x -> x isa AbstractArray ? _nested_axes(x) : axes(x), arr)
end
```

and check `_nested_axes(a) == _nested_axes(b)`.

dkarrasch:
Perhaps help a little the standard case with multiple dispatch?

```julia
_nested_axes(arr::AbstractArray{<:Number}) = axes(arr)
```

Yue-Zhengyuan (commented):
I notice some test errors on `_nested_axes` related to UniformScaling (in `test/bidiag.jl`, line 971-991 testset `"rmul!/lmul! with banded matrices" -> "non-commutative"`), but I don't quite understand how these arise, since they are not AbstractArrays at all.

Yue-Zhengyuan:
Alternatively, will it be better just use try-catch on calculating `norm(x-y)`? Then return false only when a DimensionMismatch is caught. This way I don't need to worry about all the weird edge cases. Or I just give up if using try-catch is discouraged 🫤

stevengj:
try-catch seems a lot simpler and more robust here.

PR #1627, opened 2026-05-22
simplify sortby interface

araujoms:
I think it's confusing to pass `sortby = nothing` when we want the eigenvalues to be sorted: the fact that LAPACK already sorts the eigenvalues for Hermitian matrices is an implementation detail. I've passed then `sortby = eigsortby`, and optimized this to a non-sort only in the same function where LAPACK is called.

This has the nice consequence of avoiding a useless sort when the user passes `sortby = eigsortby`.

### DIFF
--- a/src/symmetriceigen.jl
+++ b/src/symmetriceigen.jl
@@ -20,7 +20,7 @@ default_eigen_alg(@nospecialize(A)) = RobustRepresentations()
 
 # Eigensolvers for symmetric and Hermitian matrices
 function eigen!(A::RealHermSymComplexHerm{<:BlasReal,<:StridedMatrix}; alg::Algorithm = default_eigen_alg(A), sortby::Union{Function,Nothing}=eigsortby)
-    sortby2 = sortby == eigsortby ? nothing : sortby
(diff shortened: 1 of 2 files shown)


src/symmetriceigen.jl
```diff
@@ -19,13 +19,14 @@ Defaults to `LinearAlegbra.RobustRepresentations()`, which corresponds to the LA
 default_eigen_alg(@nospecialize(A)) = RobustRepresentations()
 
 # Eigensolvers for symmetric and Hermitian matrices
-function eigen!(A::RealHermSymComplexHerm{<:BlasReal,<:StridedMatrix}; alg::Algorithm = default_eigen_alg(A), sortby::Union{Function,Nothing}=nothing)
+function eigen!(A::RealHermSymComplexHerm{<:BlasReal,<:StridedMatrix}; alg::Algorithm = default_eigen_alg(A), sortby::Union{Function,Nothing}=eigsortby)
+    sortby2 = sortby == eigsortby ? nothing : sortby
```
stevengj:
```suggestion
    sortby2 = sortby == eigsortby ? nothing : sortby # eigsortby is done by LAPACK.syev*
```

stevengj (approved):

PR #1628, opened 2026-05-22
mini-optimization of eigmax/eigmin

araujoms:
Now that `sortby` is accepted everywhere, we don't need to sort the vector before maximizing/minimizing.


dkarrasch (approved):
Nice!

PR #1629, opened 2026-05-22
better error handling in stein

### LINKED ISSUE #1201: Possibly unreachable line in `lapack.jl`
Is the following line in `lapack.jl` even reachable? 

https://github.com/JuliaLang/LinearAlgebra.jl/blob/2a1696a93683e947b46991ea454c4ffb1988e794/src/lapack.jl#L4141

The call to `chklapackerror(info[])` on line 4138
https://github.com/JuliaLang/LinearAlgebra.jl/blob/2a1696a93683e947b46991ea454c4ffb1988e794/src/lapack.jl#L4138
should throw an error if any of the eigenvectors failed to converge and therefore its unlikely the if block would ever be reached. 

can someone suggest a test for this? I tried a bunch of matrices from the `MatrixDepot.jl` packages but none of them threw any errors when running `eigen`.

The relevant code for context (comments with `<--` mine)
```julia
# .... # 
# lapack.jl lines 4138 - 4142
            chklapackerror(info[]) 
            if any(ifail .!= 0) # <-- should be unreachable because of the line above
                # TODO: better error message / type
                error("failed to converge eigenvectors:\n$(findall(!iszero, ifail))") # <-- this line here
            end

# ... # 
```

araujoms:
Closes #1201

### DIFF
--- a/src/lapack.jl
+++ b/src/lapack.jl
@@ -4150,10 +4150,9 @@ for (stev, stebz, stegr, stein, elty) in
                 Ref{BlasInt}, Ptr{$elty}, Ptr{BlasInt}, Ptr{BlasInt},
                 Ptr{BlasInt}),
                 n, dv, ev, m, w, iblock, isplit, z, ldz, work, iwork, ifail, info)
-            chklapackerror(info[])
-            if any(ifail .!= 0)
-                # TODO: better error message / type
-                error("failed to converge eigenvectors:\n$(findall(!iszero, ifail))")
+            chkargsok(info[])
+            if info[] > 0
+                throw(ArgumentError(lazy"failed to converge eigenvectors: $(findall(!iszero, ifail))"))
             end
             z
         end
--- a/src/lapack.jl
+++ b/src/lapack.jl
@@ -4152,7 +4152,7 @@ for (stev, stebz, stegr, stein, elty) in
                 n, dv, ev, m, w, iblock, isplit, z, ldz, work, iwork, ifail, info)
             chkargsok(info[])
             if info[] > 0
-                throw(ArgumentError(lazy"failed to converge eigenvectors: $(findall(!iszero, ifail))"))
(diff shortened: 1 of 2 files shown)


src/lapack.jl
```diff
@@ -4150,10 +4150,9 @@ for (stev, stebz, stegr, stein, elty) in
                 Ref{BlasInt}, Ptr{$elty}, Ptr{BlasInt}, Ptr{BlasInt},
                 Ptr{BlasInt}),
                 n, dv, ev, m, w, iblock, isplit, z, ldz, work, iwork, ifail, info)
-            chklapackerror(info[])
-            if any(ifail .!= 0)
-                # TODO: better error message / type
-                error("failed to converge eigenvectors:\n$(findall(!iszero, ifail))")
+            chkargsok(info[])
+            if info[] > 0
+                throw(ArgumentError(lazy"failed to converge eigenvectors: $(findall(!iszero, ifail))"))
```
stevengj: Might want to limit output here in case this array has a few thousand entries.

```suggestion
                throw(ArgumentError(lazy"failed to converge eigenvectors: $(repr(findall(!iszero, ifail); context=:limit=>true))"))
```

stevengj: (If we're going to call `repr`, we might as well make the string non-`lazy`)

stevengj: Even better would be to have the `LazyString` method itself have a mechanism to enforce compact output when (if) the string is constructed, e.g. with `lazy"..."c` or something. 

https://github.com/JuliaLang/julia/pull/61887

araujoms: Nice, I'll wait for that to be merged and then update the PR.

stevengj: merged

araujoms: updated

stevengj (approved):

PR #1631, opened 2026-05-23
lbt: drop config caching

ViralBShah:
## Summary

`BLAS.lbt_get_config()` previously returned a cached `LBTConfig` that was invalidated only when callers went through the `BLAS.lbt_forward` Julia wrapper (via `_clear_config_with`). JLL `__init__` blocks that forward a library via raw `ccall(:lbt_forward, ...)` — done because JLLs cannot depend on LinearAlgebra — would silently leave the cached config stale. The newly-forwarded library would not appear in `lbt_get_config()` output even though the underlying libblastrampoline dispatch was correctly updated.

### Repro (before this PR)

On Julia 1.12 with \`OpenBLAS32_jll@0.3.33\` (which forwards into LBT's LP64 slot from its \`__init__\` via raw \`ccall\`):

\`\`\`julia
julia> using LinearAlgebra, OpenBLAS32_jll
julia> BLAS.lbt_get_config()
LBTConfig([ILP64] libopenblas64_.dylib)   # OpenBLAS32 missing despite being correctly forwarded
\`\`\`

\`lbt_get_forward(\"sgemm_\", LP64)\` returns the OpenBLAS32 \`sgemm_\` pointer correctly — only the cached config display was wrong.

### Fix

Fingerprint the live \`lbt_config_t\` on every \`lbt_get_config()\` call by snapshotting the NULL-terminated per-slot \`lbt_library_info_t*\` pointer array plus \`num_exported_symbols\`, and compare it against the cached fingerprint. The C-side \`lbt_get_config\` call itself is cheap, and the fingerprint comparison is O(num_libs) — typically 1–3 entries.

This correctly handles:
- **adds** (count grows → fingerprint differs → rebuild)
- **removes / \`clear=true\` replacements** (count may stay the same, but per-slot \`lbt_library_info_t*\` pointers differ → rebuild)
- **idempotent re-forwards** of an already-loaded library (fingerprint matches → cache preserved by identity)

\`_clear_config_with\` is retained for the eager-invalidation path used by \`lbt_set_default_func\`, which doesn't affect the fingerprint but may matter to downstream consumers.

## Test plan

- [x] Added \`@testset \"lbt_get_config cache invalidation\"\` in \`test/blas.jl\` covering the JLL-style raw-ccall path and the wrapper path.
- [x] Manually validated all four edge cases (stable, idempotent re-forward, raw add, raw \`clear=1\` replace) against the live LBT on \`aarch64-apple-darwin\` / Julia 1.12.6 + \`OpenBLAS32_jll@0.3.33\`.
- [x] CI on master / backport branches.

### DIFF
--- a/src/lbt.jl
+++ b/src/lbt.jl
@@ -192,32 +192,9 @@ function Base.show(io::IO, mime::MIME{Symbol("text/plain")}, lbt::LBTConfig)
     end
 end
 
-mutable struct ConfigCache
-    @atomic config::Union{Nothing,LBTConfig}
-    lock::ReentrantLock
-end
-
-# In the event that users want to call `lbt_get_config()` multiple times (e.g. for
-# runtime checks of which BLAS vendor is providing a symbol), let's cache the value
-# and clear it only when someone calls something that would cause it to change.
-const _CACHED_CONFIG = ConfigCache(nothing, ReentrantLock())
-
 function lbt_get_config()
-    config = @atomic :acquire _CACHED_CONFIG.config
-    config === nothing || return config
-    return lock(_CACHED_CONFIG.lock) do
-        local config = @atomic :monotonic _CACHED_CONFIG.config
-        config === nothing || return config
-        config_ptr = ccall((:lbt_get_config, libblastrampoline), Ptr{lbt_config_t}, ())
-        @atomic :release _CACHED_CONFIG.config = LBTConfig(unsafe_load(config_ptr))
-    end
-end
-
-function _clear_config_with(f)
-    lock(_CACHED_CONFIG.lock) do
-        @atomic :release _CACHED_CONFIG.config = nothing
-        f()
-    end
+    config_ptr = ccall((:lbt_get_config, libblastrampoline), Ptr{lbt_config_t}, ())
+    return LBTConfig(unsafe_load(config_ptr))
 end
 
 function lbt_get_num_threads()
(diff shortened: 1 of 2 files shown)


ViralBShah:
@staticfloat Would love a review here. `OpenBLAS32_jll` now automatically forwards LP64 symbols and this PR attempts to make sure that the lbt info shows forawards that were done directly from C code through a ccall.

staticfloat:
The whole point of the caching of `lbt_get_config()` was to avoid calling `lbt_get_config()` in the hot path; this PR calls it so that it can then make a comparison, so I'd say this doesn't make any sense.

That being said, the implementation of `lbt_get_config()` is extremely simple now so I think let's just not cache it at all and just call LBT every time.

staticfloat:
Oh, actually, reading through again, it's not the `lbt_get_config()` call that's slow, it's the `LBTConfig()` constructor that extracts a bunch of information from LBT.  That being said:

```
julia> @atomic :release LinearAlgebra.BLAS._CACHED_CONFIG.config = nothing

julia> @time LinearAlgebra.BLAS.lbt_get_config()
  0.000158 seconds (5.17 k allocations: 291.609 KiB)
LinearAlgebra.BLAS.LBTConfig
Libraries: 
├ [ILP64] libopenblas64_.dylib
└ [ LP64] libopenblas.0.3.33.dylib
```

I think we can just not do this caching, this is already plenty fast for an operation that isn't in hot loops.

ViralBShah:
Not caching makes a lot more sense.

ViralBShah:
Do you think it is safe to backport to 1.12 and 1.10?

KristofferC:
@staticfloat, it seems this PR changed a lot between your recent review and merge. Do you mind doing another post merge review?

staticfloat:
Yep, this looks good me, and this should be totally safe to backport.

PR #1632, opened 2026-05-24
Restrict strides methods to Transpose Number

### LINKED ISSUE #1618: Transpose with SMatrix elements falsely seems strided
Because `transpose` also takes the transpose of the elements, if the elements are changed by `transpose` the result should not be strided and `strides` should fail.
However, currently this is not the case.

MWE:

```julia-repl
julia> using LinearAlgebra, StaticArrays

julia> a = [SA[1;2;;3;4]]
1-element Vector{SMatrix{2, 2, Int64, 4}}:
 [1 3; 2 4]

julia> transpose(a)
1×1 transpose(::Vector{SMatrix{2, 2, Int64, 4}}) with eltype SMatrix{2, 2, Int64, 4}:
 [1 2; 3 4]

julia> pointer(transpose(a))
Ptr{SMatrix{2, 2, Int64, 4}}(0x000078472e337f20)

julia> strides(transpose(a))
(1, 1)

julia> unsafe_load(pointer(transpose(a))) == transpose(a)[1]
false
```

`Adjoint` only defines `strides` on `Adjoint{<:Real, <:AbstractVecOrMat}`.

The fix is to only define `Transpose` `strides` on `Transpose{<:Number, <:AbstractVecOrMat}` since `transpose` on `Number` is identity.

nhz2:
Fixes #1618


stevengj (approved):

nhz2:
bump

PR #1607, opened 2026-05-17
 extend structured broadcast to Symmetric and Hermitian 

araujoms:
Closes #643, closes #646, closes #772, closes #773.

On master we have
```julia
julia> S = Symmetric(randn(3,3)); S .*= 2
ERROR: ArgumentError: Cannot set a non-diagonal index in a symmetric matrix

julia> H = Hermitian(randn(ComplexF64, 3, 3)); H .*= 2
ERROR: ArgumentError: Cannot set a non-diagonal index in a Hermitian matrix

julia> abs.(S)
3×3 Matrix{Float64}:
 0.804083  0.100892  0.222156
 0.100892  0.341193  0.72708
 0.222156  0.72708   0.57603

julia> H .^ 2
3×3 Matrix{ComplexF64}:
  0.190678-0.0im        -0.296078-0.864739im  -0.593502+0.0136903im
 -0.296078+0.864739im    0.558651+0.0im        0.704788-0.588077im
 -0.593502-0.0136903im   0.704788+0.588077im   0.307546-0.0im
```
With this PR we get
```julia
julia> S = Symmetric(randn(3,3)); S .*= 2
3×3 Symmetric{Float64, Matrix{Float64}}:
 -4.29963    0.754017   1.08981
  0.754017  -1.30256   -4.91124
  1.08981   -4.91124    0.472419

julia> H = Hermitian(randn(ComplexF64, 3, 3)); H .*= 2
3×3 Hermitian{ComplexF64, Matrix{ComplexF64}}:
  -1.62433+0.0im       -1.92147+0.138269im  0.0249639-0.802402im
  -1.92147-0.138269im   1.95926+0.0im         2.64529-1.68015im
 0.0249639+0.802402im   2.64529+1.68015im    -1.58051+0.0im

julia> abs.(S)
3×3 Symmetric{Float64, Matrix{Float64}}:
 4.29963   0.754017  1.08981
 0.754017  1.30256   4.91124
 1.08981   4.91124   0.472419

julia> H .^ 2
3×3 Hermitian{ComplexF64, Matrix{ComplexF64}}:
   2.63846+0.0im        3.67293-0.531359im  -0.643226-0.0400622im
   3.67293+0.531359im   3.83871+0.0im         4.17465-8.88893im
 -0.643226+0.0400622im  4.17465+8.88893im     2.49801+0.0im

julia> H .*= im
ERROR: ArgumentError: cannot set a diagonal element of a hermitian matrix to a non-hermitian value
```

[...]

PR #1636, opened 2026-05-25
properly type broadcasting between Diagonal, SymTridiagonal, Symmetric, and Hermitian

araujoms:
Follow-up to #1607

Now `Diagonal` .+ `Hermitian`, `SymTridiagonal` .+ `Hermitian`, and `Symmetric` .+ `Hermitian` return properly typed matrices. With it we can remove the pseudo-broadcasting mechanism @jishnub had introduced for `Symmetric` and `Hermitian`.

I decided to sacrifice the combinations `Diagonal{<:Complex}` .+ `Hermitian{<:Real}`, `SymTridiagonal{<:Complex}` .+ `Hermitian{<:Real}`, and `Symmetric{<:Complex}` .+ `Hermitian{<:Real}`, as this drastically simplifies the code. They are of little use, as we have few methods that can handle `Symmetric{<:Complex}` (I found `bunchkaufman`, `kron`, and `dot`).

### DIFF
--- a/src/special.jl
+++ b/src/special.jl
@@ -265,25 +265,6 @@ function (-)(A::UniformScaling, B::Diagonal)
     Diagonal(Ref(A) .- B.diag)
 end
 
-for f in (:+, :-)
-    @eval function $f(D::Diagonal{<:Number}, S::Symmetric)
-        uplo = _sym_uplo(S.uplo)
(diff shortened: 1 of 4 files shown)


dkarrasch (approved):
Awesome!

PR #1638, opened 2026-05-26
more helpful error message for cholesky

### LINKED ISSUE #603: requiring type to signal Hermitian/Symmetric matrices
(follow-up to a [discussion on the forum](
https://discourse.julialang.org/t/cholesky-factorization-for-slightly-non-hermitian-matrices/20063/15))

This is a proposal to change the semantics of functions that require a symmetric/hermitian matrix as an argument.

Currently functions that require symmetric/hermitian matrices, like `LinearAlgebra.cholesky`, accept arbitrary `<: AbstractMatrix` arguments, then check for the desired symmetry. This leads to a lot of confusion among users since even if a calculation preserves symmetry in theory, in practice floating point error is very likely to break it. An example from that discussion is

```julia
julia> VERSION
v"1.2.0-DEV.219"

julia> using LinearAlgebra

julia> rot = [cos(π/4) -sin(π/4); sin(π/4) cos(π/4)]
2×2 Array{Float64,2}:
 0.707107  -0.707107
 0.707107   0.707107

julia> A = [25.0 0; 0 4.0]
2×2 Array{Float64,2}:
 25.0  0.0
  0.0  4.0

julia> cholesky(rot * A * rot')
ERROR: PosDefException: matrix is not Hermitian; Cholesky factorization failed.
Stacktrace:
 [1] checkpositivedefinite(::Int64) at /home/tamas/src/julia-git/usr/share/julia/stdlib/v1.2/LinearAlgebra/src/factorization.jl:11
 [2] #cholesky!#97(::Bool, ::Function, ::Array{Float64,2}, ::Val{false}) at /home/tamas/src/julia-git/usr/share/julia/stdlib/v1.2/LinearAlgebra/src/cholesky.jl:182
 [3] #cholesky#101 at ./none:0 [inlined]
 [4] cholesky at /home/tamas/src/julia-git/usr/share/julia/stdlib/v1.2/LinearAlgebra/src/cholesky.jl:275 [inlined] (repeats 2 times)
 [5] top-level scope at REPL[85]:1
```
but questions like this come up all the time.

The current workaround is to wrap a matrix in `Symmetric` or `Hermitian`. I am proposing that instead of working with general `<: AbstractMatrix` types, functions like `cholesky` only accept types which explicitly enforce symmetry. This would encourage a programming style that works uses wrapper types instead of (low-cost, but still significant) checks, and eliminate a surprises.

The change is breaking, even though a lot of existing code already wraps matrices so that would not be affected.

araujoms:
Closes #603

### DIFF
--- a/src/exceptions.jl
+++ b/src/exceptions.jl
@@ -45,11 +45,10 @@ end
 function Base.showerror(io::IO, ex::PosDefException)
     print(io, "PosDefException: matrix is not ")
     if ex.info == -1
-        print(io, "Hermitian")
+        print(io, "Hermitian; factorization failed. Try wrapping it with Symmetric or Hermitian.")
     else
-        print(io, "positive definite")
+        print(io, "positive definite; factorization failed.")
     end
-    print(io, "; Factorization failed.")
 end
 
 """


dkarrasch:
Is there another spot in the documentation where this could be added? Like with a nice `info` or even `warning` admonition box? This seems to be a common source of frustration.

araujoms:
That was the most general spot I could find, if you think of a better one I'm open to suggestions.

dkarrasch:
I think this is very helpful spot, indeed. I was thinking about some spot in the documentation, in case people look up method signatures and other info, maybe even the docstring? I thought we had some sort of "extended docstring", so we could hide things if otherwise the docstring becomes too lengthy.

araujoms:
The docstring of `cholesky` already mentions it, and the manual of `LinearAlgebra` as well.

I agree that this is a common source of frustration, but I don't know where else to add it.

dkarrasch (approved):

PR #1639, opened 2026-05-26
Rely on broadcasting for simple operations on `Diagonal`

### LINKED ISSUE #1448: Broadcast for diagonal should be lazy
Similar issue JuliaLang/julia#48443 
Many other functions that should be equivalent to their broadcasted version have the same problem (although some of them don't even lazy broadcast to ranges). Here is an example.

```julia
julia> Diagonal(1:5)+Diagonal(1:5) |> typeof
Diagonal{Int64, StepRangeLen{Int64, Int64, Int64, Int64}}

julia> Diagonal(1:5).+Diagonal(1:5) |> typeof
Diagonal{Int64, Vector{Int64}}

julia> (1:5) .+ (1:5) |> typeof
StepRangeLen{Int64, Int64, Int64, Int64}
```

dkarrasch:
Proposal to fix #1448.

### DIFF
--- a/src/diagonal.jl
+++ b/src/diagonal.jl
@@ -297,12 +297,7 @@ function tril!(D::Diagonal{T}, k::Integer=0) where T
 end
 
 (==)(Da::Diagonal, Db::Diagonal) = Da.diag == Db.diag
-(-)(A::Diagonal) = Diagonal(-A.diag)
-(+)(Da::Diagonal, Db::Diagonal) = Diagonal(Da.diag + Db.diag)
-(-)(Da::Diagonal, Db::Diagonal) = Diagonal(Da.diag - Db.diag)
 
-(*)(x::Number, D::Diagonal) = Diagonal(x * D.diag)
-(*)(D::Diagonal, x::Number) = Diagonal(D.diag * x)
 function lmul!(x::Number, D::Diagonal)
     if size(D,1) > 1
         # ensure that zeros are preserved on scaling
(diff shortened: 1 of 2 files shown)


dkarrasch:
There is an interesting failure with broadcasting over arrays of arrays (actually, `Diagonal`s of `Diagonal`s).

@jishnub @araujoms You have worked on specializing broadcasting (recently). Do you think this is fixable?

araujoms:
I think it is easy to fix, but I think the most sensible solution is to delete this test, trying to support matrices of matrices of matrices is the path to madness.

jishnub:
Leaving the tests aside, one of the possible consequences of this change is breaking the infinite array ecosystem that has arithmetic defined but may not specialize broadcasting. I think that situation has improved overall now that we have the `preprocess_broadcasted` mechanism to forward the broadcasting to the diagonals, so this change should be fine with the understanding that the breakages are because of package-specific implementations that are missing. To be specific, previously packages needed to specialize broadcasting for `Diagonal{T, InfRange{T}}`, whereas now they would only need to specialize broadcasting for `InfRange{T}`.

As for the tests, I agree that it makes sense to drop support for this case, as it's not a well-defined result anyway (the off-diagonals aren't defined). The reason it used to pass until now was that the summation ignored the off-diagonals, whereas it fails now because broadcasting explicitly checks that off-diagonals are preserved.

src/diagonal.jl
araujoms:
```suggestion
```
Should also be handled by broadcasting.

araujoms:
```suggestion
```
Should also be handled by broadcasting.

araujoms (approved):

PR #1641, opened 2026-05-27
Improve error message in `SymTridiagonal` constructor

jishnub:
After this, we print the types in the error message to make debugging easier.
```julia
julia> D = Diagonal(1:4)
4×4 Diagonal{Int64, UnitRange{Int64}}:
 1  ⋅  ⋅  ⋅
 ⋅  2  ⋅  ⋅
 ⋅  ⋅  3  ⋅
 ⋅  ⋅  ⋅  4

julia> SymTridiagonal(D)
ERROR: ArgumentError: the diagonal and the off-diagonal bands must have the same type.
The diagonal band has type UnitRange{Int64}, but the off-diagonal band has type Vector{Int64}.
[...]
```


araujoms:
But they don't need to have the same type:
```julia
julia> SymTridiagonal(1:2,[1.0])
2×2 SymTridiagonal{Float64, Vector{Float64}}:
 1.0  1.0
 1.0  2.0
```

jishnub:
In your example, the construction works because the eltypes are different, and we have
```julia
julia> convert(AbstractVector{Float64}, 1:2)
2-element Vector{Float64}:
 1.0
 2.0

julia> convert(AbstractVector{Int}, 1:2)
1:2
```
so in the first case it manages to convert the `UnitRange` to a `Vector`, but not in the latter. It's a weird case, but we don't have a promotion mechanism for `AbstractVector`s.

E.g. the following won't work:
```julia
julia> SymTridiagonal(1:2, [1])
ERROR: ArgumentError: diagonal vectors needed to be convertible to same type
```

araujoms:
Sure, but what I'm complaining about is that you changed the error message to claim they need to have the same type. This is not true, they just need to be convertible to the same time, as it currently says.

src/tridiag.jl
```diff
@@ -74,7 +74,9 @@ function SymTridiagonal{T}(dv::AbstractVector, ev::AbstractVector) where {T}
     e = convert(AbstractVector{T}, ev)::AbstractVector{T}
     typeof(d) == typeof(e) ?
         SymTridiagonal{T}(d, e) :
-        throw(ArgumentError("diagonal vectors needed to be convertible to same type"))
+        throw(ArgumentError(
+            lazy"""the diagonal and the off-diagonal bands must have the same type.
```
stevengj:
```suggestion
            lazy"""The diagonal and the off-diagonal bands must be convertible to a common type.
```

src/tridiag.jl
```diff
@@ -74,7 +74,9 @@ function SymTridiagonal{T}(dv::AbstractVector, ev::AbstractVector) where {T}
...
+            The diagonal band has type $(typeof(d)), but the off-diagonal band has type $(typeof(e))."""))
```
stevengj: Unfortunately, I don't think `lazy"..."` strings strip off leading whitespace.

src/tridiag.jl
```diff
@@ -74,7 +74,9 @@ function SymTridiagonal{T}(dv::AbstractVector, ev::AbstractVector) where {T}
     e = convert(AbstractVector{T}, ev)::AbstractVector{T}
     typeof(d) == typeof(e) ?
         SymTridiagonal{T}(d, e) :
```
araujoms:
```suggestion
    e = convert(typeof(d), ev)::AbstractVector{T}
    return SymTridiagonal{T}(d, e)
```
Frankly I'd rather just do a poor-man's type promotion here, and let `convert` throw the error.

The obvious disadvantage is that it's non-symmetric. It wouldn't introduce any new failure, though, and it would allow stuff like `SymTridiagonal([1, 2], 1:1)` to work.

src/tridiag.jl
```diff
@@ -71,10 +71,11 @@ SymTridiagonal(dv::V, ev::V) where {T,V<:AbstractVector{T}} = SymTridiagonal{T}(
...
 function SymTridiagonal{T}(dv::AbstractVector, ev::AbstractVector) where {T}
     d = convert(AbstractVector{T}, dv)::AbstractVector{T}
-    e = convert(AbstractVector{T}, ev)::AbstractVector{T}
-    typeof(d) == typeof(e) ?
-        SymTridiagonal{T}(d, e) :
-        throw(ArgumentError("diagonal vectors needed to be convertible to same type"))
+    e = convert(typeof(d), ev)::AbstractVector{T}
+    return SymTridiagonal{T}(d, e)
+        throw(ArgumentError(
+            lazy"""The diagonal and the off-diagonal bands must be convertible to a common type.
```
stevengj: I guess this isn't quote right either.  `d = 1:4` and `ev = [0,0,0]` are both convertible to a common type `typeof(ev)`, but they aren't convertible to `typeof(d)`, which is what the current algorithm needs.

src/tridiag.jl
```diff
@@ -71,10 +71,11 @@ SymTridiagonal(dv::V, ev::V) where {T,V<:AbstractVector{T}} = SymTridiagonal{T}(
 SymTridiagonal{T}(dv::V, ev::V) where {T,V<:AbstractVector{T}} = SymTridiagonal{T,V}(dv, ev)
 function SymTridiagonal{T}(dv::AbstractVector, ev::AbstractVector) where {T}
     d = convert(AbstractVector{T}, dv)::AbstractVector{T}
-    e = convert(AbstractVector{T}, ev)::AbstractVector{T}
-    typeof(d) == typeof(e) ?
-        SymTridiagonal{T}(d, e) :
-        throw(ArgumentError("diagonal vectors needed to be convertible to same type"))
+    e = convert(typeof(d), ev)::AbstractVector{T}
+    return SymTridiagonal{T}(d, e)
+        throw(ArgumentError(
```
stevengj: This `throw` is never reached?

araujoms: The `throw` part should have been deleted, but github's suggestion feature didn't work for it.

src/tridiag.jl
```diff
@@ -71,10 +71,11 @@ SymTridiagonal(dv::V, ev::V) where {T,V<:AbstractVector{T}} = SymTridiagonal{T}(
 SymTridiagonal{T}(dv::V, ev::V) where {T,V<:AbstractVector{T}} = SymTridiagonal{T,V}(dv, ev)
 function SymTridiagonal{T}(dv::AbstractVector, ev::AbstractVector) where {T}
     d = convert(AbstractVector{T}, dv)::AbstractVector{T}
-    e = convert(AbstractVector{T}, ev)::AbstractVector{T}
-    typeof(d) == typeof(e) ?
-        SymTridiagonal{T}(d, e) :
-        throw(ArgumentError("diagonal vectors needed to be convertible to same type"))
+    e = convert(typeof(d), ev)::AbstractVector{T}
```
stevengj, line 74: This `convert` call could throw:
```jl
julia> d = 1:4; ev = [0,0,0]; convert(typeof(d), ev)
ERROR: MethodError: Cannot `convert` an object of type Vector{Int64} to an object of type UnitRange{Int64}
```

araujoms: That's the goal.

araujoms (approved):

PR #1643, opened 2026-05-27
Remove methods for some unary operations

dkarrasch:
From what I understand, these should all work with the generic fallbacks and (except for `copy`) via broadcasting. Not surprisingly, this is in many cases even faster. For instance, broadcasting doesn't even touch the unseen triangle with `*Triangular` matrices.

### DIFF
--- a/src/bidiag.jl
+++ b/src/bidiag.jl
@@ -310,9 +310,9 @@ size(M::Bidiagonal) = (n = length(M.dv); (n, n))
 axes(M::Bidiagonal) = (ax = axes(M.dv, 1); (ax, ax))
 
 #Elementary operations
-for func in (:conj, :copy, :real, :imag)
-    @eval ($func)(M::Bidiagonal) = Bidiagonal(($func)(M.dv), ($func)(M.ev), M.uplo)
-end
+# for func in (:conj, :copy, :real, :imag)
+#     @eval ($func)(M::Bidiagonal) = Bidiagonal(($func)(M.dv), ($func)(M.ev), M.uplo)
+# end
 isreal(M::Bidiagonal) = isreal(M.dv) && isreal(M.ev)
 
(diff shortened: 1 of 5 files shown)


dkarrasch:
I wonder if in some cases it is better to forward, say, `-` to the parent, which may then be implemented differently than via broadcasting. Things like GPU arrays or `StaticArrays.jl` come to mind.

dkarrasch:
Interesting. This currently fails because of

```julia
Expression: copy(Diagonal(1:5)) === Diagonal(1:5)
Evaluated: Diagonal([1, 2, 3, 4, 5]) === Diagonal(1:5)
```

Generically, `copy(A) = copyto!(similar(A), A)`, and `similar` returns something writable, that's why the vector. For `Bidiagonal`, we have a similar (seemingly untested) behavior:

```julia
julia> B = copy(Bidiagonal(1:5, 1:4, :U))
5×5 Bidiagonal{Int64, UnitRange{Int64}}
```

araujoms:
That's a great cleanup! I wonder how the current state came about. I supposed all these methods are older than structured broadcast, and nobody remembered to remove them after it was introduced.

I don't think `copy` is fixable, because `similar` is documented to return something mutable, and neither `Diagonal` nor `UnitRange` are. I'd just leave it as it is, the current behaviour is pretty useful.

`StaticArrays` is a problem. It relies on these methods, as it doesn't have specific structured broadcast rules. In fact the ones I already removed cause regressions: `Hermitian{SMatrix} + Hermitian{SMatrix}`, `real(::Hermitian{SMatrix})`, etc. Either I add them back, or add `StaticArrays` needs to add broadcast rules.

stevengj:
In fact, these methods pre-date dot-syntax broadcasting, dating back to https://github.com/JuliaLang/julia/pull/5277

araujoms:
It also breaks `SparseArrays`, and probably half the ecosystems. We need to keep these methods. It would be very useful to be able to run Nanosoldier from this repository.

What could be done is to let structured broadcasting handle `StridedMatrix`, and keep these methods as a generic fallback.

PR #1645, opened 2026-05-27
make eig*(::Bidiagonal) accept sortby

### LINKED ISSUE #1644: eig*(::Bidiagonal) doesn't accept sortby
@araujoms recently added `sortby` support to multiple specialized eigensolvers (see #1608), but `Bidiagonal` seems to have been overlooked: https://github.com/JuliaLang/LinearAlgebra.jl/blob/1cea6da8003716de37b96a60492fb4a178b2ba7c/src/bidiag.jl#L1442-L1475

araujoms:
Closes #1644

In my defence I had tested it; I just didn't realise that `eigvals(B; sortby)` was falling back to the general case instead of using the dedicated `Bidiagonal` methods.

### DIFF
--- a/src/bidiag.jl
+++ b/src/bidiag.jl
@@ -1452,10 +1452,7 @@ function eigvecs(M::Bidiagonal{T}; sortby=eigsortby) where T
             for j = blks[idx_block] + 1:i - 1 #Starting from j=i, eigenvector elements will be 0
                 v[j+1] = (M.dv[i] - M.dv[j])/M.ev[j] * v[j]
             end
-            c = norm(v)
-            for j = 1:n
-                Q[j, i] = v[j] / c
-            end
+            Q[:, i] = normalize!(v)
         end
     else
         for idx_block = 1:length(blks) - 1, i = blks[idx_block + 1]:-1:blks[idx_block] + 1 #index of eigenvector
@@ -1464,10 +1461,7 @@ function eigvecs(M::Bidiagonal{T}; sortby=eigsortby) where T
             for j = (blks[idx_block+1] - 1):-1:max(1, (i - 1)) #Starting from j=i, eigenvector elements will be 0
                 v[j] = (M.dv[i] - M.dv[j+1])/M.ev[j] * v[j+1]
             end
-            c = norm(v)
-            for j = 1:n
-                Q[j, i] = v[j] / c
-            end
+            Q[:, i] = normalize!(v)
         end
     end
     return isnothing(sortby) ? Q : sorteig!(copy(M.dv), Q, sortby)[2]
--- a/src/bidiag.jl
+++ b/src/bidiag.jl
@@ -1470,7 +1470,7 @@ function eigvecs(M::Bidiagonal{T}; sortby=eigsortby) where T
             end
         end
     end
-    return sorteig!(copy(M.dv), Q, sortby)[2]
+    return isnothing(sortby) ? Q : sorteig!(copy(M.dv), Q, sortby)[2]
 end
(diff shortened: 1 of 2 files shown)


src/bidiag.jl
stevengj: I noticed that this is vulnerable to overflow and subnormal numbers.  Probably better to just do
```jl
Q[:, i] = normalize!(v)
```
even though it requires an extra pass over `v`.

stevengj: (This is a separate issue from sorting, but might as well fix it in this PR?)

araujoms: That makes the tests fail, because they test `Bidiagonal(::BigFloat)`, and it turns out that `issubnormal(::BigFloat)` is not defined.

I think it is rather sensible to define it as unconditionally false, but this definition needs to be in `Base`. I'll make a PR.

araujoms: https://github.com/JuliaLang/julia/pull/61929

stevengj (approved):

src/bidiag.jl
stevengj: Fair enough, improving the normalization can happen in a later PR.  Or you can update this PR once the `issubnormal(::BigFloat)` PR is merged.

src/bidiag.jl
```diff
@@ -1470,9 +1470,9 @@ function eigvecs(M::Bidiagonal{T}) where T
             end
         end
     end
-    Q #Actually Triangular
+    return sorteig!(copy(M.dv), Q, sortby)[2]
```
stevengj: avoid `copy(M.dv)` if `sortby === nothing`:
```suggestion
    return isnothing(sortby) ? Q : sorteig!(copy(M.dv), Q, sortby)[2]
```

test/bidiag.jl
```diff
@@ -1296,4 +1296,16 @@ end
...
 end
 
+@testset "eigenvalue sorting" begin
+    for T in (Float64, ComplexF64, Float16, ComplexF16)
+        B = Bidiagonal(randn(T, 4), randn(T, 3), :U)
+        @test eigvals(B; sortby=nothing) == diag(B)
+        F = eigen(B)
+        @test issorted(F.values, by=LinearAlgebra.eigsortby) #sort by default
+        @test B * F.vectors ≈ F.vectors * Diagonal(F.values)
+        @test F.values ≈ eigvals(B; sortby = LinearAlgebra.eigsortby)
```
stevengj:
```suggestion
        @test F.values == sort!(diag(B), by=LinearAlgebra.eigsortby) == eigvals(B; sortby = LinearAlgebra.eigsortby)
```

araujoms: The goal here is to test whether `eigvals` sorts correctly, so I need to call it. `F.values` was already tested to be sorted.

src/bidiag.jl
araujoms: I'll update the PR once its merged.

test/bidiag.jl
stevengj: You ideally want to test that it not only sorts correctly but that it returns the correct values.

e.g. code that returned `fill(B[1,1], n)`(and the corresponding eigenvector repeated `n` times) if `sortby === LinearAlgebra.eigsortby` would also pass your test.

stevengj: Anyway, I updated the suggestion so that it tests both.

Note also that you can use `==` since the sorting should not change the values.

araujoms: Fair enough. My idea was to test only sorting here, since `eigen` was already tested before, but you're right, it had no test with `sortby=eigsortby`.

PR #1649, opened 2026-05-31
ci: remove unused TagBot workflow

arpitjain099:
Small CI hardening: declares an explicit workflow-level `permissions: contents: read` on 1 workflow(s) that currently inherit the default broad read-write GITHUB_TOKEN.

I inspected each file before including it; none publish, push, comment on issues/PRs, or otherwise write via the GitHub API, so the read-only default does not change behavior. Workflows that need to write (stale, release, gh-pages-deploy, publish actions, etc.) are intentionally left out of this PR.

This is the post-CVE-2025-30066 hardening pattern for default token scope.


DilumAluthge:
Thank you for this PR! For this specific case (TagBot), unfortunately I think (IIRC) that TagBot will need at least some write permissions, in order to be able to create tags and releases. Off the top of my head I don't actually remember what specific permissions TagBot requires.

I do agree with you that in almost every case, `contents: read` is sufficient, but unfortunately I don't think TagBot is one of those cases.

However, we should still try to narrow the permissions for TagBot; TagBot doesn't need `write-all`.

dkarrasch:
I'm surprised that we even have a TagBot workflow, given that we don't have releases for `LinearAlgebra.jl` itself, but do releases via Julia.

arpitjain099:
You're right, contents: read was too tight here. TagBot needs contents: write to push the tag and create the release, so plain read would break it.

I have switched to the scoped permission block from the JuliaRegistries/TagBot docs:

```yaml
permissions:
  actions: read
  checks: read
  contents: write
  deployments: read
  issues: read
  packages: read
  pull-requests: read
  statuses: read
```

That keeps the one write scope TagBot actually needs and makes everything else read-only, so it is still a real narrowing from the default write-all token.

On @dkarrasch's point: if TagBot is genuinely unused because releases go through Julia itself, removing the workflow outright would be cleaner than scoping it. Happy to switch this PR to a removal instead if that is the direction you prefer.

DilumAluthge:
Actually yeah, I think @dkarrasch is right, we don't need TagBot here - @arpitjain099 can you switch this PR to remove the TagBot workflow entirely?

arpitjain099:
Done. Switched the PR to remove `.github/workflows/tagbot.yml` entirely, since releases go through Julia rather than a standalone tag here. Updated the title to match.

DilumAluthge (approved):

DilumAluthge:
Thank you!

PR #1653, opened 2026-06-09
Add `abs` matrix function

### LINKED ISSUE #1648: Add `abs` matrix function
I think this is the last matrix function still missing from `LinearAlgebra`, in the sense that it is both common enough to belong in the standard library and nontrivial enough to deserve a dedicated implementation.

In principle one can define it equally well as `abs(A) = sqrt(A'A)` or `abs(A) = sqrt(A*A')`, but the convention is that the former is used; this fits with the convention of writing the polar decomposition as `A ≈ U * abs(A)` instead of `A ≈  abs(A) * U`.

The definition is however not a good way to compute it. In the Hermitian case one can simply compute the eigendecomposition and take the absolute value of the eigenvalues; in the general (even non-square) case, probably the best way to compute is via the singular value decomposition: let `u, s, v = svd(A)`, then `abs(A) = v' * Diagonal(s) * v`. See discussion at https://discourse.julialang.org/t/absolute-value-of-a-matrix/98301

akshhkaushik:
Closes #1648.

Adds `abs(A::AbstractMatrix)` — the matrix absolute value, i.e the Hermitian positive-semidefinite factor of the polar decomposition, `P = sqrt(A'A)`.

As suggested in the issue I went through the SVD rather than forming `A'A` directly (which squares the condition number): for `A = U*Diagonal(s)*V'`, `abs(A) = V*Diagonal(s)*V'`. Symmetric/Hermitian inputs take the cheaper `eigen` path with abs'd eigenvalues, and `Diagonal` just maps `abs` over the diagonal.

Couple of things worth pointing out:
- the result is always Hermitian PSD, so i symmetrise the SVD product (real diagonal + mirrored triangle) so it comes out *exactly* hermitian instead of only up-to-roundoff.
- `abs(A')` is `sqrt(A*A')`, **not** `adjoint(abs(A))`, so unlike `sqrt` there's deliberately no Adjoint/Transpose method — those fall through to the SVD path.
- it's the *matrix* abs, not the elementwise `abs.(A)`. The docstring has a note since that's an easy mixup.

Tests (general, non-square both ways, complex, Herm/Sym, diagonal, empty) pass locally:

```
Test Summary:                              | Pass  Total   Time
abs(A::AbstractMatrix)                     |   54     54  20.5s
  general Float64                          |   11     11  14.2s
  general ComplexF64                       |   11     11   2.8s
  Hermitian/symmetric fast path Float64    |   17     17   1.4s
  Hermitian/symmetric fast path ComplexF64 |    9      9   0.8s
  diagonal                                 |    4      4   0.6s
  empty                                    |    2      2   0.6s
```

Also updated the docs list and `HISTORY.md`.


araujoms:
I will neither review nor merge AI-written PRs. This is not Julia's policy, this is my policy, so maybe somebody else will be willing. I'd just remark that this is an easy issue, which I created with the goal of attracting contributors - not contributions. Letting Claude do it defeats the purpose. If I just wanted to get it done I would have done it myself.

aksh08022006:
> I will neither review nor merge AI-written PRs. This is not Julia's policy, this is my policy, so maybe somebody else will be willing. I'd just remark that this is an easy issue, which I created with the goal of attracting contributors - not contributions. Letting Claude do it defeats the purpose. If I just wanted to get it done I would have done it myself.

Sorry, I didn't mean to undermine the purpose of the issue or your effort in creating it.

I genuinely enjoy solving algorithmic and mathematical problems myself, but I do use Claude as a productivity tool for navigating and managing larger codebases. Looking back, I can see why this PR wasn't aligned with what you were hoping contributors would get out of the issue.

I'll go ahead and close this PR , would push another completely by myself

PR #1654, opened 2026-06-09
Implement `ordschur` with arbitrary reordering of eigenvalues

aravindh-krishnamoorthy:
Potential fix for https://github.com/JuliaLang/LinearAlgebra.jl/issues/1047

```julia
function ordschur!(S::Schur, p::AbstractVector{<:Integer})
```

Code implements the algorithm for arbitrary reordering of eigenvalues.

For LAPACK-supported types, `trexc!` is used, which swaps the blocks using the numerically stable algorithm from [BD93]. For other types, a generic pure Julia implementation follows the classical unblocked method, without windowing, see introductory discussion in [BD93]. It solves the Sylvester equation using LU factorization. I've used this approach in https://github.com/JuliaLang/LinearAlgebra.jl/blob/46a75502b271c54d44a926e9461d9defbd3a331e/src/triangular.jl#L3073 about $2.5$ years ago and it seems Ok so far.

[BD93] Zhaojun Bai and James W. Demmel, "On swapping diagonal blocks in real Schur form," Linear Algebra and its Applications Volume 186 pp 75-95, 1993, https://doi.org/10.1016/0024-3795(93)90286-W

- [X] First implementation
- [ ] $2\times 2$ block index clarification (see code TODO)
- [ ] Optimised implementation
- [X] Documentation
- [ ] Documentation compilation check
- [X] Unit test
- [ ] Unit test check


aravindh-krishnamoorthy:
# Optimisation Tracker
_(This comment will be continually updated with the current optimisation status)_

$1024\times1024$ matrix single reordering: about $3\times$ LAPACK for https://github.com/JuliaLang/LinearAlgebra.jl/pull/1654/changes/9f7bfdbc84eecc30358625049e028113d6460c57 (Pure Julia and LU based Sylvester solve)

```julia
julia> using LinearAlgebra
julia> using BenchmarkTools
julia> N = 1024;
julia> X = randn(N,N);
julia> S = schur(X);
julia> idx = findlast(x->isreal(x), S.values);
julia> p = circshift(1:N,N-idx+1);
julia> b = falses(N);
julia> b[idx:end] .= true;
julia> S1 = ordschur(S,b);
julia> S2 = ordschur!(copy(S),p);
julia> norm(S1.values - S2.values)
1.5089416132685045e-13
julia> norm(S1.values - S.values)
1096.6594764777658
julia> norm(S1.Z*S1.T*S1.Z' - X)
1.892660374773724e-11
julia> norm(S2.Z*S2.T*S2.Z' - X)
1.8935267963444116e-11
julia> @btime ordschur(S,b);
  5.880 ms (27 allocations: 16.05 MiB)
julia> @btime ordschur!(copy(S),p);
  18.555 ms (6808 allocations: 46.56 MiB)
```

$1024\times1024$ matrix single reordering: about the same as LAPACK for https://github.com/JuliaLang/LinearAlgebra.jl/pull/1654/changes/98f408f2b17306ef59862f43e92e278d90dd975d (LAPACK's `trexc!` based)

```julia
julia> using LinearAlgebra
julia> using BenchmarkTools
julia> N = 1024;
julia> X = randn(N,N);
julia> S = schur(X);
julia> idx = findlast(x->isreal(x), S.values);
julia> p = circshift(1:N,N-idx+1);
julia> b = falses(N);
julia> b[idx:end] .= true;
julia> S1 = ordschur(S,b);
julia> S2 = ordschur!(copy(S),p);
julia> norm(S1.values - S2.values)
1.4568041557268838e-13
julia> norm(S1.values - S.values)
1102.482335591407
julia> norm(S1.Z*S1.T*S1.Z' - X)
1.979303345030665e-11
julia> norm(S2.Z*S2.T*S2.Z' - X)
1.979303345030665e-11
julia> @btime ordschur(S,b);
  5.897 ms (27 allocations: 16.05 MiB)
julia> @btime ordschur!(copy(S),p);
  6.614 ms (1098 allocations: 20.23 MiB)
```

src/schur.jl
```diff
@@ -426,6 +430,7 @@ a complex conjugate pair of eigenvalues must be either both included in `p` succ
     return S
 end
 
+# Generic Julia routine
 @views @inline function _swap_adj_schur_blocks!(T::AbstractMatrix, Z::AbstractMatrix,
     i1::Int, i2::Int, j1::Int, j2::Int, s1::Int, s2::Int, n::Int,
     Δ::AbstractMatrix, M_K0::AbstractMatrix, M_K1::AbstractMatrix, M_rhs::AbstractVector,
```
dkarrasch: If you only need dispatch w.r.t. the first two arguments, maybe you can skip the type annotations of all the other arguments?

src/schur.jl
```diff
@@ -466,6 +471,14 @@ end
     end
 end
 
+# LAPACK routine for BLAS types
+@views @inline function _swap_adj_schur_blocks!(T::StridedMatrix{<:LinearAlgebra.BlasFloat}, Z::StridedMatrix{<:LinearAlgebra.BlasFloat},
+    i1::Int, i2::Int, j1::Int, j2::Int, s1::Int, s2::Int, n::Int,
+    Δ::AbstractMatrix, M_K0::AbstractMatrix, M_K1::AbstractMatrix, M_rhs::AbstractVector,
+    M_X::AbstractMatrix, M_Q::AbstractMatrix)
```
dkarrasch: Unused arguments could be left out, and if you remove the type annotations for some of the arguments, you can replace them by `_`.

src/schur.jl
aravindh-krishnamoorthy: Thank you! Done.

src/schur.jl
aravindh-krishnamoorthy: Thank you! Done.

aravindh-krishnamoorthy:
@dkarrasch Is there a clean way to test both implementations with `BlasFloat` inputs? The LAPACK path is selected automatically, but I would also like to force the generic helper without adding a Boolean switch. Ideally, the test should still call `ordschur!` rather than the helper directly... It's not just to test the generic functionality but also to compare their performance.

dkarrasch:
I don't know any other way but to comment out the LAPACK method locally. As for testing in CI, you could promote some test matrix to `big`?

araujoms:
I don't think this is possible via calling `ordschur!`. Calling `_swap_adj_schur_blocks!` directly can be done though with `@invoke`.

aravindh-krishnamoorthy:
Thank you @dkarrasch and @araujoms. For CI, since the matrix sizes will be small, I'll use `BigFloat` to test the generic path. For internal tests and benchmarking, I'll just comment out the LAPACK function.

aravindh-krishnamoorthy:
Update: Sorry that this has been idle for some time now. I'll work on this this weekend and open the PR up for reviews. The only major task remaining is removal of a few duplicate variables, leading to a small improvement.

PR #1657, opened 2026-06-18
Support `I * X = X` for Arrays

### LINKED ISSUE #1656: Support `I * X = X` for Arrays, i.e., UniformScaling as a true identity operator
The docstring for `UniformScaling` describes it as an "identity operator"
yet the support multiplication works _only_ for a `Vector` or `Matrix`,
not for a general `Array`, as illustrated by this MWE:

```julia
using LinearAlgebra: I
5I * ones(2,3) # works of course
5I * ones(2,3,4) # fails with MethodError
```

The culprit is this one line of source code:
```julia
*(J::UniformScaling, A::AbstractVecOrMat) = J.λ*A
```
https://github.com/JuliaLang/LinearAlgebra.jl/blob/27b0f1321777fd2f935b8b9f1cefc61e12114b29/src/uniformscaling.jl#L279
that limits `A` to be a matrix or vector.

My feature request is to generalize this to
```julia
*(J::UniformScaling, A::AbstractArray) = J.λ*A
```

With this tiny change, `I` really become an "identity operator" rather than merely an identity "matrix."

For context, I've put this more general method in one of my packages for years, but Aqua.jl tells me that it is type piracy, so it would be better to have this generality here in `LinearAlgebra`.
I will make a PR.

JeffFessler:
Closes #1656


stevengj:
Let's keep discussion of whether this is desirable at #1656

JeffFessler:
Oh drat, I tried to rename the branch to something more informative than `patch-1` and then github closed this PR.  Sorry for the extra noise here.  So I leave it open with the original branch name.

JeffFessler:
@araujoms thanks for the branch merge.  Would you be willing to weigh in at issue #1656 on the merits?

PR #1658, opened 2026-06-18
Check that matrix is square in `det` triangular fast-path

jmert:
The fast-path for `det(A)` where `A` is lower triangular (in the `istril(A)` sense)

$$
A = \begin{bmatrix} 1 & 0 \\\\ 0 & 1 \\\\ 1 & 1 \end{bmatrix}
$$

(or similar for the transpose, which is upper triangular) skips checking that `A` is square and incorrectly computes the determinant of just the upper (or left) block of the input.

Add `checksquare(A)` to the fast path branch, and add unit tests.

---

_Alternatively, I see that `logabsdet(A)` wraps the triangular inputs in `UpperTriangular` which then throws if non-square, so I could copy that instead. I'm just not sure whether there was a reason for inlining the product of diagonals in the `det` case._


dkarrasch:
I'd be in favor of wrapping the `AbstractMatrix` in a `UpperTriangular`, and move the type promotion mechanism there. Seems inconsistent to have that in the abstract case, and to simply multiply diagonal entries in the `*Triangular` cases. I assume that this is needed in the abstract case to guarantee the same output eltype as for the `lu` branch.

src/triangular.jl
```diff
@@ -3008,8 +3008,14 @@ logdet(A::UnitUpperTriangular{T}) where {T} = zero(T)
...
 logabsdet(A::UnitLowerTriangular{T}) where {T} = zero(T), one(T)
-det(A::UpperTriangular) = prod(diag(A.data))
-det(A::LowerTriangular) = prod(diag(A.data))
+function det(A::Union{UpperTriangular{T},LowerTriangular{T}}) where T
+    S = promote_type(T, typeof((one(T) * zero(T) + zero(T)) / one(T)))
+    v = one(S)
+    @inbounds for i in axes(A.data, 1)
+        v *= convert(S, A.data[i,i])
+    end
+    return v
```
```suggestion
    return prod(Base.Fix1(convert, S), diagview(A.data); init=one(S))
```

stevengj:
> I'd be in favor of wrapping the AbstractMatrix in a UpperTriangular, and move the type promotion mechanism there. Seems inconsistent to have that in the abstract case, and to simply multiply diagonal entries in the *Triangular cases. I assume that this is needed in the abstract case to guarantee the same output eltype as for the lu branch.

`AbstractMatrix{BigInt}` has a specialized method that gives an exact answer, so `UpperTriangular{BigInt}` should also be exact.

src/triangular.jl
```diff
@@ -3008,8 +3008,10 @@ logdet(A::UnitUpperTriangular{T}) where {T} = zero(T)
 logdet(A::UnitLowerTriangular{T}) where {T} = zero(T)
 logabsdet(A::UnitUpperTriangular{T}) where {T} = zero(T), one(T)
 logabsdet(A::UnitLowerTriangular{T}) where {T} = zero(T), one(T)
-det(A::UpperTriangular) = prod(diag(A.data))
-det(A::LowerTriangular) = prod(diag(A.data))
+function det(A::Union{UpperTriangular{T},LowerTriangular{T}}) where T
+    S = promote_type(T, typeof((one(T) * zero(T) + zero(T)) / one(T)))
+    return prod(Base.Fix1(convert, S), diagview(A.data); init=one(S))
+end
```
stevengj, line 3014: `BigInt` determinants should be exact:
```suggestion
end
det(A::Union{UpperTriangular{BigInt},LowerTriangular{BigInt}}) = prod(diagview(A.data))
```

stevengj:
Note that `det(::Diagonal)` also just multiplies the diagonals without promoting.

dkarrasch:
> Note that `det(::Diagonal)` also just multiplies the diagonals without promoting.

Should we discuss consistency in another issue? I'm going to merge this as the "conservative approach", i.e., without changes in behavior.

PR #1659, opened 2026-06-24
Fix integer overflow in generic matrix-vector multiplication

maleadt:
The inner loop formed each product in the narrow input eltype before widening, so narrow-integer inputs overflowed. Form products in the accumulator type via muladd, as already done for matrix-matrix: https://github.com/JuliaLang/LinearAlgebra.jl/blob/27b0f1321777fd2f935b8b9f1cefc61e12114b29/src/matmul.jl#L1216-L1220


stevengj (approved):

simeonschaub:
Looks like this broke the array ops tests in Julia base: https://buildkite.com/julialang/julia-pr/builds/1646#01a01983-34fd-4a29-af13-35567a603910 (ref https://github.com/JuliaLang/julia/pull/62823)

dkarrasch:
Interesting. The (seemingly non-existent) corresponding matmat test case would have failed as well. The underlying "problem" is that `muladd` by default promotes its arguments (all three of them!), which is obviously nonsense for unitful arguments. To change that promotion behavior would be very breaking, so perhaps that artificial unitful number type needs to overload `muladd` without using promotion?

PR #1663, opened 2026-07-05
Make `norm_sqr` efficient, auto-differentiable, and exported

### LINKED ISSUE #1662: A 2-norm squared function would be helpful, e.g., fix and export `norm_sqr`
Least-squares minimization is ubiquitous and a typical implementation involves minimizing a cost function of the form `cost(x) = norm(model(x) - data)^2`

However, `norm(x)^2` is not auto-differentiable at zero because of the `sqrt` in the implementation of `norm`.
See: https://juliadiff.org/ForwardDiff.jl/stable/user/limitations/

It would be convenient for LinearAlgebra to provide the square of the 2-norm
as a built-in exported function to encourage use of the auto-differentiable version, e.g.:
`norm_sqr(x) = sum(abs2, x)`
Importantly, this version is also _more efficient_ because it avoids the redundant `sqrt` and squaring operations.

The predictable pushback here is that it is a simple one-line function that anyone can implement themselves.  My counter is that Julia base provides `abs2(x::Real) = x*x` (and something related for complex values), presumably both for efficiency and for differentiability compared to the "obvious" way `abs(x)^2`.

A main question would be the name.  A natural parallel to `abs2` would be `norm2` but that name is already used internally in the package and could be confused with the 2-norm itself.

The package already has `norm_sqr` hidden inside it:

https://github.com/JuliaLang/LinearAlgebra.jl/blob/46a75502b271c54d44a926e9461d9defbd3a331e/src/generic.jl#L571

but it is not exported, there is no docstring, and sadly it uses the inefficient and non-differentiable version!

Proposal:
- fix the definition to be `norm_sqr(x) = sum(abs2, x)` or `norm_sqr(x) = sum(abs2 ∘ float , x)`
- add docstring
- export it.

I'd make a PR if there is openness to this.

I am unsure whether the `∘ float` part should be included.  I saw that version in https://github.com/JuliaLang/LinearAlgebra.jl/issues/1650

JeffFessler:
Fixes #1662, following the approach suggested here:
https://github.com/JuliaLang/LinearAlgebra.jl/issues/1662#issuecomment-4883588884


src/generic.jl
```diff
@@ -567,10 +567,22 @@ generic_normInf(x) = float(mapreduce(norm, max, x))
...
+fall back to `norm(x)^2`,
+which is less efficient
+and is not auto-differentiable at `zero`
+due to the `sqrt` in `norm`.
+"""
+norm_sqr(x::AbstractArray) = sum(norm_sqr, x)
 # faster computation of norm(x)^2, avoiding overflow for integers
-norm_sqr(x) = norm(x)^2
-norm_sqr(x::Number) = abs2(x)
 norm_sqr(x::Union{T,Complex{T},Rational{T}}) where {T<:Integer} = abs2(float(x))
```
stevengj: This method seems unnecessary now?

src/generic.jl
```diff
@@ -567,10 +567,22 @@ generic_normInf(x) = float(mapreduce(norm, max, x))
...
+"""
+    norm_sqr(x)
+Compute the squared Euclidean norm of `x` efficiently,
+i.e., typically as `sum(abs2, x)`.
+
+For types other than `AbstractArray` or `Number`,
+fall back to `norm(x)^2`,
+which is less efficient
+and is not auto-differentiable at `zero`
+due to the `sqrt` in `norm`.
```
stevengj:
```suggestion
Compute the squared 2-norm, equivalent to `norm(x)^2` but
potentially more efficiently (and perhaps friendlier to automatic differentiation).
```

araujoms:
If the function is going to be exported I think it would be better to name it `norm_squared`, arbitrary abbreviations are not user friendly.

src/generic.jl
JeffFessler: Agreed - I removed it.

Some form of comment seems needed for why `float` is in the next line,
so I took a stab at that, but feel free to modify.

JeffFessler:
Making it a "draft" until the name choice is resolved.

dkarrasch:
If this is exported, I believe the corresponding docstring needs to be included in the docs.

stevengj:
As I commented in the linked issue, I'm not 100% convinced of the need for this function.  It's not strictly needed for AD gradients, and the performance advantage over `norm(x)^2` is likely to be negligible in most cases once the performance of `norm` is improved.  On the other hand, it is more needed for AD Hessians.

But I think that discussion should happen in #1662.

Issue #1664, opened 2026-07-15
Docstrings tests do not run in CI (and currently fail)


topolarity:
> I don't know why it's not failing on github, but in any case public names should be documented.

Looks like the Docstrings tests are being accidentally skipped:

> The repo's Buildkite runs .ci/run_tests.jl, which calls Base.runtests(["LinearAlgebra"]) — julia's harness. Julia's harness (test/choosetests.jl:246-249) sees LinearAlgebra's testgroups file and expands the test set into the individual group files (matmul.jl, triangular.jl, …), executing them directly — it never executes runtests.jl itself, and the Docstrings (and versioninfo) testsets live only there, after the include loop.

(analysis from Claude Fable)

looks like `Base.runtests(["LinearAlgebra"])` skips these but `Pkg.test("LinearAlgebra")` hits them

_Originally posted by @topolarity in https://github.com/JuliaLang/LinearAlgebra.jl/issues/1584#issuecomment-4976482019_

PR #1666, opened 2026-07-23
Move CI from buildkite to GithubActions

dkarrasch:


.github/workflows/ci.yml
```diff
@@ -12,6 +12,41 @@ concurrency:
...
+      fail-fast: false
+      matrix:
+        julia-version:
+          - 'nightly'
+        os:
+          - ubuntu-latest
+          - windows-latest
+        julia-arch:
+          - x64
+          - x86
```
fredrikekre: Just to keep the keys in the job name consistent with the `include:`d entries:
```suggestion
        os:
          - ubuntu-latest
          - windows-latest
        julia-arch:
          - x64
          - x86
        julia-version:
          - 'nightly'
```

fredrikekre:
https://github.com/JuliaLang/LinearAlgebra.jl/blob/master/.ci/run_tests.jl can be removed too I think? Was just used in the buildkite pipeline.

araujoms:
Amazing! Tests are failing because of #1664

dkarrasch:
I'm working on it, but probably lacking the right words. Will post something soon, and hopefully people will quickly suggest improvements.

dkarrasch:
What does it take to make the new docstrings be seen? Do we first need to bump the stdlib? Have I missed something in terms of providing and including the docstrings in #1667?

jishnub:
Yes I think the bump is needed

dkarrasch:
Hallelujah!

PR #1667, opened 2026-07-23
RFC: Add docstrings to `generic_mat[mat/vec]mul!`

### LINKED ISSUE #1664: Docstrings tests do not run in CI (and currently fail)

dkarrasch:
This is just to get started. Please help with improving wording and clarity.

Fixes #1664.

### DIFF
--- a/src/matmul.jl
+++ b/src/matmul.jl
@@ -1049,6 +1049,19 @@ end
 # legacy method, retained for backward compatibility
 generic_matvecmul!(C::AbstractVector, tA, A::AbstractVecOrMat, B::AbstractVector, _add::MulAddMul = MulAddMul()) =
     generic_matvecmul!(C, tA, A, B, _add.alpha, _add.beta)
+
+"""
+    generic_matvecmul!(c::AbstractVector, tA, A::AbstractVecOrMat, b::AbstractVector, α::Number, β::Number)
+
+Combined inplace matrix-vector multiply-add ``A b α + c β`` (`tA == 'N'`),
+``A^⊤ b α + c β`` (`tA == 'T'`), or ``A' b α + c β`` (`tA == 'C'`).
+The result is stored in `c` by overwriting it.  Note that `c` must not be
+aliased with either `A` or `b`.
+
+This is an abstraction layer below [`mul!`](@ref) used to dispatch on storage types.
+Packages that provide their own storage type are advised to overload `generic_matvecmul!`
+instead of `mul!`.
+"""
 @inline function generic_matvecmul!(C::AbstractVector, tA, A::AbstractVecOrMat, B::AbstractVector,
                                     alpha::Number, beta::Number)
     tA_uc = uppercase(tA) # potentially convert a WrapperChar to a Char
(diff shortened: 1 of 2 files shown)


araujoms (approved):
I think it's fine as it is, these are deep technical functions, just need to be clear and correct, no point in making them beginner-friendly.

jishnub:
Seems fine to me, although the style recommendation in Julia is to use the imperative tone.

dkarrasch:
> the style recommendation in Julia is to use the imperative tone.

~~The first paragraph is very close to the `mul!` docstring. Do you think the second one should be more imperative?~~

EDIT: I think I got it. Perhaps the `mul!` docstring may need some overhaul.

dkarrasch:
Since this doesn't seem too bad, let's go with this and fix LinearAlgebra's CI. Any upcoming comments could be included afterwards in another PR.

src/matmul.jl
```diff
@@ -1049,6 +1049,19 @@ end
...
+"""
+    generic_matvecmul!(c::AbstractVector, tA, A::AbstractVecOrMat, b::AbstractVector, α::Number, β::Number)
+
+Calculates the combined matrix-vector multiply-add ``A b α + c β`` (`tA == 'N'`),
+``A^⊤ b α + c β`` (`tA == 'T'`), or ``A' b α + c β`` (`tA == 'C'`).
+The result is stored in `c` by overwriting it.  Note that `c` must not be
+aliased with either `A` or `b`.
+
+This is an abstraction layer below [`mul!`](@ref) used to dispatch on storage types.
+Packages that provide their own storage type are advised to overload `generic_matvecmul!`
```
andreasnoack, line 1062: I wasn't aware of this (but haven't followed issues closely for some time). Was this discussed in an issue?

dkarrasch: This "idea" was introduced in https://github.com/JuliaLang/julia/pull/49806.

andreasnoack: Thanks. I do realize that the change there was made a while ago, but given that it is now being documented, it will be a bit more official, so I'll comment anyway. In my understanding, the generic linear algebra implementations were the ones that would not need to be overloaded but would work for most things, if sometimes inefficiently. Hence, I'm curious why you didn't use `mul!` with the same arguments as the common "interface"?

dkarrasch: No special reason. Nobody had that idea at the time. Just to clarify: you mean `mul!(C, tA, tB, A, B, alpha, beta)`? I feel a bit embarrassed because it does sound like a smart idea. Then we wouldn't have to make this clumsy `generic_...mul!` public.

andreasnoack: > you mean `mul!(C, tA, tB, A, B, alpha, beta)`?

Yes, exactly. Given that the functions haven't been documented until now (which hasn't made it into a release yet) as the way to overload multiplication, you could argue that it hasn't been official. Are you aware of any downstream packages are relying on this?

dkarrasch: Yes, a few GPU-arrays-related packages. Basically packages that would use the characters sooner or later in their pipeline anyway.

But we could rename the functions internally and have the most generic char-based `mul!` call `generic_xyzmul!` and actually recommend in NEWS.md that packages with own storage types should overload `mul!`. I'm not sure we have announced this mechanism "officially".

PR #1668, opened 2026-07-26
use dedicated uppercase function instead of adding methods to `Base.uppercase`

araujoms:
Closes #1384
```julia
julia> Base.infer_effects(*, Tuple{Matrix{Int}, Matrix{Int}})
(!c,?e,!n,!t,+s,?m,!u,+o,+r)
```
Unfortunately that doesn't solve the more important problem of
```julia
julia> Base.infer_effects(*, Tuple{Matrix{Float64}, Matrix{Float64}})
(!c,!e,!n,!t,!s,!m,!u,+o,!r)
```
But I'm afraid that's impossible, because
```julia
julia> Base.infer_effects(BLAS.gemm!, Tuple{Char,Char,Bool,Matrix{Float64},Matrix{Float64},Bool,Matrix{Float64}})
(!c,!e,!n,!t,!s,!m,!u,+o,!r)
```

PR #1669, opened 2026-07-27
Prune old LA in GHA CI

dkarrasch:
We need some dummy change to test whether it's working. Should the `env` variables be passed by the `test_args` option, and then be read in `runtests.jl` from `ARGS`?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,6 +42,11 @@ jobs:
       - uses: julia-actions/cache@v3
       - uses: julia-actions/julia-buildpkg@v1
       - uses: julia-actions/julia-runtest@v1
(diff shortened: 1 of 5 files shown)


dkarrasch:
Boy, now test coverage says that my dummy function is untested. So, which tests are running actually?

dkarrasch:
Alright, code coverage may not be 100% correct. I checked that the code and the tests from this repo are loaded.

KristofferC:
Such a simple function probably gets constant propagated and I think coverage doesn't track that properly

dkarrasch:
Thanks, @KristofferC, for your help. I think this PR resolved the issue. #1668 is running smoothly now.

PR #1671, opened 2026-07-31
RFC: Rename `generic_matxyzmul!` into `mul!`

dkarrasch:
@andreasnoack suggested in https://github.com/JuliaLang/LinearAlgebra.jl/pull/1667#discussion_r3668321941 to use the name `mul!` for the unwrapped multiplication methods. This can be done in a non-breaking way, and reserves the name `generic_mat[vec/mat]mul!` for actually generic methods. Unfortunately, they have at least(!) one underscore in front of them.

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -616,8 +616,6 @@ below (e.g. `mul!`) according to the usual Julia convention.
 
 ```@docs
 LinearAlgebra.mul!
-LinearAlgebra.generic_matvecmul!
-LinearAlgebra.generic_matmatmul!
 LinearAlgebra.lmul!
 LinearAlgebra.rmul!
 LinearAlgebra.ldiv!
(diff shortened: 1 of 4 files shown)


jishnub:
We may deprecate the generic_xyzmul functions and suggest mul! as the official one to overload?

dkarrasch:
But we would still need to make sure that LinAlg code runs by them to avoid breakage? @maleadt

araujoms:
The `public` names have never made it into a release, so it's fair game to change them. IMHO `mul!` is a nicer name.

dkarrasch:
Shall we move forward with this? The shift to the `generic_mat***mul!` machinery was introduced in v1.10, and probably at that time also introduced in some packages.

I guess if we accept this, then we should backport to v1.13 and revert the `public`action of the generic methods?

PR #1672, opened 2026-08-05
cancellation: Hook up OpenBLAS

Keno:
The relevant infrastructure just merged on nightly

### DIFF
--- a/src/blas.jl
+++ b/src/blas.jl
@@ -140,15 +140,29 @@ function _cancel_detect()
     lock(_cancel_lock) do
         _cancel_state[] == 0 || return
         state = 1
-        for lib in get_config().loaded_libs
-            h = Libdl.dlopen_e(lib.libname)
-            h == C_NULL && continue
-            ptok = Libdl.dlsym_e(h, :openblas_cancel_token)
-            ptok == C_NULL && continue
-            _cancel_cancel_f[] = Libdl.dlsym(h, :openblas_cancel)
-            _cancel_tok_f[] = ptok
-            state = 2
-            break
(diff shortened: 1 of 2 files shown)


src/blas.jl
```diff
@@ -1651,15 +1746,20 @@ for (gemm, elty) in
...
-                 transA, transB, m, n,
-                 ka, alpha, A, max(1,stride(A,2)),
-                 B, max(1,stride(B,2)), beta, C,
-                 max(1,stride(C,2)), 1, 1)
+            _cancel_prepare(Float64(m) * Float64(n) * Float64(ka))
+            # The annotation implies no cancellation point: the pre-call
+            # check binds the calling task's governing source (which gates
+            # delivery), and the post-call check throws a cancellation
+            # delivered mid-operation before the aborted (garbage) output
+            # can escape.
```
Keno: Drop comment.

src/blas.jl
```diff
@@ -1651,15 +1746,20 @@ for (gemm, elty) in
...
+            Base.@cancel_check
```
Keno: Let's factor our the read of the bound cancellation source, since that's not super cheap.

src/blas.jl
```diff
@@ -109,6 +110,95 @@ Return an object representing the current `libblastrampoline` configuration.
 """
 get_config() = lbt_get_config()
 
+# ── Cancellable BLAS calls ───────────────────────────────────────────────────
+# If the loaded BLAS provides the (staged) OpenBLAS cancellation extension,
```
Keno: It's not staged, it's merged upstream.

PR #1674, opened 2026-08-14
Backports to release 1.13

araujoms:
- [x] #1602 
- [x] #1666


dkarrasch:
Seems like we need to backport the CI changes and perhaps adapt them? Currently, there are not tests running.

.github/workflows/ci.yml
```diff
@@ -12,6 +12,41 @@ concurrency:
...
+      fail-fast: false
+      matrix:
+        os:
+          - ubuntu-latest
+          - windows-latest
+        julia-arch:
+          - x64
+          - x86
+        julia-version:
+          - 'nightly'
```
dkarrasch: Thanks for taking care of the backport, @araujoms! Now we need to tell GHA to use v1.13 on this branch. Because it's not fully released yet, it requires ''~1.13.0-0'', here and below.

araujoms: Thanks, I had no idea how to do this. Let's see if it works.

PR #1677, opened 2026-08-18
Respect the `norm` keyword in `isapprox` for arrays

### LINKED ISSUE #1675: `isapprox()` handles `norm` argument incorrectly
Since https://github.com/JuliaLang/LinearAlgebra.jl/commit/35a4427fd1a638afe92b5473d1c5f96e920b7436, `isapprox(x::AbstractArray, y::AbstractArray; norm)` does not use the `norm` keyword argument correctly. To calculate the error, it now always uses `norm_x_minus_y(x, y)` which is hard-coded to `norm(x - y)` [here](https://github.com/JuliaLang/LinearAlgebra.jl/blame/03f32350d47494ebcf4173415adc60cc6e1bc696/src/generic.jl#L2017). On the other hand `norm()` (from the kwarg) is used for `norm(x)` and `norm(y)` that are used with the relative tolerance [here](https://github.com/JuliaLang/LinearAlgebra.jl/blame/03f32350d47494ebcf4173415adc60cc6e1bc696/src/generic.jl#L2009).

This causes test failures, for example on  julia-1.12.6
```julia
julia> using Test

julia> @test isapprox([0.057618841449997994, -0.055947092101983294, 0.7492941853741162], [0.057618841449997994, -0.055947092101983335, 0.7492941853741162]; norm=(x)->NaN, rtol = 4.0e-11, atol = 4.0e-11)
Test Passed
```
but on julia-1.13.0-rc3
```julia
julia> using Test

julia> @test isapprox([0.057618841449997994, -0.055947092101983294, 0.7492941853741162], [0.057618841449997994, -0.055947092101983335, 0.7492941853741162]; norm=(x)->NaN, rtol = 4.0e-11, atol = 4.0e-11)
Test Failed at REPL[2]:1
  Expression: isapprox([0.057618841449997994, -0.055947092101983294, 0.7492941853741162], [0.057618841449997994, -0.055947092101983335, 0.7492941853741162]; norm = (x->begin
                #= REPL[2]:1 =#
                NaN
            end), rtol = 4.0e-11, atol = 4.0e-11)
   Evaluated: isapprox([0.057618841449997994, -0.055947092101983294, 0.7492941853741162], [0.057618841449997994, -0.055947092101983335, 0.7492941853741162]; norm = var"#2#3"(), rtol = 4.0e-11, atol = 4.0e-11)
ERROR: There was an error during testing
```

The docstring says
> When `x` and `y` are arrays, if `norm(x-y)` is not finite (i.e. `±Inf`or `NaN`), the comparison falls back to checking whether all elements of `x` and `y` are approximately equal component-wise.

The maximum (element-wise) difference between the two arrays in this example is 4.163336342344337e-17, which is less than `atol = 4.0e-11`, so the test failure is incorrect.

Maybe `norm_x_minus_y` should be the default for the kwarg, i.e.
```julia
function isapprox(x::AbstractArray, y::AbstractArray;
    atol::Real=0,
    rtol::Real=Base.rtoldefault(promote_leaf_eltypes(x),promote_leaf_eltypes(y),atol),
    nans::Bool=false, norm::Function=norm_x_minus_y)
    d = norm(x, y)
    if isfinite(d)
        return iszero(rtol) ? d <= atol : d <= max(atol, rtol*max(norm(x), norm(y)))
    else
        # Fall back to a component-wise approximate comparison
        # (mapreduce instead of all for greater generality [#44893])
        return mapreduce((a, b) -> isapprox(a, b; rtol=rtol, atol=atol, nans=nans), &, x, y)
    end
end
```
?

PS I guess there is some missing test coverage - maybe it would be useful to add a test for `isapprox()` with `norm=(x)->NaN` and `norm=(x)->Inf` somewhere?

jishnub:
Fixes #1675, and addresses https://github.com/JuliaLang/LinearAlgebra.jl/pull/1378#discussion_r2410297577.

Since #1378, `isapprox(x::AbstractArray, y::AbstractArray; norm)` evaluated the distance using `norm_x_minus_y(x, y)`, which was hard-coded to use the 2-norm, while the `norm` keyword was still used for the relative tolerance. A custom `norm` was therefore ignored where it mattered most, and in particular the documented fallback to a component-wise comparison for a non-finite `norm(x-y)` was never triggered:

```julia
julia> using LinearAlgebra

julia> x = [0.057618841449997994, -0.055947092101983294, 0.7492941853741162];

julia> y = [0.057618841449997994, -0.055947092101983335, 0.7492941853741162];

julia> isapprox(x, y; norm = v -> NaN, rtol = 4e-11, atol = 4e-11)
false # master
true  # this PR
```

Following @stevengj's suggestion, the keyword is passed on to `norm_x_minus_y`, and the allocation-free iterator path added in #1378 is dispatched on `::typeof(norm)`, so that it only applies for the default norm. The fast path is unchanged for the default case, and `isapprox` remains allocation-free for `Array`s, contiguous views of `Array`s and `Diagonal`s.

Tests are added for a non-finite `norm` (which must fall back to the component-wise comparison) and for a finite custom `norm` (which must determine the result), for each of the types that have a specialized `norm_x_minus_y` method.

### DIFF
--- a/src/diagonal.jl
+++ b/src/diagonal.jl
@@ -1155,7 +1155,7 @@ function generic_normp(D::Diagonal, p)
     end
     return v
 end
-norm_x_minus_y(D1::Diagonal, D2::Diagonal, nrm::typeof(norm)) = norm_x_minus_y(D1.diag, D2.diag, nrm)
+norm_x_minus_y(D1::Diagonal, D2::Diagonal, norm) = norm_x_minus_y(D1.diag, D2.diag, norm)
 
 _opnorm1(A::Diagonal) = maximum(norm(x) for x in A.diag)
 _opnormInf(A::Diagonal) = maximum(norm(x) for x in A.diag)
--- a/src/diagonal.jl
+++ b/src/diagonal.jl
@@ -1155,7 +1155,7 @@ function generic_normp(D::Diagonal, p)
     end
     return v
 end
-norm_x_minus_y(D1::Diagonal, D2::Diagonal) = norm_x_minus_y(D1.diag, D2.diag)
+norm_x_minus_y(D1::Diagonal, D2::Diagonal, nrm::typeof(norm)) = norm_x_minus_y(D1.diag, D2.diag, nrm)
 
 _opnorm1(A::Diagonal) = maximum(norm(x) for x in A.diag)
 _opnormInf(A::Diagonal) = maximum(norm(x) for x in A.diag)
--- a/src/diagonal.jl
+++ b/src/diagonal.jl
@@ -1155,7 +1155,7 @@ function generic_normp(D::Diagonal, p)
     end
     return v
 end
-norm_x_minus_y(D1::Diagonal, D2::Diagonal, nrm::typeof(norm)) = norm_x_minus_y(D1.diag, D2.diag, nrm)
+norm_x_minus_y(D1::Diagonal, D2::Diagonal, ::typeof(norm)) = norm_x_minus_y(D1.diag, D2.diag, norm)
 
 _opnorm1(A::Diagonal) = maximum(norm(x) for x in A.diag)
 _opnormInf(A::Diagonal) = maximum(norm(x) for x in A.diag)
--- a/src/diagonal.jl
+++ b/src/diagonal.jl
@@ -1155,7 +1155,7 @@ function generic_normp(D::Diagonal, p)
     end
     return v
 end
-norm_x_minus_y(D1::Diagonal, D2::Diagonal, norm) = norm_x_minus_y(D1.diag, D2.diag, norm)
+norm_x_minus_y(D1::Diagonal, D2::Diagonal, nrm::typeof(norm)) = norm_x_minus_y(D1.diag, D2.diag, nrm)
 
 _opnorm1(A::Diagonal) = maximum(norm(x) for x in A.diag)
 _opnormInf(A::Diagonal) = maximum(norm(x) for x in A.diag)
--- a/src/generic.jl
+++ b/src/generic.jl
@@ -2008,7 +2008,7 @@ function isapprox(x::AbstractArray, y::AbstractArray;
     atol::Real=0,
     rtol::Real=Base.rtoldefault(promote_leaf_eltypes(x),promote_leaf_eltypes(y),atol),
     nans::Bool=false, norm::Function=norm)
-    d = norm_x_minus_y(x, y)
+    d = norm_x_minus_y(x, y, norm)
     if isfinite(d)
         return iszero(rtol) ? d <= atol : d <= max(atol, rtol*max(norm(x), norm(y)))
     else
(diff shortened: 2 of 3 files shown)


stevengj (approved):

src/generic.jl
```diff
@@ -2018,10 +2018,10 @@ function isapprox(x::AbstractArray, y::AbstractArray;
     end
 end
 
-norm_x_minus_y(x, y) = norm(x - y)
+norm_x_minus_y(x, y, nrm::F) where {F} = nrm(x - y)
 FastContiguousArrayView{T,N,P<:Array,I<:Tuple{AbstractUnitRange, Vararg{Any}}} = Base.SubArray{T,N,P,I,true}
 const ArrayOrFastContiguousArrayView = Union{Array, FastContiguousArrayView}
-function norm_x_minus_y(x::ArrayOrFastContiguousArrayView, y::ArrayOrFastContiguousArrayView)
+function norm_x_minus_y(x::ArrayOrFastContiguousArrayView, y::ArrayOrFastContiguousArrayView, ::typeof(norm))
```
dkarrasch, line 2024: x-ref: https://github.com/JuliaLang/LinearAlgebra.jl/pull/1378#discussion_r3806026933

Why `LinearAlgebra.norm` only?

stevengj: Because we know for sure that `LinearAlgebra.norm` on `Array` is an L2 norm that works on iterators, whereas other norms might not, e.g. `opnorm`.

src/generic.jl
```diff
@@ -2018,10 +2018,10 @@ function isapprox(x::AbstractArray, y::AbstractArray;
     end
 end
 
-norm_x_minus_y(x, y) = norm(x - y)
+norm_x_minus_y(x, y, norm::F) where {F} = norm(x - y)
 FastContiguousArrayView{T,N,P<:Array,I<:Tuple{AbstractUnitRange, Vararg{Any}}} = Base.SubArray{T,N,P,I,true}
 const ArrayOrFastContiguousArrayView = Union{Array, FastContiguousArrayView}
-function norm_x_minus_y(x::ArrayOrFastContiguousArrayView, y::ArrayOrFastContiguousArrayView)
+function norm_x_minus_y(x::ArrayOrFastContiguousArrayView, y::ArrayOrFastContiguousArrayView, norm::F) where {F}
```
stevengj:
```suggestion
function norm_x_minus_y(x::ArrayOrFastContiguousArrayView, y::ArrayOrFastContiguousArrayView,::typeof(norm))
```
You have to restrict it to `LinearAlgebra.norm` because an arbitrary `norm` function might not work on iterators.

stevengj: For example, suppose they pass in `norm = opnorm`.  (Would be good to add a test for this.)

test/generic.jl
```diff
@@ -963,6 +963,24 @@ end
...
+            @test !isapprox(x, y .+ 1; norm=nonfinitenorm, rtol=4e-11, atol=4e-11)
+            @test isapprox(Diagonal(x), Diagonal(y); norm=nonfinitenorm, rtol=4e-11, atol=4e-11)
+            @test isapprox(view(x, :), view(y, :); norm=nonfinitenorm, rtol=4e-11, atol=4e-11)
+        end
+        # the specified norm is used to evaluate the distance
+        @test isapprox(x, y; norm=v -> 0.0, atol=0, rtol=0)
+        @test !isapprox(x, x; norm=v -> 1.0, atol=0, rtol=0)
+        @test !isapprox(Diagonal(x), Diagonal(x); norm=v -> 1.0, atol=0, rtol=0)
+        @test !isapprox(view(x, :), view(x, :); norm=v -> 1.0, atol=0, rtol=0)
+        @test isapprox(x, y; norm=v -> 1.0, atol=2)
```
stevengj, line 990: Maybe add a test to be sure that it handles non-iterable norms like `opnorm` correctly, as noted above, e.g.
```jl
X = [1 0; 0 1]; Y = X + [1 0; 0 2]*1e-3
@test !isapprox(X, Y, atol=0.0021) # norm(X - Y) > opnorm(X - Y)
for (X, Y) in ((X, Y), (Diagonal(X), Diagonal(Y)))
    @test isapprox(X, Y, atol=0.0021, norm=opnorm) && !isapprox(X, Y, atol=0.00199, norm=opnorm)
end
```
…in case someone tries to broaden `::typeof(norm)` again.

src/diagonal.jl
```diff
@@ -1155,7 +1155,7 @@ function generic_normp(D::Diagonal, p)
     end
     return v
 end
-norm_x_minus_y(D1::Diagonal, D2::Diagonal) = norm_x_minus_y(D1.diag, D2.diag)
+norm_x_minus_y(D1::Diagonal, D2::Diagonal, nrm::typeof(norm)) = norm_x_minus_y(D1.diag, D2.diag, nrm)
```
```suggestion
norm_x_minus_y(D1::Diagonal, D2::Diagonal, ::typeof(norm)) = norm_x_minus_y(D1.diag, D2.diag, norm)
```

PR #1679, opened 2026-08-18
restore codepoint and conversion from UInt32 for WrapperChar

simeonschaub:
These were removed in #1668, but SparseArrays still relies on these methods in `generic_matmatmul!` since it calls `uppercase`. Found tracking down regressions in AMDGPU.jl on nightly

ref https://github.com/JuliaGPU/AMDGPU.jl/pull/1027

### DIFF
--- a/src/LinearAlgebra.jl
+++ b/src/LinearAlgebra.jl
@@ -624,6 +624,11 @@ function Base.Char(w::WrapperChar)
     end
 end
 WrapperChar(c::Char) = WrapperChar(c, _isuppercase(c))
+# The `AbstractChar` interface requires `codepoint` and construction from `UInt32`;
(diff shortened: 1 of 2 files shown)


araujoms:
Perhaps we should also remove `uppercase` from SparseArrays, for the same reason we removed it from LinearAlgebra.

simeonschaub:
Sure, want to make a PR? I believe we should still merge this though, since we do subtype AbstractChar, so we should conform to the interface. This type does occasionally leak into other code as the AMDGPU example shows, so we should make sure we don't break anyone's code

araujoms:
Sure, I just think you should add a test here otherwise these methods are bound to get removed again.

simeonschaub:
Ok, will do!

araujoms (approved):
Thanks! And sorry for removing those methods in the first place.

PR #1680, opened 2026-08-23
feat: implement matrix absolute value support with unit tests

### LINKED ISSUE #1648: Add `abs` matrix function

Cedriq1astaken:
Closes: #1648

Implements the matrix absolute value $|A| = \sqrt{A^\dagger A}$ for abstract matrices using the SVD factorization.
Specialized dispatches are included for:
- `SelfAdjoint` (`Hermitian`/`Symmetric`): Spectral decomposition via `eigen`.
- `Diagonal`: Element-wise absolute value via `applydiagonal`.
- Empty matrices: Returns `zeros(float(T), n, n)`.

I tried @[stevengj](https://github.com/stevengj) suggestion to try the qdwh algorithm
But it seemed consistently slower than SVD factorization

Benchmark:

```
Matrix Size (N×N)   SVD Time      QDWH Time    
================================================
10 × 10             0.026 ms      0.093 ms      
25 × 25             0.136 ms      0.469 ms     
50 × 50             0.885 ms      3.148 ms      
100 × 100           3.953 ms      18.092 ms     
200 × 200           23.303 ms     84.571 ms     
400 × 400           89.852 ms     295.538 ms   
800 × 800           338.54 ms     1100.807 ms   
================================================

```

### DIFF
--- a/src/dense.jl
+++ b/src/dense.jl
@@ -1175,11 +1175,7 @@ julia> abs(A)
 """
 function abs(A::AbstractMatrix{T}) where T
     if isdiag(A)
-        if T <: Real
-            return Hermitian(applydiagonal(abs ∘ float, A))
-        else
-            return Hermitian(applydiagonal(complex ∘ abs ∘ float, A))
-        end
+        return Hermitian(convert(AbstractMatrix{float(T)}, applydiagonal(abs, A)))
     elseif ishermitian(A)
         return abs(Hermitian(A))
     end
--- a/src/dense.jl
+++ b/src/dense.jl
@@ -2058,12 +2058,13 @@ julia> abs(A)
 function abs(A::AbstractMatrix{T}) where T
     m, n = size(A)
     if isempty(A)
-        return Hermitian(zeros(float(real(T)), n, n))
+        return Hermitian(zeros(float(T), n, n))
     elseif isdiag(A)
-        return Hermitian(applydiagonal(abs ∘ float, A))
+        return Hermitian(applydiagonal(x -> float(T)(abs(x)), A))
     elseif ishermitian(A)
         return abs(Hermitian(A))
     end
     u, s, v = svd(A)
     return Hermitian(v * Diagonal(s) * v')
 end
+
--- a/src/dense.jl
+++ b/src/dense.jl
@@ -2067,4 +2067,3 @@ function abs(A::AbstractMatrix{T}) where T
     u, s, v = svd(A)
     return Hermitian(v * Diagonal(s) * v')
 end
-
--- a/src/dense.jl
+++ b/src/dense.jl
@@ -2058,7 +2058,7 @@ julia> abs(A)
 function abs(A::AbstractMatrix{T}) where T
     m, n = size(A)
     if isdiag(A)
-        return Hermitian(applydiagonal(x -> float(T)(abs(x)), A))
+        return Hermitian(diagm(float(T).(abs.(diag(A)))))
     elseif ishermitian(A)
         return abs(Hermitian(A))
     end
--- a/src/dense.jl
+++ b/src/dense.jl
@@ -2057,8 +2057,6 @@ julia> abs(A)
 
 function abs(A::AbstractMatrix{T}) where T
     m, n = size(A)
-    if isempty(A)
-        return Hermitian(zeros(float(T), n, n))
     elseif isdiag(A)
         return Hermitian(applydiagonal(x -> float(T)(abs(x)), A))
     elseif ishermitian(A)
--- a/src/dense.jl
+++ b/src/dense.jl
@@ -2058,9 +2058,9 @@ julia> abs(A)
 function abs(A::AbstractMatrix{T}) where T
     m, n = size(A)
     if isempty(A)
-        return zeros(float(T), n, n)
+        return Hermitian(zeros(float(real(T)), n, n))
     elseif isdiag(A)
-        return applydiagonal(abs, A)
+        return Hermitian(applydiagonal(abs ∘ float, A))
     elseif ishermitian(A)
         return abs(Hermitian(A))
     end
--- a/src/dense.jl
+++ b/src/dense.jl
@@ -1159,6 +1159,32 @@ end
 @inline _broadcast!!(f, dest, args...) = broadcast(f, args...)
 
 """
+    abs(A::AbstractMatrix)
+
+Compute the matrix absolute value `|A| = sqrt(A'A)`
+
+# Examples
+```jldoctest
+julia> A = [1.0 2.0; 3.0 4.0];
+
+julia> abs(A)
+2×2 Hermitian{Float64, Matrix{Float64}}:
+ 2.05798  2.40098
+ 2.40098  3.77297
+```
+"""
+function abs(A::AbstractMatrix{T}) where T
+    m, n = size(A)
+    if isdiag(A)
+        return Hermitian(diagm(float(T).(abs.(diag(A)))))
+    elseif ishermitian(A)
+        return abs(Hermitian(A))
+    end
+    u, s, v = svd(A)
+    return Hermitian(v * Diagonal(s) * v')
+end
+
+"""
     cos(A::AbstractMatrix)
 
 Compute the matrix cosine of a square matrix `A`.
@@ -2037,31 +2063,3 @@ function lyap(A::AbstractMatrix{T}, C::AbstractMatrix{T}) where {T<:BlasFloat}
     rmul!(Q * Y * Q', inv(scale))
 end
 lyap(a::Union{Real,Complex}, c::Union{Real,Complex}) = -c/(2real(a))
-
-"""
-    abs(A::AbstractMatrix)
-
-Compute the matrix absolute value `|A| = sqrt(A'A)`
-
-# Examples
-```jldoctest
-julia> A = [1.0 2.0; 3.0 4.0];
-
-julia> abs(A)
-2×2 Hermitian{Float64, Matrix{Float64}}:
- 2.05798  2.40098
- 2.40098  3.77297
-```
-"""
-
-
-function abs(A::AbstractMatrix{T}) where T
-    m, n = size(A)
-    if isdiag(A)
-        return Hermitian(diagm(float(T).(abs.(diag(A)))))
-    elseif ishermitian(A)
-        return abs(Hermitian(A))
-    end
-    u, s, v = svd(A)
-    return Hermitian(v * Diagonal(s) * v')
-end
--- a/src/dense.jl
+++ b/src/dense.jl
@@ -2037,3 +2037,34 @@ function lyap(A::AbstractMatrix{T}, C::AbstractMatrix{T}) where {T<:BlasFloat}
     rmul!(Q * Y * Q', inv(scale))
 end
 lyap(a::Union{Real,Complex}, c::Union{Real,Complex}) = -c/(2real(a))
+
+"""
+    abs(A::AbstractMatrix)
+
+Compute the matrix absolute value `|A| = sqrt(A'A)`
+
+# Examples
+```jldoctest
+julia> A = [1.0 2.0; 3.0 4.0];
+
+julia> abs(A)
+2×2 Hermitian{Float64, Matrix{Float64}}:
+ 2.05798  2.40098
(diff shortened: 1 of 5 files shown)


araujoms:
That's surprising. Can you share your implementation of the QDWH algorithm so that I can benchmark it?

Cedriq1astaken:
Sure
Here is my implementation

``` julia
function h(i)
    if i ≥ 1
        return 3.0, 1.0, 3.0
    end
    d = cbrt(4 * (1 - i^2) / i^4)
    b0 = sqrt(1 + d)
    a = b0 + 0.5 * sqrt(8 - 4*d + (8 * (2 - i^2)) / (i^2 * b0))
    b = (a - 1)^2 / 4
    c = a + b - 1
    return a, b, c
end

function qdwh(A::AbstractMatrix{T}; 
        maxiter::Int = 6, 
        tol::Real = cbrt(5 * eps(real(T)))) where T
    n = size(A, 1)
    α = norm(A)               
    normA1 = opnorm(A, 1)
    F_lu = lu(A)
    rcond = LAPACK.gecon!('1', copy(F_lu.factors), normA1)
    β = (rcond * normA1) / sqrt(n)
    X = A / α                    
    l = β / α                   
    for k in 1:maxiter
        a, b, c = h(l)
        sqrtc = sqrt(c)

        Q, R = qr([sqrtc * X; Matrix{T}(I, n, n)])
        Q₁ = @view Q[1:n, 1:n]
        Q₂ = @view Q[n+1:2n, 1:n]

        Xnew = (b / c) * X + (1 / sqrtc) * (a - b / c) * (Q₁ * Q₂')

        if norm(Xnew - X) <= tol * norm(Xnew)
            X = Xnew
            break
        end
        X = Xnew
        l = l * (a + b * l^2) / (1 + c * l^2)
    end

    U = X
    H = U' * A
    return Hermitian((H + H') / 2)
end

```

stevengj:
That implementation has a lot more allocations than needed…

src/dense.jl
```diff
@@ -2037,3 +2037,33 @@ function lyap(A::AbstractMatrix{T}, C::AbstractMatrix{T}) where {T<:BlasFloat}
...
+```
+"""
+
+
+function abs(A::AbstractMatrix{T}) where T
+    m, n = size(A)
+    if isempty(A)
+        return zeros(float(T), n, n)
+    elseif isdiag(A)
+        return applydiagonal(abs, A)
```
```suggestion
        return Hermitian(zeros(float(real(T)), n, n))
    elseif isdiag(A)
        return Hermitian(applydiagonal(abs ∘ float, A))
```
You should always return the same `Hermitian` type regardless of the contents of `A`.   And would be good to have an `@inferred` test for type stability.

stevengj:
That being said, I suspect that we will want to start with the simpler SVD method in the first PR for this feature, with a to-do to experiment with faster versions in the future.

src/dense.jl
araujoms: `abs` is a nice function that can be made fully type-stable. You should add it to the list here https://github.com/JuliaLang/LinearAlgebra.jl/blob/51b1b94cdade9ee6771f7359cdd7a53233ba33d9/test/symmetric.jl#L1320 (and make the test pass)

araujoms:
I wrote an in-place version of the function. It doesn't make any difference, the time taken by the QR decompositions alone is already larger than the time taken by the SVD decomposition. Doesn't look like there is hope for this algorithm.

```julia
function qdwh(A::AbstractMatrix{T}; maxiter::Int = 6, tol::Real = cbrt(5 * eps(real(T)))) where {T}
    m, n = size(A)
    α = norm(A)
    normA1 = opnorm(A, 1)
    F_lu = lu(A)
    rcond = LAPACK.gecon!('1', copy(F_lu.factors), normA1)
    β = (rcond * normA1) / sqrt(n)
    X = A / α
    l = β / α
    M = zeros(T, m + n, n)
    Xnew = zeros(T, m, n)
    diff = zeros(T, m, n)
    for i in 1:n
        M[m+i, i] = 1
    end
    for k in 1:maxiter
        a, b, c = h(l)
        sqrtc = sqrt(c)

        for j in 1:n, i in 1:m
            M[i, j] = sqrtc * X[i, j]
        end
        Q, R = qr(M)
        Q = Matrix(Q)
        Q₁ = @view Q[1:m, 1:n]
        Q₂ = @view Q[m+1:m+n, 1:n]

        mul!(Xnew, Q₁, Q₂', (1 / sqrtc) * (a - b / c), false)
        Xnew .+= (b / c) .* X

        diff .= Xnew .- X
        X .= Xnew
        if norm(diff) <= tol * norm(Xnew)
            break
        end
        l = l * (a + b * l^2) / (1 + c * l^2)
    end

    return Hermitian(X' * A)
end
```

src/dense.jl
Cedriq1astaken: Okay I updated `abs` to return a Hermitian matrix consistently across all branches and added `@inferred` tests in test/dense.jl.  For `abs` being a nice function, since it returns a `Hermitian` rather than a `Matrix`, I updated the loop branching to assert `@inferred(abs(a)) isa Hermitian{T}.`

src/dense.jl
```diff
@@ -2037,3 +2037,33 @@ function lyap(A::AbstractMatrix{T}, C::AbstractMatrix{T}) where {T<:BlasFloat}
...
+ 2.05798  2.40098
+ 2.40098  3.77297
+```
+"""
+
+
+function abs(A::AbstractMatrix{T}) where T
+    m, n = size(A)
+    if isempty(A)
+        return Hermitian(zeros(float(T), n, n))
```
araujoms:
```suggestion
```
Please remove this dispatch, it's not useful. The empty case should be taken care of by the diagonal case.

src/dense.jl
```diff
@@ -2037,3 +2037,33 @@ function lyap(A::AbstractMatrix{T}, C::AbstractMatrix{T}) where {T<:BlasFloat}
...
+    elseif isdiag(A)
+        return Hermitian(applydiagonal(x -> float(T)(abs(x)), A))
+    elseif ishermitian(A)
+        return abs(Hermitian(A))
```
araujoms: These paths are not tested.

araujoms:
What was wrong with `Hermitian(applydiagonal(abs ∘ float, A))`?

Also, you should put this function together with the other spectral functions, around line 1500.

Cedriq1astaken:
> What was wrong with `Hermitian(applydiagonal(abs ∘ float, A))`?
> 
> Also, you should put this function together with the other spectral functions, around line 1500.

I tried `Hermitian(applydiagonal(abs ∘ float, A))`, but the `@inferred` test failed for complex matrices because abs returns real numbers, `Hermitian{Float64}` instead of `Hermitian{ComplexF64}`. Converting back with float(T) fixed the test.

Alternatively, we could do `if T <: Real` / `else` with `complex ∘ abs ∘ float` if preferred.?

araujoms:
Yes, please do that, it's more readable and allocates less.

src/dense.jl
```diff
@@ -1158,6 +1158,32 @@ end
...
+julia> A = [1.0 2.0; 3.0 4.0];
+
+julia> abs(A)
+2×2 Hermitian{Float64, Matrix{Float64}}:
+ 2.05798  2.40098
+ 2.40098  3.77297
+```
+"""
+function abs(A::AbstractMatrix{T}) where T
+    m, n = size(A)
```
araujoms:
```suggestion
```
This line is useless.

araujoms (approved):

src/dense.jl
```diff
@@ -1158,6 +1158,31 @@ end
...
+
+julia> abs(A)
+2×2 Hermitian{Float64, Matrix{Float64}}:
+ 2.05798  2.40098
+ 2.40098  3.77297
+```
+"""
+function abs(A::AbstractMatrix{T}) where T
+    if isdiag(A)
+        return Hermitian(convert(AbstractMatrix{float(T)}, applydiagonal(abs, A)))
```
stevengj: You need `float` *before* the abs so that it doesn't fail for e.g. `Int8(-128)` due to integer overflow, and then putting `float(T)` after the `abs` ensures that it gets the correct (possibly complex) output type without allocating a whole new matrix.
```suggestion
        return Hermitian(applydiagonal(float(T) ∘ abs ∘ float, A)))
```

test/dense.jl
```diff
@@ -1456,4 +1456,46 @@ end
...
+    # Complex valued non-square
+    A = randn(ComplexF64, N, N+2)
+    H = @inferred abs(A)
+    @test H'H ≈ A'A
+    @test H isa Hermitian{ComplexF64}
+
+    # Dense diagonal matrix
+    D = diagm([1.0, -2.0, 3.0, -4.0])
+    @test (@inferred abs(D)) ≈ diagm([1.0, 2.0, 3.0, 4.0])
+    @test abs(D) isa Hermitian{Float64}
```
stevengj, line 1477: Maybe add a test for `diagm(Int8[-128, 127])` to make sure you get `diagm([128.0, 127.0])`, making sure you guard against integer overflow for `abs(Int8(-128))` as noted above.

src/dense.jl
araujoms: That doesn't work, you can't apply `float(T)` to a matrix.

stevengj: Oh, I thought `applydiagonal` was applying to the diagonal elements.

Then maybe:
```suggestion
        D = Diagonal(map(float(T) ∘ abs ∘ float, diagview(A)))
        return copyto!(similar(A, eltype(D)), D)
```

stevengj: (Though it seems like we could just modify `applydiagonal` to apply the given function to the diagonal element-by-element and it would work in all of the existing use cases, while being more flexible for cases like this…)

araujoms: @jishnub added `applydiagonal` in https://github.com/JuliaLang/julia/pull/56483 , what do you think? It seems to me that the reasoning was to allow for matrices of matrices.

In any case, I'm loathe to use something other than `applydiagonal` here, the matrix functions are pretty well standardized with it. Another possibility is
```julia
        return Hermitian(convert(AbstractMatrix{float(T)}, applydiagonal(abs ∘ float, A)))
```

jishnub: The reasoning in that PR was simply to reroute a call f(A) to the potentially optimized f(Diagonal(A)) if the cheaper O(N^2) check indicated that the matrix was diagonal.

araujoms: I see, the point is to avoid having two independent implementations of f(Diagonal(A)), which is very sensible.

What do you think @stevengj , is it acceptable now?

stevengj: It's really annoying to make an extra copy of the whole dense matrix, which is what the current version does…

src/dense.jl
```diff
@@ -1158,6 +1158,31 @@ end
...
+
+julia> abs(A)
+2×2 Hermitian{Float64, Matrix{Float64}}:
+ 2.05798  2.40098
+ 2.40098  3.77297
+```
+"""
+function abs(A::AbstractMatrix{T}) where T
+    if isdiag(A)
+        return Hermitian(convert(AbstractMatrix{float(T)}, applydiagonal(abs ∘ float, A)))
```
araujoms:
```suggestion
        f(x) = convert(AbstractMatrix{float(T)}, x) #preserves complexes and quaternions
        return Hermitian(applydiagonal(f ∘ abs ∘ float, A))
```
I see. This avoids the dense copy.

stevengj (approved):

PR #1681, opened 2026-08-26
Backports release 1.13

dkarrasch:
- [x] #1651
- [x] #1671
- [x] #1676
- [x] #1689 
- [x] #1669


dkarrasch:
Could it be that the code gets loaded from the sysimage? It doesn't seem to see the added methods.

araujoms:
I think I forgot to include https://github.com/JuliaLang/LinearAlgebra.jl/pull/1669 in the backports.

PR #1686, opened 2026-08-27
[Apple Silicon] Set BLAS threads to all but efficiency cores

### LINKED ISSUE #61281: M5 Pro and M5 Max Core detection
The following [discourse post](https://discourse.julialang.org/t/multithreading-on-apple-m5-chips/136109) asks about Julia's handling/detection of the new "Performance" cores on the recently announced M5 Pro and Max. https://github.com/JuliaLang/julia/pull/42099 seems to justifiably disable the efficiency cores, but maybe these new middling cores should be enabled and used by Julia?

christiangnrd:
Now based on #1691

On Apple Silicon, the recommendation for number of BLAS threads is to ignore the Efficiency cores. Julia has been doing this by setting the number of CPU for the whole runtime to the number of higher performance tier cores. This worked when all Apple Silicon only had Performance and Efficiency cores, but now some chips have no Efficiency cores (M5 Pro and M5 Max) and the upcoming M6 has 3 performance tiers.

This PR moves that logic to just BLAS thread selection (sister PR in Julia to see all the cores: https://github.com/JuliaLang/julia/pull/62891).

The new logic for BLAS thread selection on Apple Silicon is now:
- If the processor has Efficiency cores, subtract those from the total number of cores
- Otherwise, set it to total number of cores -1 to prevent the computer from freezing up.

@gbaraldi

Also closes https://github.com/JuliaLang/julia/issues/61281

### DIFF
--- a/src/LinearAlgebra.jl
+++ b/src/LinearAlgebra.jl
@@ -886,9 +886,9 @@ end
             elseif count > 1 && _sysctl_int32("hw.cpufamily") == 0x1b588bb3  # CPUFAMILY_ARM_FIRESTORM_ICESTORM (Apple M1)
                 offset = Int32(4)  # the M1 has 4 efficiency cores
             end
-            return count - offset
+            return min(Sys.EFFECTIVE_CPU_THREADS, count - offset)
         catch
-            return @ccall(jl_effective_threads()::Cint) - 1
+            return Sys.EFFECTIVE_CPU_THREADS - 1
         end
     end
 end
@@ -907,7 +907,7 @@ function lbt_openblas_onload_callback()
         @static if Sys.isapple() && Sys.ARCH === :aarch64
             nthreads = max(1, cpus_to_use())
         else
-            nthreads = max(1, @ccall(jl_effective_threads()::Cint) ÷ 2)
+            nthreads = max(1, Sys.EFFECTIVE_CPU_THREADS ÷ 2)
         end
         BLAS.lbt_set_num_threads(nthreads)
     end
--- a/src/LinearAlgebra.jl
+++ b/src/LinearAlgebra.jl
@@ -844,6 +844,55 @@ function versioninfo(io::IO=stdout)
     return nothing
 end
 
+@static if Sys.isapple() && Sys.ARCH === :aarch64
+    @noinline function _sysctl_str(name)
+        size = Ref{Csize_t}()
+        err = @ccall sysctlbyname(name::Cstring, C_NULL::Ptr{Cvoid}, size::Ref{Csize_t},
+                                  C_NULL::Ptr{Cvoid}, 0::Csize_t)::Cint
+        Base.systemerror("sysctlbyname", err != 0)
+
+        str = Base._string_n(size[] - 1) # implicitly allocates trailing NUL byte
+        err = @ccall sysctlbyname(name::Cstring, str::Ptr{UInt8}, size::Ptr{Csize_t},
+                                    C_NULL::Ptr{Cvoid}, 0::Csize_t)::Cint
+        Base.systemerror("sysctlbyname", err != 0)
+
+        return str
+    end
+
+    @noinline function _sysctl_int32(name)
+        value = Ref{Int32}(0)
+        size = Ref{Csize_t}(sizeof(Int32))
+        err = @ccall sysctlbyname(name::Cstring, value::Ptr{Cvoid}, size::Ref{Csize_t},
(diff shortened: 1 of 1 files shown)


christiangnrd:
Should the `@ccall(jl_effective_threads()::Cint)` calls be replaced with `Sys.EFFECTIVE_CPU_THREADS` to respect the user's `JULIA_CPU_THREADS` value?

christiangnrd:
Also not sure what to do about coverage. The lines lacking coverage require bare-metal, macos <11, and sandbox execution with some of the syscalls blacklisted.

staticfloat:
The spirit of this is good and I'm in favor.  It'd be nice if we could verify our assumptions with some data; for instance on my M1 Pro with 8 performance and 2 efficiency cores, running [this vibed-up code](https://github.com/staticfloat/AppleMSeriesBLASCoreComparison.jl) to compare performance and efficiency cores, I get:

dgemm_cores_m1pro

I'd be interested in seeing what you get on an M5 or (eventually) an M6.

christiangnrd:
@staticfloat I vibed it into a package so that I can ask my non-technical friends to run it on their devices as a one-liner. LMK if you want me to open a PR. On my M2 Max. It's looking pretty similar to your M1 Pro.

(Updated to include https://github.com/christiangnrd/AppleMSeriesBLASCoreComparison.jl/pull/1) M2 Max:
dgemm_cores

christiangnrd:
(updated to include https://github.com/christiangnrd/AppleMSeriesBLASCoreComparison.jl/pull/1) M5 (with updated colours for Super cores):
IMG_2999

staticfloat (approved):

christiangnrd:
@staticfloat As I mentioned in https://github.com/JuliaLang/LinearAlgebra.jl/pull/1691#discussion_r3889522332, I'm hoping the logic move can make it in 1.13 so that people with devices like a neo don't have julia pretending to detect only 2/6 cores. If that's still possible (I realize it's pretty late in the 1.13 cycle), then #1691 should be merged first so that only that one can be backported. Otherwise we may as well just merge this one directly

src/LinearAlgebra.jl
```diff
@@ -844,6 +844,49 @@ function versioninfo(io::IO=stdout)
...
+        offset = Int32(1)
+        # hw.nperflevels/hw.perflevel* only exist on Darwin ≥ 21 (macOS ≥ 12)
+        count = _sysctl_int32("hw.physicalcpu")
+        if parse(VersionNumber, _sysctl_str("kern.osrelease")) >= v"21"
+            for i in 0:_sysctl_int32("hw.nperflevels")-1
+                _sysctl_str("hw.perflevel$i.name") == "Efficiency" || continue
+                offset = _sysctl_int32("hw.perflevel$i.physicalcpu")
+                break
+            end
+        elseif ncpu > 1 && _sysctl_int32("hw.cpufamily") == 0x1b588bb3  # CPUFAMILY_ARM_FIRESTORM_ICESTORM (Apple M1)
```
IanButterworth: Claude:

---

`ncpu` is undefined here (the variable is `count`), so the macOS < 12 branch throws `UndefVarError`.

```suggestion
        elseif count > 1 && _sysctl_int32("hw.cpufamily") == 0x1b588bb3  # CPUFAMILY_ARM_FIRESTORM_ICESTORM (Apple M1)
```

christiangnrd:
(Updated to include https://github.com/christiangnrd/AppleMSeriesBLASCoreComparison.jl/pull/1)) M5 Pro from @IanButterworth:
IMG_1230

christiangnrd:
I cleaned of the commits and added a comment since it would otherwise be easy to miss that the logic is all-1 on devices without efficiency cores.

I was planning on following up with changing the use of `@ccall(jl_effective_threads()::Cint)` with `Sys.EFFECTIVE_CPU_THREADS` since that'll better respect user intent. **Is that something we would want backported too or would that be pushing it for 1.13?**

Currently, `JULIA_CPU_THREADS` does nothing to the the default number of BLAS threads and it probably should?

christiangnrd:
@IanButterworth agreed with the `Sys.EFFECTIVE_CPU_THREADS` change on Slack. I've also added a try catch block and a fallback for the unlikely case that `sysctlbyname` ever fails.

The PR is currently 4 somewhat clean commits with the last 2 being changes for review. Once approved I can squash the first 3 since the 4th should probably be merged as a separate commit.

I'm done touching this it's safe to review

IanButterworth (approved):

christiangnrd:
All ready! (**DO NOT SQUASH**)

PR #1688, opened 2026-08-27
[CI] Fix slow aarch64 macos times

christiangnrd:
Forcing too many jobs on the macOS runners with 7GB memory leads to https://github.com/JuliaTesting/ParallelTestRunner.jl/issues/124, which ends up with the testsuite taking 90 minutes with 3 parallel workers as opposed to 60 minutes when letting PTR pick 1 job.

All other runners are unaffected because they all have at least 2GB of available memory per available core.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -17,10 +17,6 @@ end
 # file without it and falls back to the serial path.
 if Base.find_package("ParallelTestRunner") !== nothing
     using ParallelTestRunner
-    # Auto CPU thread count detection in ParallelTestRunner is bad
-    if Sys.isapple() && Sys.ARCH === :aarch64
-        push!(ARGS, "--jobs=2")
-    end
     push!(ARGS, "--verbose")
     testsuite = Dict{String,Expr}(splitext(f)[1] => :(include($(joinpath(@__DIR__, f))))
                                   for f in testfiles)
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -18,6 +18,9 @@ end
 if Base.find_package("ParallelTestRunner") !== nothing
     using ParallelTestRunner
     # Auto CPU thread count detection in ParallelTestRunner is bad
+    if Sys.isapple() && Sys.ARCH === :aarch64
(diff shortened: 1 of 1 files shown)


KristofferC:
Please compare all timings (I mean for all OSs) before and after this change.

christiangnrd:
> Please compare all timings (I mean for all OSs) before and after this change.

All non aarch64 macos runners end up with the same number of jobs because they have a reasonable amount of ram so on CI this shouldn't affect the other platforms.

[60 minutes](https://github.com/JuliaLang/LinearAlgebra.jl/actions/runs/33114321453/job/98664931878?pr=1688) on 1 core instead of [90 with 3](https://github.com/JuliaLang/LinearAlgebra.jl/actions/runs/33107675849/job/98642186731?pr=1686). I'm curious to see what 2 jobs will do. Edit: [73 minutes with 2 threads](https://github.com/JuliaLang/LinearAlgebra.jl/actions/runs/33120454844/job/98685739778?pr=1688)

christiangnrd:
[This](https://github.com/JuliaLang/LinearAlgebra.jl/actions/runs/33276933975/job/99165181118) apple silicon run with 3 cores ran in 50 minutes. I still think a consistent 60 minutes is better than usually 90+ minutes with the occasional fast run but I could also restore the forced 3 jobs and this PR just adds the --verbose flag so I can see if there's a pattern for those 50 minute runs

KristofferC:
Is it worth experimenting with a lower `JULIA_TEST_MAXRSS_MB` while keeping the parallelism high?

christiangnrd:
> Is it worth experimenting with a lower `JULIA_TEST_MAXRSS_MB` while keeping the parallelism high?

Probably. Although I think this should get merged right away for the `--verbose` flag and the obvious improvements.

I opened https://github.com/JuliaLang/LinearAlgebra.jl/pull/1693 with the max rss values I use in Metal.jl for the 8GB workers. I don't know how much of a difference it'll make since from the 2 worker run it seems to have caused slowdowns before worker RSS got high but fingers crossed 🤞

PR #1691, opened 2026-08-29
[Apple Silicon] Move BLAS thread selection logic from Base

christiangnrd:
This is like #1686, but it just moves over the thread selection logic from Base (https://github.com/JuliaLang/julia/pull/62891) so it's hopefully easier to merge.

### DIFF
--- a/src/LinearAlgebra.jl
+++ b/src/LinearAlgebra.jl
@@ -844,6 +844,43 @@ function versioninfo(io::IO=stdout)
     return nothing
 end
 
+@static if Sys.isapple() && Sys.ARCH === :aarch64
+    @noinline function _sysctl_str(name)
+        size = Ref{Csize_t}()
+        err = @ccall sysctlbyname(name::Cstring, C_NULL::Ptr{Cvoid}, size::Ptr{Csize_t},
+                                  C_NULL::Ptr{Cvoid}, 0::Csize_t)::Cint
+        Base.systemerror("sysctlbyname", err != 0)
+
+        buf = Vector{UInt8}(undef, size[])
+        err = @ccall sysctlbyname(name::Cstring, buf::Ptr{Cvoid}, size::Ptr{Csize_t},
+                                  C_NULL::Ptr{Cvoid}, 0::Csize_t)::Cint
+        Base.systemerror("sysctlbyname", err != 0)
+
+        # drop the trailing NUL terminator
+        return String(resize!(buf, size[]-1))
+    end
+
+    @noinline function _sysctl_int32(name)
+        value = Ref{Int32}(0)
+        size = Ref{Csize_t}(sizeof(Int32))
+        err = @ccall sysctlbyname(name::Cstring, value::Ptr{Cvoid}, size::Ptr{Csize_t},
+                                  C_NULL::Ptr{Cvoid}, 0::Csize_t)::Cint
+        Base.systemerror("sysctlbyname", err != 0)
+        return value[]
+    end
+
+    const cpus_to_use = OncePerProcess{Int32}() do
+        # hw.nperflevels/hw.perflevel* only exist on Darwin ≥ 21 (macOS ≥ 12)
+        count = _sysctl_int32("hw.physicalcpu")
+        if parse(VersionNumber, _sysctl_str("kern.osrelease")) >= v"21"
+            count = _sysctl_int32("hw.perflevel0.physicalcpu")
+        elseif count > 1 && _sysctl_int32("hw.cpufamily") == 0x1b588bb3  # CPUFAMILY_ARM_FIRESTORM_ICESTORM (Apple M1)
+            count -= Int32(4)  # the M1 has 4 efficiency cores
+        end
+        return count
+    end
+end
+
 function lbt_openblas_onload_callback()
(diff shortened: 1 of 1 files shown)


src/LinearAlgebra.jl
```diff
@@ -844,6 +844,43 @@ function versioninfo(io::IO=stdout)
     return nothing
 end
 
+@static if Sys.isapple() && Sys.ARCH === :aarch64
+    @noinline function _sysctl_str(name)
+        size = Ref{Csize_t}()
+        err = @ccall sysctlbyname(name::Cstring, C_NULL::Ptr{Cvoid}, size::Ptr{Csize_t},
```
stevengj: Slightly better style:
```suggestion
        err = @ccall sysctlbyname(name::Cstring, C_NULL::Ptr{Cvoid}, size::Ref{Csize_t},
```
(also elsewhere).

src/LinearAlgebra.jl
```diff
@@ -844,6 +844,43 @@ function versioninfo(io::IO=stdout)
...
+                                  C_NULL::Ptr{Cvoid}, 0::Csize_t)::Cint
+        Base.systemerror("sysctlbyname", err != 0)
+
+        buf = Vector{UInt8}(undef, size[])
+        err = @ccall sysctlbyname(name::Cstring, buf::Ptr{Cvoid}, size::Ptr{Csize_t},
+                                  C_NULL::Ptr{Cvoid}, 0::Csize_t)::Cint
+        Base.systemerror("sysctlbyname", err != 0)
+
+        # drop the trailing NUL terminator
+        return String(resize!(buf, size[]-1))
```
stevengj:
```suggestion
        str = Base._string_n(size[] - 1) # implicitly allocates trailing NUL byte
        err = @ccall sysctlbyname(name::Cstring, str::Ptr{UInt8}, size::Ptr{Csize_t},
                                  C_NULL::Ptr{Cvoid}, 0::Csize_t)::Cint
        Base.systemerror("sysctlbyname", err != 0)
        
        return str
```

src/LinearAlgebra.jl
```diff
@@ -855,8 +892,8 @@ function lbt_openblas_onload_callback()
 
     # https://github.com/xianyi/OpenBLAS/blob/c43ec53bdd00d9423fc609d7b7ecb35e7bf41b85/README.md#setting-the-number-of-threads-using-environment-variables
     if !haskey(ENV, "OPENBLAS_NUM_THREADS") && !haskey(ENV, "GOTO_NUM_THREADS") && !haskey(ENV, "OMP_NUM_THREADS")
-        @static if Sys.isapple() && Base.BinaryPlatforms.arch(Base.BinaryPlatforms.HostPlatform()) == "aarch64"
-            nthreads = max(1, @ccall(jl_effective_threads()::Cint))
```
stevengj, line 859: Shouldn't we be fixing the `jl_effective_threads()` function instead?

christiangnrd: From my understanding, this is a decent heuristic for BLAS threads because of the kind of workload being run, however with other threading workloads (that yield more?), the scheduler is better able to split the work between cores and including the efficiency cores will be beneficial for those.

That's why I'm proposing Base not lie about the cores it sees and the threading logic for BLAS live here.

I'm doing it piecemeal because I'm secretly hoping this smaller change can make it into 1.13 so that BLAS behaviour remains unchanged but Base detects all the cores, and then the more involved change (https://github.com/JuliaLang/LinearAlgebra.jl/pull/1686) can be left for 1.14.

I also plan on looking into the use of jl_effective_threads here, I think using the new-ish `Sys.EFFECTIVE_CPU_THREADS` might better respect hardware limits and user wishes but I want to take the time to fully understand the repercussions of that before I propose any more changes to default behaviour

stevengj: Maybe Base should have separate functions to return the number of high-performance cores and the total number of cores?   So that the CPU detection still lives there?

christiangnrd: Eventually I would like to add some sort of API to do that and I'd fix this code to get rid of the ugly and out of place helpers, but I would like to see this logic moved to LinearAlgebra sooner rather than later since of the typical workloads, this behaviour only helps BLAS

staticfloat: I think I agree with Christian here in that the difference between efficiency and performances cores matters a whole lot more for BLAS than it does for arbitrary code.  Rather than try to make a statement at the `Base` level about what we should do here, doing it in `BLAS`, at least at first, seems reasonable.  The main issue in my mind is that the fork-join parallelism of OpenBLAS is a particularly bad choice for heterogenous cores, since it forces the faster cores to wait for the slower cores to finish their work before a chunk of work can be completed.  With a smarter scheduler (e.g. work-stealing, consumer-producer queue, etc...) this can be mitigated and adding efficiency cores to the pool actually helps (slightly).

I should note that simply choosing the correct number of threads here doesn't actually prevent your code from running on the efficiency cores; the OS scheduler can still choose to put things on efficiency cores if it so decides, but since that depends on the load on the rest of the system, I think we're doing the best we can here.

christiangnrd:
I'm closing this in favour of just #1686 since #1686 was marked for backport and I only opened this as a less disruptive change for backport

PR #1569, opened 2026-04-02
Force SymTridiagonal to have consistent lengths

musoke:
https://github.com/JuliaLang/LinearAlgebra.jl/issues/629 pointed out that is possible to initialise SymTridiagonal with subdiagonal of length equal to the diagonal. This is inconsistent with Tridiagonal and Bidiagonal.

Check the lengths and error if the subdiagonal's length is not 1 less than the diagonal's length.

Remove tests for SymTridiagonal with length(ev) == length(dv).

Todo: 
- [x] add test that `len(dv) == len(ev)` now errors
- [x] remove tests with `len(dv)==len(ev)`
- [x] fix kron
- [x] @andreasnoack mentioned that changes may be need in the wrapper for [dstemr](https://www.netlib.org/lapack/explore-html/d4/dec/group__stemr_ga71cbf49a0387762afa39582e6abbe466.html#ga71cbf49a0387762afa39582e6abbe466). 
I don't think that any changes are needed for dstemr. As far as I can tell from https://github.com/JuliaLang/LinearAlgebra.jl/issues/1491, `dstemr` is called via `dstgr!`. It already has checks in place for the lengths of `de` and `ev`:
https://github.com/JuliaLang/LinearAlgebra.jl/blob/17325b74ec807f5237228512db1131ddfe4048e3/src/lapack.jl#L4043-L4058

Possible followups:
- [x] simplify `_evview`, now that `ev` always has the same length.
- [x] remove most calls to `_evview`

PR #1692, opened 2026-08-31
Restore support for `length(ev) == length(dv)` in `SymTridiagonal`

KristofferC:
Since #1569, `SymTridiagonal(dv, ev)` throws a `DimensionMismatch` when `ev` has the same length as `dv`, which was previously accepted with the last element ignored (LAPACK convention). PkgEval on 1.14 shows this breaks released packages, via KrylovKit's `rayleighquotient` (fixed in KrylovKit v0.10.4), ExponentialUtilities' `expT!` (still present on master), and direct construction:

- [CustomGaussQuadrature](https://pkgeval.s3.us-east-2.amazonaws.com/runs/gh-5428371050/logs/primary/CustomGaussQuadrature.log)
- [Fermi](https://pkgeval.s3.us-east-2.amazonaws.com/runs/gh-5428371050/logs/primary/Fermi.log)
- [FirstPassageTools](https://pkgeval.s3.us-east-2.amazonaws.com/runs/gh-5428371050/logs/primary/FirstPassageTools.log)
- [KryburyCompress](https://pkgeval.s3.us-east-2.amazonaws.com/runs/gh-5428371050/logs/primary/KryburyCompress.log)
- [MRFingerprintingRecon](https://pkgeval.s3.us-east-2.amazonaws.com/runs/gh-5428371050/logs/primary/MRFingerprintingRecon.log)

This accepts the equal-length form again in the outer constructor, with a depwarn, dropping the last element of `ev` in a type-preserving way (`getindex` generally, `view` for `SubArray`s). The strict `length(ev) == length(dv) - 1` field invariant from #1569 and all its internal simplifications are kept, so only the constructor entry point changes. `Pkg.test` runs with `--depwarn=yes`, so downstream packages keep passing while maintainers see the warning.

The deprecated branch only triggers when the strict check would fail, keeping infinite matrices (where `length(dv) - 1 == length(dv)`) working.

Verified against the constructor patterns from the failing packages: plain `Vector`s, equal-length `SubArray`s (including the `StepRange`-into-`ReshapedArray` views from the ExponentialUtilities failure, where the stored `ev` keeps the input view type), ranges, and mixed eltypes.

### DIFF
--- a/src/tridiag.jl
+++ b/src/tridiag.jl
@@ -68,7 +68,19 @@ julia> A[2,1]
 ```
 """
 SymTridiagonal(dv::V, ev::V) where {T,V<:AbstractVector{T}} = SymTridiagonal{T}(dv, ev)
-SymTridiagonal{T}(dv::V, ev::V) where {T,V<:AbstractVector{T}} = SymTridiagonal{T,V}(dv, ev)
+function SymTridiagonal{T}(dv::V, ev::V) where {T,V<:AbstractVector{T}}
+    # the second condition is false for infinite vectors, where length(dv) - 1 == length(dv)
+    if length(ev) == length(dv) != 0 && length(ev) != length(dv) - 1
+        Base.depwarn("constructing a `SymTridiagonal` with `length(ev) == length(dv)` is deprecated; " *
+            "`ev` should have one element less than `dv`. The last element of `ev` is ignored.", :SymTridiagonal)
+        ev = _droplast(ev)
+    end
+    return SymTridiagonal{T,V}(dv, ev)
+end
+# drop the last element, preserving the array type where possible so that
+# the result converts back to `V` in the inner constructor
+_droplast(v::AbstractVector) = v[begin:end-1]
+_droplast(v::SubArray) = view(v, firstindex(v):(lastindex(v) - 1))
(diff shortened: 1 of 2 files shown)


stevengj (approved):

araujoms:
I don't think this is necessary. KrylovKit is already fixed, ExponentialUtilities has a PR to fix it, and CustomGaussQuadrature has a PR to fix it. All other packages fail because of KrylovKit and ExponentialUtilities, once they're updated they will be automatically fixed. There's no hurry, we still have a long time before 1.14.

And I'd prefer keeping the codebase clean.

KristofferC:
Keeping the code base non-breaking is also desirable?

araujoms:
It's not breaking. These packages were relying on undocumented internals, the bug's on them.

dkarrasch:
Any updates here? Are there any packages left that broke due to this?

araujoms:
There are two PRs that haven't been merged yet: https://github.com/SciML/ExponentialUtilities.jl/pull/277 and https://github.com/andrew-saydjari/KryburyCompress.jl/pull/6

The latter package seems abandoned, but the former one should be just a matter of pinging @ChrisRackauckas

ViralBShah:
In case we don’t hear from @ChrisRackauckas in some time here, I do have merge access there.

PR #1693, opened 2026-08-31
[CI] Use 2 threads on macOS aarch64 CI

christiangnrd:


christiangnrd:
[First run](https://github.com/JuliaLang/LinearAlgebra.jl/actions/runs/33430204911/job/99613630795) seems promising! The rebase will give another run and then I want to test 3 runners but that might be pushing it.

KristofferC:
Seems like that pushed it indeed :laughing:

christiangnrd:
If setting to 1 thread helps, we should keep, otherwise it slightly reduces coverage so I'll remove

KristofferC:
I doubt much actual time is spent within BLAS

christiangnrd:
> I doubt much actual time is spent within BLAS

Good point. I'll update all in one go once https://github.com/JuliaTesting/ParallelTestRunner.jl/pull/170 is released (or rejected)

christiangnrd:
Seems like 2 threads with lower max rss (committed upstream) is the sweet spot. I think this PR and https://github.com/JuliaLang/LinearAlgebra.jl/pull/1688 should eventually make it into the 1.13 branch

ViralBShah:
Should we merge?

christiangnrd:
Bump

PR #1694, opened 2026-08-31
fix left division with abstractly-typed rhs

araujoms:
Part of https://github.com/JuliaLang/julia/issues/62896

The rest is not due to #1475, but rather #1578. I'm not even sure it should be fixed, after all if you give an abstractly-typed vector as an input it is not a surprise that you get another as an output.


stevengj:
I'm inclined to give an abstract output in this sort of case…

PR #1701, opened 2026-09-20
Generic `lmul!`/`rmul!` for `QRCompactWYQ` and its adjoint

dkarrasch:
Multiplication by a `QRCompactWYQ` factor so far only had LAPACK-backed methods (`gemqrt!`), so it threw a `MethodError` as soon as either the factor or the other operand was not BLAS-compatible — for instance a `Float64` factorization applied to a `BigFloat` matrix:

```julia
julia> Q = qr(randn(7, 4), NoPivot(); blocksize=2).Q;

julia> lmul!(Q, big.(randn(7, 3)))
ERROR: MethodError: no method matching lmul!(::LinearAlgebra.QRCompactWYQ{…}, ::Matrix{BigFloat})
```

`QRPackedQ` already has generic implementations following LAPACK's `ormqr`; this adds the counterparts for the compact WY representation, covering all four directions.

## Implementation

With `V = I + tril(Q.factors, -1)`, `nb = size(Q.T, 1)` and `k = size(Q.T, 2)`,

```
Q  = (I - V₁T₁V₁') ⋯ (I - V_b T_b V_b')
Q' = (I - V_b T_b' V_b') ⋯ (I - V₁T₁'V₁')
```

so the four cases differ only in the order in which the blocks are applied and in whether `Tⱼ` or `Tⱼ'` is used. The `AQ` loop runs opposite to the `QB` one, mirroring how the existing `QRPackedQ` methods are arranged:

| | block order | triangular factor |
|---|---|---|
| `QB` | `b → 1` | `Tⱼ` |
| `QcB` | `1 → b` | `Tⱼ'` |
| `AQ` | `1 → b` | `Tⱼ` |
| `AQc` | `b → 1` | `Tⱼ'` |

Each side is therefore a single helper — `_lmul_compactwy!` and `_rmul_compactwy!` — taking that choice as a `Val` argument. Per block, `lmul!` forms `W = Vⱼ'B`, multiplies by the triangular factor in place and updates `B -= VⱼW`, treating the rows meeting `V`'s implicit unit diagonal separately from those strictly below it; `rmul!` is the transpose of that, working on columns. The only allocation is one workspace (`nb`-by-`size(B, 2)`, resp. `size(A, 1)`-by-`nb`) reused across blocks.

The new methods are strictly more general than the existing ones, so `gemqrt!` still wins dispatch for strided `BlasFloat` arguments.

## Tests

Added to `test/qr.jl`:

* agreement with `gemqrt!` over `Float32`/`Float64`/`ComplexF32`/`ComplexF64`, seven matrix shapes (tall, wide, square, degenerate, and one exceeding the default block size) and block sizes `1, 2, 3, 36`, for matrix and vector right-hand sides;
* the blocked application against the unblocked `QRPackedQ` one in full `BigFloat` precision, using reflectors whose `τ` is recomputed so that the representation is orthogonal to that precision rather than only to `Float64` accuracy — this is what actually pins down the block traversal, since a `Float64` reference cannot distinguish it;
* round-trip orthogonality, and consistency between the two sides, `A*Q == (Q'*A')'`;
* that the generic method is the one selected for non-BLAS element types, and the `DimensionMismatch` paths.

`test/qr.jl` goes from 5692 to 6304 passing tests; `test/abstractq.jl` is unchanged at 171. Both pass locally against this branch.

https://claude.ai/code/session_01Vv9VYe4nM23v8ruYwk9AQx

---
_Generated by [Claude Code](https://claude.ai/code/session_01Vv9VYe4nM23v8ruYwk9AQx)_

EDIT: This addresses https://github.com/JuliaLang/LinearAlgebra.jl/issues/1097#issuecomment-2500472934.


dkarrasch:
Multiplication via `Q[']*A` used to work prior to v1.10 and the `AbstractQ` overhaul. It fell back to the most generic matmat multiplication with elementwise indexing into `Q`.

araujoms:
I'm sorry I'm not reviewing AI-generated code.

